Issue #5, opened 2021-08-18
Define a simple workflow job and execute one-cmd tools through this Job


seanshpark:
This next task is to execute compiler commands in level of
- one-import-tf, one-import-tflite, one-import-onnx
- one-optimize
- one-quantize

What to do in this task is to define a Job and invoke command line tools through this Job.
- Job itself can be created in fixed hard-code.

seanshpark:
Next related tasks
- Reflect to TreeView GUI
- Import .cfg and construct workflow of jobs
- Full implementation of options - there's huge amount of options in one-optimize
- Implement with one-quantize

Need to investigate
- User scenario of editing workflow of jobs
- Better term for "workflow of jobs"

lemmaa:
> This next task is to execute compiler commands in level of
> 
> * one-import-tf, one-import-tflite, one-import-onnx
> * one-optimize
> * one-quantize

I think one-pack and one-codegen should also be included in this scope. In addition, as the next step, the case of one-profile can also be considered.

lemmaa:
To properly support one-codegen, we need a backend compiler :(

I created a separate issue for this. https://github.com/Samsung/ONE-vscode/issues/7

PR #14, opened 2021-08-30
DRAFT Initial workflow

seanshpark:
on-going draft of initial workflow of one commands.


seanshpark:
For #5

seanshpark:
close as initial workflow is now working

PR #17, opened 2021-08-30
Revise .gitignore

seanshpark:
This will revise .gitignore to include format.patch and local Makefile.
And remove package-lock.json from ignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
 out
 node_modules
-package-lock.json
+format.patch
+Makefile


.gitignore
```diff
@@ -2,3 +2,5 @@
 out
 node_modules
 package-lock.json
+format.patch
+Makefile
```
mhs4670go, line 6: Is `Makefile` needed?

seanshpark: This is for local for personal convenience usage.

seanshpark: Our ONE frontend also has `Makefile` in gitignore.

seanshpark:
From draft #14

.gitignore
llFreetimell: Hmm... if `Makefile` is not officially used/generated file and the purpose of adding this is just personal, I think this is not that good :(

seanshpark: Um? Can you explain what is not good?
In ONE and here, I'd like to use `Makefile` as my personal makefile for preparing stuffs.
First one is to `make format` that will run format checker.

llFreetimell: In my point of view, ONE and ONE-vscode seems different because

- ONE has **official** `Makefile`(https://github.com/Samsung/ONE/blob/master/Makefile.template). 
- ONE-vscode has **no official** `Makefile` neither in master nor #14.

So I thought that adding **non official**(personal) file to `.gitignore` seems not that good.
If we can add out `personal_script_1.sh` to `.gitignore`, it would be convenient but we do not add :(

llFreetimell: This is just my curiosity... does typescript usually use `Makefile` @.@?

seanshpark: Adding `Makefile` to gitignore is to make ones own custom Makefile (for ONERT, there is Makefile.template, but not for ONE frontend)... I can't exactly catch what is `official` or not. ONE frontend doesn't have it also.
I could propose a `Makefile.template` but not now, I have not much of build experience with TypeScript + VSCode extensions.

> If we can add our personal_script_1.sh to .gitignore, it would be convenient but we do not add :(

Um?  Makefile is for `make` tool, to make someone's build process efficient. I don't get how we can compare this with any scripts like `personal_script_1.sh `.

> This is just my curiosity... does typescript usually use Makefile @.@?

Not sure but I don't think so. As I wrote above, currently I use this for `make format` to run format checker.
This is my current `Makefile`
```
format:
	@npm run lint
	@bash infra/format $(ARGS)
```

And I think I can use this as of now
```
format:
	@npm run lint
	@bash infra/format $(ARGS)

prep:
	npm install

build:
	npm run compile
```

llFreetimell (approved):

mhs4670go (dismissed):
LGTM

.gitignore
```diff
@@ -2,3 +2,5 @@
 out
 node_modules
 package-lock.json
```
mhs4670go, line 4:
```suggestion
```
I think `package-lock.json` should not be ignored. You can see below document with `npm help package-lock.json` command.

> This file is intended to be committed into source repositories,  and  serves  various  pur-
>       poses:
>
>       o Describe  a  single representation of a dependency tree such that teammates, deployments,
>         and continuous integration are guaranteed to install exactly the same dependencies.
>
>       o Provide a facility for users to "time-travel" to previous states of node_modules  without
>         having to commit the directory itself.
>
>       o To facilitate greater visibility of tree changes through readable source control diffs.
>
>       o And  optimize  the installation process by allowing npm to skip repeated metadata resolu-
>         tions for previously-installed packages.

seanshpark: I'll revise this PR tomorrow :)

mhs4670go (approved):
LGTM

llFreetimell (approved):

Issue #24, opened 2021-08-31
Introduce webview


mhs4670go:
## What

Let's introduce webview for our one-vscode extension.

## Why

I described it a bit at https://github.com/Samsung/ONE-vscode/issues/10#issuecomment-906205645. 
> How about using the vscode extension's GUI a little more? It will also be helpful for users who are not familiar with ONE.

![image](https://user-images.githubusercontent.com/24587354/130928253-711d72e0-7c1f-4121-bf5f-66ed155bdfb3.png)

The webview allows our extension to create fully customizable views, which can be used to build complex user interfaces. I'm going to start drafting with basic webview panel API first. I hope it will help you guys who want to start our project.

PR #25, opened 2021-08-31
[DRAFT] Introduce webview

mhs4670go:
This draft is to introduce webview of vscode extension.

You can compile it and do some modification with these commands.
```bash
$ code ONE-vscode
$ npm run watch
# do whatever you want
```

In VSCode,
1. ctrl + shift + p
2. type "one-build"
3. You can see simple webview panel!
---

There are two parts that make a view to panel.
1. `private _getHtmlForWebview(webview: vscode.Webview)` in `src/OneBuildPanel.ts` (need "reload Window")
2. `media/main.js` (need "reload Webviews")

Also, you can see the console log of webview with
1. ctrl + shift + p
2. type "Open Webview Developer Tools"

Related : #24 
Signed-off-by: seongwoo <mhs4670go@naver.com>


OTKRyu:
i guess this vscode extension will be combined with many sub tasks.
so i suggests `media/compile/main.js` rather than `media/main.js` .

mhs4670go:
@OTKRyu Actually, I named it temporarily. It could be `media/OneBuild.js` as well, which aligns with `OneBuildPanle.ts`.
xxxPanel.ts -> xxx.js

Directory level and hierarchy should be discussed.

PR #31, opened 2021-09-05
Introduce initial project builder

seanshpark:
This will introduce initial empty project builder.

### DIFF
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -14,23 +14,11 @@
  * limitations under the License.
  */
 
-import * as vscode from 'vscode';
-import {Project} from './Project';
+import {Builder as _Builder} from './Project/Builder';
 
-export function activate(context: vscode.ExtensionContext) {
-  console.log('one-vscode activate OK');
+export namespace Project {
 
-  let projectBuilder = new Project.Builder();
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const Builder = _Builder;
 
-  projectBuilder.init();
-
-  let disposableOneVsc = vscode.commands.registerCommand('onevscode.build', () => {
-    console.log('one build...');
-    projectBuilder.build(context);
-  });
-  context.subscriptions.push(disposableOneVsc);
-}
-
-export function deactivate() {
-  // TODO do cleanup
-}
+}  // namespace Project
--- a/src/Project/Builder.ts
+++ b/src/Project/Builder.ts
@@ -15,22 +15,18 @@
  */
 
 import * as vscode from 'vscode';
-import {Project} from './Project';
 
-export function activate(context: vscode.ExtensionContext) {
-  console.log('one-vscode activate OK');
+export class Builder {
+  constructor() {}
 
-  let projectBuilder = new Project.Builder();
+  // TODO import .cfg file to BuildFlow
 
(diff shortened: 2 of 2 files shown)


seanshpark:
From draft #14

mhs4670go (approved):
LGTM

src/Project.ts
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Builder as _Builder} from './Project/Builder';
+
+export namespace Project {
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
mhs4670go, line 21: Just curiosity, what is it for?

seanshpark: lint will try to change `Builder` to `builder`. it uses camel case that starts with lower case.

src/Project.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Builder as _Builder} from './Project/Builder';
+
+export namespace Project {
```
llFreetimell, line 19: I think this file is related with `src/Project/*`, right?
Then `Project` means which one among following options?

- Project = `one-build`
- Project = `one-build` + other functions(like showing profiling json)

seanshpark: I'm not sure I can catch your question but `Project` here is about project management, like other IDE does.
I've used this term; as vscode-ext does not provide this feature, but we need to manage something like it.
`Project` here is a management of compilation workflow as of now... but not sure this is enough...

So for `Profiling` I could say `Profile.ts` + `Profile` folder; for your question, `src/NewFeature/`

This file+folder management style is from my draft coding and as I am just a novice extension starter,
I am not sure this style is good or bad...

We may have to refactor someday if this blocks something somehow...

llFreetimell: > So for Profiling I could say Profile.ts + Profile folder; for your question, src/NewFeature/

It seems that `src/Project` manages `ONE build` related features :)
If so, how about changing the name from `Project` to `ONE`, `Onecc`, or other name?
At first, I misunderstood that `Project` includes all of our features.
But if `src/dir1` manages only `dir1` related features, I think `Project` is so general to distinguish different features :)

seanshpark: As I wrote, this structure is may be refactored later.
Currently I do not want to change to `ONE` or `Onecc` as (1) I do not know `ONE` is also appropriate (2) I may messup the draft code that may not work anymore; which will take more time to solve.

seanshpark: Are you willing to use term `Project` ?

llFreetimell: I am not objecting to the structure, just asked about name.
I will not use the term `Project` but reason for question was just because `Project` seems to manage all the "project" in this repository.

llFreetimell (approved):

Issue #20, opened 2021-08-30
Enable PR format checker


seanshpark:
### WHAT
Let's enable format checker for the PRs.
- mainly TypeScript (JavaScript) for now
- expand to other files gradually

### WHY
To make our codes clean and good readability.

### INFO
I've used this as of now
```bash
npm run lint
bash infra/format $(ARGS)
```

seanshpark:
(Anyone interested in this task can assign oneself and go on,)

seanshpark:
#29 task should be prepared before to work on this

seanshpark:
To do: enable clang format

seanshpark:
done

PR #34, opened 2021-09-06
DRAFT enable format checker

seanshpark:
Test with actions to enable format checker


seanshpark:
For #20

PR #40, opened 2021-09-06
Initialize Builder current workspace

seanshpark:
This will initialize Builder current workspace folder with
introducing obtainWorkspaceRoot method in Helpers in Utils.

### DIFF
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


seanshpark:
From draft #14

src/Utils/Helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @brief Get Workspace root folder as string
+ * @note  will throw if not working in workspace mode.
+ */
+export function obtainWorkspaceRoot(): string {
+  const workspaceFolders = vscode.workspace.workspaceFolders;
+  if (!workspaceFolders) {
+    console.log('obtainWorkspaceRoot: NO workspaceFolders');
+    // TODO revise message
+    throw new Error('Need workspace to build with ONE commands');
```
llFreetimell, line 29: If this function is not dedicated for ONE commands, how about changing msg?

```suggestion
    throw new Error('Need workspace folders');
```

Or if this function is dedicated for ONE commands, how about changing the function name like `obtainOneCmdWorkspaceRoot` or filename like `src/Utils/OneCmdHelpers.ts` or etc?

seanshpark: > If this function is not dedicated for ONE commands, how about changing msg?

I'm not sure of this so I added `TODO`... anyway I'll make this shorter ambigious.

mhs4670go (approved):
LGTM

llFreetimell (approved):

PR #41, opened 2021-09-06
Check format checker

seanshpark:
Check if this gives format.patch file.


seanshpark:
patch upload is now enabled

PR #53, opened 2021-09-07
Introduce WorkJobs

seanshpark:
This will introduce WorkJobs as alias of Array<Job>.

### DIFF
--- a/src/Project/WorkJobs.ts
+++ b/src/Project/WorkJobs.ts
@@ -14,22 +14,6 @@
  * limitations under the License.
  */
 
-import {Logger} from '../Utils/Logger';
-import {WorkJobs} from './WorkJobs';
+import {Job} from './Job';
 
-export class WorkFlow {
-  logger: Logger;
-  workspace: string;
-  jobs: WorkJobs;
-
-  constructor(logger: Logger) {
-    this.logger = logger;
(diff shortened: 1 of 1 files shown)


seanshpark:
From draft #14

src/Project/WorkJobs.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Job} from './Job';
+
+export class WorkJobs extends Array<Job> {}
```
mhs4670go, line 19: I'm not sure but when it comes to alias, this seems be normally used.
```suggestion
export type WorkJobs = Array<Job> {};
```

seanshpark: I too am not sure and don't know indepth about `type` in TS, I would just like to go with class extends for now...

mhs4670go (approved):
LGTM

llFreetimell (approved):

PR #55, opened 2021-09-07
Introduce JobBase

seanshpark:
This will introduce JobBase class as base class for Jobs implementations.


seanshpark:
From draft #14

mhs4670go (approved):
LGTM

llFreetimell (approved):

PR #60, opened 2021-09-08
Introduce JobImportTF

seanshpark:
This will introduce JobImportTF and JobImprt for base class of import jobs.

NE-vscode-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

### DIFF
--- a/src/Project/JobImportTF.ts
+++ b/src/Project/JobImportTF.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


seanshpark:
From draft #14

src/Project/JobImportTF.ts
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+import {Job} from './Job';
+import {JobImport} from './JobImport';
+import {ToolArgs} from './ToolArgs';
+
+export class JobImportTF extends JobImport {
+  inputArrays?: string;
+  outputArrays?: string;
```
mhs4670go, line 23: Maybe?
```suggestion
  inputArrays: string;
  outputArrays: string;
```

seanshpark: These two maybe undefined if not given, from the user.

mhs4670go (approved):
LGTM

llFreetimell (approved):

Issue #68, opened 2021-09-09
Import cfg file to WorkFlow


seanshpark:
Let's import cfg file to WorkFlow
- Add more Job classes: for `one-pack` , `one-codegen`, `one-profile` (?)
- Support minimum options for functional implementation

In this task, we will NOT
- export to .cfg file
- support all options, like there are lots of optimization options

seanshpark:
Next task from #5

seanshpark:
Use model of https://github.com/Samsung/ONE-vscode/issues/4#issuecomment-900675346

and add quantize
```
[onecc]
one-import-tf=True
one-import-tflite=False
one-import-bcq=False
one-import-onnx=False
one-optimize=True
one-quantize=True
one-pack=True
one-codegen=True

[one-import-tf]
input_path=./inception_v3.pb
output_path=./inception_v3.circle
input_arrays=input
output_arrays=InceptionV3/Predictions/Reshape_1

[one-optimize]
input_path=./inception_v3.circle
output_path=./inception_v3.opt.circle

[one-quantize]
input_path=./inception_v3.opt.circle
output_path=./inception_v3.opt.q8.circle
granularity=channel

[one-pack]
input_path=./inception_v3.opt.q8.circle
output_path=inception_v3_pack

[one-codegen]
backend=dummy
command=-o inception_v3.bin ./inception_v3.opt.q8.circle
```

seanshpark:
Todo
- [ ] put reference .cfg file into source tree... where?
- [ ] Implement: invoke file open dialog and get the filepath
- [ ] Implement: Parsing INI format
- [ ] Implement: Simple import of .cfg items
- [ ] Implement: Add `JobPack`
- [ ] Implement: Add `JobCodegen`
- [ ] ...

seanshpark:
`profile` will be added in anoter task

seanshpark:
done with task

PR #70, opened 2021-09-10
DRAFT Import .cfg file

seanshpark:
on-going draft to import .cfg file to WorkFlow.


seanshpark:
For #68

seanshpark:
done landing

Issue #87, opened 2021-09-15
Question about the file directory and `gitignore`


parkjisu6239:
## Question
I was going to run an extension on the main branch.
However, the `.vscode` is contained in the `gitignore`, so there is no `.vscode` folder in the repository.
So it can't be executed 😥 @seanshpark 

**_1. Could you upload the `.vscode` folder additionally?_**

**_2. May I know what kind of function the extension you're making is performing?_**

</br>

## Problem
This is the  `Draft` #52  I posted last time and the [link](https://github.com/SOSCON-TheONE/ONE-vscode-extension-prototype/tree/barchart) to the visualization function currently under development.

We use `React` for visualization. React was needed for selection of bar graphs or various interactions.
However, in the process of using React, **there have been many changes in file structure and defense.**

So I'm having a big problem `writing PR and draft` now. Our file structure can be seen [here](https://github.com/SOSCON-TheONE/ONE-vscode-extension-prototype/tree/barchart).

How can we solve this problem? If you have an idea, I'll accept it!

</br>

## The functions we created are as follows.
![image](https://user-images.githubusercontent.com/42528189/133377496-541ca8a9-7bd0-4fb8-b9f1-ef3da0711c60.png)

### Done

- Click `Load`, a file selection dialog will pop up
- You can `visualize` JSON files.
- `Zoom-in/Zoom-out` is possible.
- `pid`(GPU/CPU) usage is displayed.
- You can click on the bar to see `More info`.
- You can `open/close`  the panel.
- Fixed left bar
- Make the color a bit prettier
- When zoomed in/out, the scale is dynamically created/deleted.
- Zoom in with the `slider`.
- It's linked to the me in vcode. All elements follow the vscdoe theme color.

| dark+                           | light+                            | All vcode theme support.      |
| ------------------------------- | --------------------------------- | ----------------------------- |
| ![image](https://user-images.githubusercontent.com/42528189/133377624-d77d6f13-a4f7-4691-8ffd-6eb25486a1ad.png) | ![image](https://user-images.githubusercontent.com/42528189/133377649-862205d1-abd4-41d0-be7f-5739a8c6e05d.png) | ![image](https://user-images.githubusercontent.com/42528189/133377664-d67504ba-11fa-44a3-a5dd-508ea88568f0.png) |

(I'll leave a draft after I finish what I'm currently developing.)
</br>

**We hope this feature will really help one developers. So I hope this problem will be solved and merge.** 🙏

[...]

PR #103, opened 2021-09-17
[Visualizer] Create React App for visualizer

parkjisu6239:
React will be used for visualization. These files are basically created when create a React app.


seanshpark:
I've used `[Project]` prefix to add files in `Project` folder and related.
I think you should use another prefix if this is not your intention.

seanshpark:
Please split files like `.gitignore` and others, by your purpose

src/Visualizer/.eslintrc.json
```diff
@@ -0,0 +1,19 @@
+{
```
seanshpark, line 1: same question as https://github.com/Samsung/ONE-vscode/pull/93#discussion_r709817145

.gitignore
```diff
@@ -2,3 +2,24 @@ out
...
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
seanshpark, line 25: Please remoe this red circle image

image

You need to add an empty line at the end

seanshpark:
Read this again carefully

https://github.com/Samsung/ONE-vscode/issues/87#issuecomment-919753850

These are by thoughts about PRs
- one context per PR, that is, one purpose of objective in one PR; that is there can be lots of PRs; this will take time to land a huge change
- too many purpose makes understanding codes difficult
- one who wrote the code knows what one is doing but others' don't
- communicating by written word is very hard

parkjisu6239:
@seanshpark

The files I uploaded now are default files created when I create an application with React framework.
It's not a folder that we created for random purposes.
So I thought it was unnecessary to subdivide it.
[React docs](https://ko.reactjs.org/docs/create-a-new-react-app.html)

Should we divide these into different PRs?
There are no folders or files like this in the PR created later.

.gitignore
parkjisu6239: thanks your feedback!

seanshpark:
> React docs
This file is a structure, so it doesn't work without any.

Please check like #93 

> hould we divide these into different PRs?

Yes

> There are no folders or files like this in the PR created later.

I don't understand what you mean

src/Visualizer/.eslintrc.json
wnsdud4197: When eslint encounters `package.json` or `.esconfig.json` during the `npm run lint`, the nearest file is overriding. There is a problem with the eslintconfig set in the `Visualizer/package.json` , so we will remove the setting and remove the `Visualizer/.eslintrc.json`

Thank you for the feedback! 🙏

parkjisu6239:
I'll subdivide it and upload it again! I'll close this PR

PR #107, opened 2021-09-17
[Visualizer] Add package.json and package-lock.json

parkjisu6239:
It includes the React library.


src/Visualizer/package.json
```diff
@@ -0,0 +1,35 @@
+{
```
seanshpark, line 1: we already have package.json in root folder. can you please merge packages needed into the root file?

parkjisu6239: This is because they are react-related libraries used only by Visualizer folder.

src/Visualizer/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "name": "vscode",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "react": "^17.0.2",
```
glistening, line 6: IANAL, but let me write a note.

A few years ago, react's license was not good for us because of patent. But it seems to be changed to `MIT`. See https://www.asapdevelopers.com/react-native-update-to-mit-license/ for more informaiton, I am not sure whether other dependencies are safe to use.

src/Visualizer/package.json
seanshpark: Can you please add a comment here, about explaining multiple package.json structure?

src/Visualizer/package.json
wnsdud4197: React is MIT license and has become a widely used framework. Can I ask you what kind of problems you're worried about?

llFreetimell: I checked that all the modules in `dependencies` have MIT or Apache 2.0 License.
But not for package-lock.json.
I will check it.

src/Visualizer/package.json
wnsdud4197: `package.json` used here are only packages for build react, and we divided the packages because there are dependencies that we need to create visualizers(build by react) but do not need to operate extensions.

seanshpark: Not related with `react` but in general. I just want to leave some histroy about this PR with multiple package.json files.
This is the first PR, as of now, with another package.json.

src/Visualizer/package.json
glistening: > React is MIT license and has become a widely used framework. 

I already am aware of. Please find my comment I've already written. : )

> Can I ask you what kind of problems you're worried about?

Please find what I've written —  _"I am not sure whether other dependencies are safe to use."_.
@llFreetimell answered for my question. He checked other modules' licenses.

src/Visualizer/package.json
parkjisu6239: I didn't understand exactly what you mean.
Are you saying that you need a reason to use multiple package.json files?

glistening: > `package.json` used here are only packages for build react, and we divided the packages because there are dependencies that we need to create visualizers(build by react) but do not need to operate extensions.

Do you mean the `dependencies` are used at development time only, and not used at runtime?
Then, can they be put in `devDependencies`?

seanshpark: > you need a reason to use multiple package.json files?

No, I'd like to leave a comment here about multiple package,.json files. I googled but failed (I'm not used to this domain) to get information about it.

parkjisu6239: What you mean is just transferring the content of the `dependencies` to the `devDependencies`?
If so, there is no problem and it is possible.

parkjisu6239: > No, I'd like to leave a comment here about multiple package,.json files. I googled but failed (I'm not used to this domain) to get information about it.

The app inside the `visualizer folder` recognizes `visualizer folder` as root.
package.json is usually placed in the root directory in a file structure.
If you look at this [link](https://github.com/yarnpkg/yarn/issues/6160), it refers to package management for multi-routes.

glistening: > What you mean is just transferring the content of the `dependencies` to the `devDependencies`?
> If so, there is no problem and it is possible.

No, just your answer is okay. For your information, I am interested in what is the part of license check and at this moment, checking the main license for each modules seems enough.

seanshpark (approved):
LGTM

PR #108, opened 2021-09-17
[Visualizer] Add content to .gitignore

parkjisu6239:
It includes the React stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,24 @@ out
 node_modules
 format.patch
 Makefile
+
+# dependencies
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# production
+build
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
--- a/.gitignore
+++ b/.gitignore
@@ -3,23 +3,5 @@ node_modules
 format.patch
 Makefile
 
-# dependencies
-/.pnp
-.pnp.js
-
-# testing
-/coverage
-
 # production
 build
-
-# misc
-.DS_Store
-.env.local
-.env.development.local
-.env.test.local
-.env.production.local
-
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,5 @@ node_modules
 format.patch
 Makefile
 
-# production
+# production for react-app
 build


.gitignore
```diff
@@ -2,3 +2,24 @@ out
 node_modules
 format.patch
 Makefile
+
+# dependencies
+/.pnp
+.pnp.js
```
seanshpark, line 8: Can I ask what these are from?
And it would be nice if you can add other items why they are needed.
(I'm not certain about all the items here)

parkjisu6239: This is the gitignore content that occurs automatically when you enter the command create-react-app.
In fact, it doesn't apply to our folder right now.
It seems no matter to erase, but it is probably considered a react recommendation.

seanshpark: If they are not used, please remove them. We can add when we need.

parkjisu6239: okay. Thank you for the feedback.

llFreetimell (approved):
LGTM

.gitignore
```diff
@@ -2,3 +2,6 @@ out
 node_modules
 format.patch
 Makefile
+
+# production
+build
```
seanshpark, line 7: Are the generated files put in `builder` folder? or is this for production (I think this is like package file) ?
We have `out` as generated files from `ts` to `js`.
I don't know much about react; is it possible to use `out` ?

parkjisu6239: > Are the generated files put in builder folder? or is this for production (I think this is like package file) ?

the former is correct!

>We have out as generated files from ts to js.
I don't know much about react; is it possible to use out ?

When the react is build, a compressed `output` is produced in the form of html, js, css, etc.
`Build` folder is the folder containing the output.
So, I think there are different kinds.

seanshpark: I curious about;
- when making a vscode extension package,  it is OK to have two output folders for artifactoy...

I've found some string is previous draft #93
- src/Visualizer/tsconfig.extension.json
```
        "target": "es6",
        "outDir": "build",
```
and some others
```
      // And restric the webview to only loading content from our extension's `media` directory.
      localResourceRoots: [vscode.Uri.file(path.join(this._extensionPath, 'build'))]
```
```
  private _getHtmlForWebview() {
    const manifest = require(path.join(this._extensionPath, 'build', 'asset-manifest.json'));
    const scriptPathOnDisk = vscode.Uri.file(path.join(this._extensionPath, 'build', mainScript));
    const stylePathOnDisk = vscode.Uri.file(path.join(this._extensionPath, 'build', mainStyle));
```
if `build` is changed to `out`, does it hurt the functionality?

parkjisu6239: > if build is changed to out, does it hurt the functionality?

If you're just changing the name, there's no problem, [this post](https://stackoverflow.com/questions/41495658/use-custom-build-output-folder-when-using-create-react-app#54) mentioned about that
"_Therefore, **I'd recommend you to not use a custom folder naming**, if possible. Try to stick with the default naming._"

seanshpark: Thanks for the link!

.gitignore
```diff
@@ -2,3 +2,6 @@ out
 node_modules
 format.patch
 Makefile
+
+# production
```
seanshpark, line 6: as this is for `read` can you revise this to `# production for react-app` or meaningful?

parkjisu6239: thank you for feedback : )

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

PR #113, opened 2021-09-17
DRAFT add lint for css files

seanshpark:
on-going draft to add lint for css files


media/Config/reset.css
```diff
@@ -0,0 +1,30 @@
+html {
```
seanshpark, line 1: from https://github.com/Samsung/ONE-vscode/pull/111/files

src/Visualizer/public/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
seanshpark, line 1: From https://github.com/Samsung/ONE-vscode/pull/115/files

seanshpark:
`htmlhint` doesn't seem to check valid html attributes or extra spaces (usual html does)
but it does catch missing `"` like in
> <meta nameeee="viewport content="width=device-width, initial-scale=1" />

seanshpark:
done

PR #115, opened 2021-09-17
[Visualizer] Add index.html

parkjisu6239:
this file is entry point for react

### DIFF
--- a/src/Visualizer/public/index.html
+++ b/src/Visualizer/public/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta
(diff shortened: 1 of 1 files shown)


src/Visualizer/public/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
seanshpark, line 1: @llFreetimell do we need license comment for html ?

parkjisu6239: I don't think it's necessary.
The basic structure of all HTML files looks like above.

llFreetimell: Only saying conclusion, everything is okay :)

For convention with other files, adding the license comments seems better.
But it is not mandatory because we already have `LICENSE` and `COPYRIGHT` files at the root directory of the repository.

llFreetimell:
@parkjisu6239 Could you add the title to the commit message too? :)

PR #126, opened 2021-09-24
[Draft] Introduce Jsontracer(bar-chart) structure and update features

parkjisu6239:
This draft was created to announce changes in the bar chart and to ask some questions.

## File structure

```
├─.github
│  └─workflows
├─.vscode
├─infra
├─media
│  └─Jsontracer
│       │  index.js
│       │  index.html
│       │  style.css
│       └─ features....js
├─out
│  ├─Codelens
│  ├─Project
│  └─Utils
├─res
│  └─samples
│      └─cfg
└─src
    │  extension.ts
    │  Jsontracer.ts
    │  Project.ts
    │  Utils.ts
    ├─Codelens
    ├─Project
    ├─Snippets
    └─Utils
```

- media/Jsontracer : A static file for webview(html, css, js).
- src/Jsontracer.ts : It contains code and static file import for webview.

## Changed
- I don't use React anymore. Therefore, there are no library issues related to this.
- Static files such as js, css, and html were moved to the media folder.

## Question
- What should the name rule of the js file follow?
    - `PascalCase` or `snake_case` or `camelCase` or etc
- Is the structure of the file okay?

Issue #94, opened 2021-09-16
[Jsontracer] Show profiling json file and bar-chart


parkjisu6239:
## What?
Let's show following profiling json file in vscode extension.

## Sample data
- [sample.profile.json.txt](https://github.com/Samsung/ONE-vscode/files/7175092/sample.profile.json.txt)
- [sample2.profile.json.txt](https://github.com/Samsung/ONE-vscode/files/7175093/sample2.profile.json.txt)

## Preview
![image](https://user-images.githubusercontent.com/42528189/133554365-e6389219-e880-48c9-9a88-3ca31902539f.png)

## Todo
- [x] Register this feature at Command list. #154
- [x] Click `Load`, a file selection dialog will pop up. #159 
- [x] Show timeline (with scroll bar) and bar-chart. #184 
- [x] Show pid`(GPU/CPU) utilization ratio. #170 
- [x] Open or close the panel. #170 
- [x] Make the color a bit prettier. #155
- [x] Zoom in / Zoom out #141 
  - [x] by button
  - [x] by scroll
- [x] Change The scale dynamically as you zoom in or out. #171 
- [x] You can click on the bar to see More info. #188 
  - [x] A single choice.
  - [x] A multi-selection
- [x] All elements follow the vscdoe theme color. #142 
- [ ] Show the node graph as a new web view, If you click `Origin`.
- [ ] Capture 
  - [ ] Capture complete message pop-up
  - [ ] Solve `the Rejected problem` from vcode webview due to security issues
- [x] If the scroll bar moves too fast, the scale will not be deleted or created normally. #141 

## How to run

1. `npm install`
2. `npm run build`
3. Press `F5` to run extension
4. When a new window in vscode opens, press `ctrl + shift + P` to enter `ONE: one-vscode-extension` command.

## Linked draft and Issue
- #93 
- #28 

## Need
It hasn't been PR yet, so if you want to use the demo version, please refer to the draft. :)
After using it, please write in the comments about `the feature you need` or `want to fix`.
We will develop it after reviewing it.

/cc @CODe1995 @yesfordev @parkjisu6239 @donggillee-dev @wnsdud4197
/thanks to @llFreetimell

llFreetimell:
@parkjisu6239 

Could you change all the things in "Done" to "Todo" and add bullets?
Because we did only in draft, not official PR :D

parkjisu6239:
@llFreetimell

Okay, I'll fix it!

[...]

PR #155, opened 2021-09-28
[Jsontracer] Add a color list for the bar background color

parkjisu6239:
It colors the background of the bar graph.

ONE-vscode-DCO-1.0-Signed-off-by: Jisu Park <parkjisu6239@gmail.com>


parkjisu6239:
from draft #126
from issue #94

wnsdud4197 (approved):

media/Jsontracer/colorList.js
```diff
@@ -0,0 +1,54 @@
+const colorList = [
```
llFreetimell, line 1: Can we keep this list with alphabetic order?

parkjisu6239: Or does it mean to rearrange it in alphabetical order?
Since this is assigned from the color in front, it is difficult to do so because it has a visually good color in front.

llFreetimell: > The pretty colors are arranged in the front.

I didn't know this :D Thanks!

parkjisu6239: When the number of nodes is small, only the color in front is shown, so I have a pretty color in front. In other words, it is meaningful in the current order.

image

llFreetimell (approved):
LGTM

PR #159, opened 2021-09-28
[Jsontracer] Add a processData to open the json file data and process it.

parkjisu6239:
This file defines the behavior after the user clicks load button. Open the json file and preprocess it to visualize it.

ONE-vscode-DCO-1.0-Signed-off-by: Jisu Park <parkjisu6239@gmail.com>

### DIFF
--- a/media/Jsontracer/index.js
+++ b/media/Jsontracer/index.js
@@ -124,7 +124,7 @@ function changeSlider(inputValue, inputMax, inputMin) {
 
   // set delay
   slider.disabled = true;
-  setTimeout(() => (slider.disabled = false), 100);
+  setTimeout((slider.disabled = false), 100);
 }
 
 function initData() {
--- a/media/Jsontracer/index.js
+++ b/media/Jsontracer/index.js
@@ -44,7 +44,7 @@
 
 // This file referenced the result of https://github.com/catapult-project/catapult/tree/444aba89e1c30edf348c611a9df79e2376178ba8/tracing
 
-// TODO import openFileSelector from "./processData.js";
+import openFileSelector from "./processData.js";
 // TODO import dynamicGraduation from "./dynamicGraduation.js";
 
 const graph = document.querySelector(".graph");
@@ -54,7 +54,7 @@ const sliderMinLimit = 100;
 const loadBtn = document.querySelector(".load-btn");
 loadBtn.addEventListener("click", () => {
   initData();
-  // TODO openFileSelector();
+  openFileSelector();
 });
 
 let ratio = 100;
@@ -124,7 +124,7 @@ function changeSlider(inputValue, inputMax, inputMin) {
 
   // set delay
   slider.disabled = true;
-  setTimeout((slider.disabled = false), 100);
+  setTimeout(() => (slider.disabled = false), 100);
 }
 
 function initData() {
--- a/media/Jsontracer/processData.js
+++ b/media/Jsontracer/processData.js
@@ -48,7 +48,6 @@
 import getColorList from "./colorList.js";
 
 const setData = document.querySelector(".set-data");
-const colorList = getColorList();
 
 export default function openFileSelector() {
   const input = document.createElement("input");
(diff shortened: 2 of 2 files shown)


parkjisu6239:
from draft #126
from issue #94

media/Jsontracer/index.js
```diff
@@ -124,7 +124,7 @@ function changeSlider(inputValue, inputMax, inputMin) {
 
   // set delay
   slider.disabled = true;
-  setTimeout((slider.disabled = false), 100);
+  setTimeout(() => (slider.disabled = false), 100);
```
llFreetimell, line 127: This change is necessary for this PR?

parkjisu6239: That's not.
It was too small to divide, so I put it in together. There is no error in the operation, but we changed it for uniformity.

media/Jsontracer/processData.js
```diff
@@ -0,0 +1,132 @@
...
+    setData.dataset["displayTimeUnit"] = data.displayTimeUnit;
+  };
+  reader.readAsText(file, "euc-kr");
+}
+
+function processData(data) {
+  const processedData = {};
+  const backgroundColor = {};
+  const utility = {};
+  const colorLen = colorList.length;
```
llFreetimell, line 85: If this is only used in one block, how about moving to the block?
Any maybe `colotList.length` can be used instead of allocating new variable :)

parkjisu6239: > If this is only used in one block, how about moving to the block?

I got it, I will fix it

> Any maybe colotList.length can be used instead of allocating new variable :)

I don't understand what you mean

media/Jsontracer/processData.js
```diff
@@ -0,0 +1,132 @@
...
+      return;
+    }
+
+    // set data object
+    processedData[ele.pid] = processedData[ele.pid]
+      ? processedData[ele.pid]
+      : {};
+    processedData[ele.pid][ele.tid] = processedData[ele.pid][ele.tid]
+      ? processedData[ele.pid][ele.tid]
+      : [];
```
llFreetimell, line 101: How about following?

```suggestion
    if (processedData[ele.pid] === undefined) {
      processedData[ele.pid] = {};
    }
    if (processedData[ele.pid][ele.tid] === undefined) {
      processedData[ele.pid][ele.tid] = [];
    }
```

media/Jsontracer/index.js
llFreetimell: If so, please separate this diff :)
There is no problem that PR is so small :D

media/Jsontracer/processData.js
parkjisu6239: Good! Thank you for the feedback.

media/Jsontracer/index.js
parkjisu6239: Okay! Thank you 👍

media/Jsontracer/processData.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    if (processedData[ele.pid][ele.tid] === undefined) {
+      processedData[ele.pid][ele.tid] = [];
+    }
+
+    // set backgroud-color
+    if (!backgroundColor[ele.name]) {
+      backgroundColor[ele.name] = colorList[colorIdx];
+      colorIdx += 1;
+      colorIdx %= colorLen;
```
llFreetimell, line 107: From above comment, I mean this one!

```suggestion
      colorIdx %= colorList.length;
```

parkjisu6239: I got it! I will change this.
But I'm curious. If we modify it with `colorList.length`, wouldn't it be slower than the existing code by calculating the length every time?
Is there a reason why you want to fix it?

llFreetimell: > wouldn't it be slower than the existing code by calculating the length every time?

 `Array` has a metadata which includes length information too.
Therefore, time complexity of `Array.length` is O(1) which is exactly same with current implementation :)

> Is there a reason why you want to fix it?

Hmm... technically, it is not mandatory.
But usually, we recommends to define the variable according to the scope for preventing mistake and better readability!

parkjisu6239: Oh, I see. Thank you so much! I'll fix it.

llFreetimell (approved):
LGTM

wnsdud4197 (approved):

PR #171, opened 2021-09-29
[Jsontracer] Add dynamic graduation

wnsdud4197:
This will render graduation. Graduation will work dynamically.

ONE-vscode-DCO-1.0-Signed-off-by: Junyeong Yang <wnsdud4197@naver.com>

### DIFF
--- a/media/Jsontracer/dynamicGraduation.js
+++ b/media/Jsontracer/dynamicGraduation.js
@@ -86,11 +86,11 @@ function addGraduation(ruler, cnt) {
     const tempGraduation = document.createElement('div');
     tempGraduation.className = 'graduation';
 
-    for (let i = 0; i < 5; i++) {
+    for (let j = 0; j < 5; j++) {
       const tempSmallGraduation = document.createElement('div');
       tempSmallGraduation.className = 'small-graduation';
 
-      if (i === 0) {
+      if (j === 0) {
         const index = document.createElement('div');
         index.className = 'index';
         tempSmallGraduation.append(index);
@@ -118,9 +118,11 @@ function calculateGraduation(graduation) {
     return Math.round((graduation / 1000) * 10) / 10 + 'ms';
   } else if (graduation >= 1) {
     return Math.round(graduation) + 'us';
+  } else if (graduation > 0) {
+    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
   } else if (graduation === 0) {
     return 0;
   } else {
-    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
+    console.log('[WARNING] graduation is negative');
   }
 }
--- a/media/Jsontracer/dynamicGraduation.js
+++ b/media/Jsontracer/dynamicGraduation.js
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//    * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//    * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//    * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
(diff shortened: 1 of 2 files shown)


wnsdud4197:
from draft #126
from issue #94

parkjisu6239 (dismissed):

media/Jsontracer/dynamicGraduation.js
```diff
@@ -0,0 +1,126 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// This file referenced the result of
+// https://github.com/catapult-project/catapult/tree/444aba89e1c30edf348c611a9df79e2376178ba8/tracing
+
+export default function dynamicGraduation() {
+  const ruler = document.querySelector('.ruler');
+
+  if (!ruler) {
+    return;
```
llFreetimell, line 52: Just curious, is this possible case...?

media/Jsontracer/dynamicGraduation.js
```diff
@@ -0,0 +1,126 @@
...
+  } else {
+    addGraduation(ruler, graduationList.length);
+  }
+  updateGraduation(ruler, timeLimit);
+}
+
+function removeGraduation(ruler, cnt) {
+  for (let i = 0; i < cnt / 2; i++) {
+    const graduation = document.querySelector('.graduation');
```
llFreetimell, line 79: `cnt` is small usually?
The reason why I asked this is because time complexity of `querySelector` is O(n)..!

```suggestion
  const graduation = document.querySelector('.graduation');
  for (let i = 0; i < cnt / 2; i++) {
```

llFreetimell, line 79: `cnt` is small usually?
The reason why I asked this is because time complexity of `querySelector` is O(n)..!

```suggestion
  const graduation = document.querySelector('.graduation');
  for (let i = 0; i < cnt / 2; i++) {
```

media/Jsontracer/dynamicGraduation.js
```diff
@@ -0,0 +1,126 @@
...
+
+function calculateGraduation(graduation) {
+  if (graduation >= 1000) {
+    return Math.round((graduation / 1000) * 10) / 10 + 'ms';
+  } else if (graduation >= 1) {
+    return Math.round(graduation) + 'us';
+  } else if (graduation === 0) {
+    return 0;
+  } else {
+    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
```
llFreetimell, line 124: What is exact condition for this? `0 < graduation and graduation < 1` ?

media/Jsontracer/dynamicGraduation.js
wnsdud4197: If you don't import the file, ruler dose not rendered and press the `Zoom In/Zoom Out` button, it will return.

media/Jsontracer/dynamicGraduation.js
parkjisu6239: If you declare a variable outside the loop statement, only the first one will be deleted and it will not work as desired.

media/Jsontracer/dynamicGraduation.js
wnsdud4197: yes, else condition is `0 < graduation and graduation < 1`.

media/Jsontracer/dynamicGraduation.js
llFreetimell: Ah....! I misunderstood that only one `.graduation` exists ^.T

media/Jsontracer/dynamicGraduation.js
llFreetimell: Then it would be better to revise the condition statements like following.

```javascript
  if (graduation >= 1000) {
    return Math.round((graduation / 1000) * 10) / 10 + 'ms';
  } else if (graduation >= 1) {
    return Math.round(graduation) + 'us';
  } else if (graduation > 0) {
    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
  } else if (graduation === 0) {
    return 0;
  } else {
    // assert here
  }
```

wnsdud4197: graduation always greater than or equal to zero. Then it  would be better like this?
```
  if (graduation >= 1000) {
    return Math.round((graduation / 1000) * 10) / 10 + 'ms';
  } else if (graduation >= 1) {
    return Math.round(graduation) + 'us';
  } else if (graduation > 0) {
    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
  } else if (graduation === 0) {
    return 0;
  }
```

llFreetimell: Just for insurance :D

```javascript
  if (graduation >= 1000) {
    return Math.round((graduation / 1000) * 10) / 10 + 'ms';
  } else if (graduation >= 1) {
    return Math.round(graduation) + 'us';
  } else if (graduation > 0) {
    return Math.round(graduation * 1000 * 10) / 10 + 'ns';
  } else if (graduation === 0) {
    return 0;
  } else {
    console.log('[WARNING] graduation is negative');
  }
```

media/Jsontracer/dynamicGraduation.js
```diff
@@ -0,0 +1,126 @@
...
+function addGraduation(ruler, cnt) {
+  for (let i = 0; i < cnt; i++) {
+    const tempGraduation = document.createElement('div');
+    tempGraduation.className = 'graduation';
+
+    for (let i = 0; i < 5; i++) {
+      const tempSmallGraduation = document.createElement('div');
+      tempSmallGraduation.className = 'small-graduation';
+
+      if (i === 0) {
```
llFreetimell, line 93: There are two `i` in for-loop. Please avoid variable name duplication :)

wnsdud4197: thank you for your feedback, i will fix it

parkjisu6239 (approved):

llFreetimell (approved):
LGTM

Issue #66, opened 2021-09-09
Implement the matching function between json and circle node


CODe1995:
## Pre-step
We finished analyzing the `circle file` structure and identifying the relationship between nodes using typescript. (#36, #46)
* Node relationship analysis was performed with reference to [netron](https://netron.app/).

## Current-step
We match the information of the `json file` generated from Backend to the corresponding node and implement the visualization function.
This function compares the `old json file` with the `new json file` and visualizes it. (Both json files refer to the same circle file)

## Current Progress
![circletracer_2](https://user-images.githubusercontent.com/47562093/135200733-83a19550-d7a5-4d6d-9b0a-0c2b7107ffc3.gif)

## Todo
- [x] Register the command to execute the function #183 
- [x] Read circle file with typescript
   - [x] Add a dependency to use flatbuffers #195 
   - [x] Compile Flatbuffers with typescript #199 
   - [x] Define the type of tensor #207 
   - [x] Analyze builtinOptions #213 
   - [x] Register tensor in subgraph in advance #217 
   - [x] Get the input data corresponding to each operator #209 
   - [x] Get the output data corresponding to each operator #211 
- [x] Analyze the circle file and process the data
- [x] Reads two json files and maps them to processed data
- [x] Visualize the processed data in the form of a node graph
- [x] Display duration next to each node graph

llFreetimell:
@CODe1995

Could you add "What you want to do in this issue" at the main comment? :)
And what you are doing for achieving the goal can be written below main comments :D
(You do not need to fix it now, just FYI)

llFreetimell:
You can use following materials to support this issue :)

- [sample2.profile.json.txt](https://github.com/Samsung/ONE-vscode/files/7096080/sample2.profile.json.txt)
- [sample3.profile.json.txt](https://github.com/Samsung/ONE-vscode/files/7105526/sample3.profile.json.txt)
- [sample.circle.zip](https://github.com/Samsung/ONE-vscode/files/7141331/sample.circle.zip)

llFreetimell:
Please only check the checkbox when related PR is merged :)

llFreetimell:
Please remove `.txt` from the file name.

[simple_add.circle.txt](https://github.com/Samsung/ONE-vscode/files/7282959/simple_add.circle.txt)

Issue #21, opened 2021-08-30
VSCode and Node version?


seanshpark:
### WHAT
Let's discuss and decide program versions that are related with our task.

### WHY
Working on different version may make confusion.

### ALT
Or is `package.json` file enough?

seanshpark:
This is the initial version made from `yo` tool (#18)

```json
  "devDependencies": {
    "@types/vscode": "^1.46.0",
    "@types/glob": "^7.1.3",
    "@types/mocha": "^8.0.4",
    "@types/node": "^12.11.7",
    "eslint": "^7.15.0",
    "@typescript-eslint/eslint-plugin": "^4.9.0",
    "@typescript-eslint/parser": "^4.9.0",
    "glob": "^7.1.6",
    "mocha": "^8.1.3",
    "typescript": "^4.1.2",
    "vscode-test": "^1.4.1"
  }
```

OTKRyu:
i guess if (#18) pr will be successfully merged, everyone do something based on remote repo.
so i think `package.json` is enough.

seanshpark:
For now, we'll use version as below
- "@types/vscode": "^1.46.0",
- @types/node": "^12.11.7",

Version for these two may change in (near) future by any requirements.
- currently there is no requirement as I know.

seanshpark:
From https://github.com/Samsung/ONE-vscode/pull/195#issuecomment-930769225

> I confirmed with the team that there was nothing wrong with npm7. Perhaps it is because the npm version of the format checker is 6.

> Interesting... the format.patch includes reverting all the changes in this PR @.@ (Except package.json)

> This is script to run check format in CI
```
  check-format:
    runs-on: ubuntu-20.04

    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: '12'
```
> node version 12 be the reason?

llFreetimell:
With #202, node-version 16 seems have no problem with our current repository.

If so, my only concern is maintenance of Node.js.
According to https://nodejs.org/ko/about/releases/, following schedule can be checked.

Release | Status | Codename | Initial Release | Active LTS Start | Maintenance LTS Start | End-of-life
-- | -- | -- | -- | -- | -- | --
v12 | Maintenance LTS | Erbium | 2019-04-23 | 2019-10-21 | 2020-11-30 | 2022-04-30
[...]

PR #195, opened 2021-09-30
[Circletracer] add dependency to package

CODe1995:
Add flatbuffers and dependency files.

ONE-vscode-DCO-1.0-Signed-off-by: Seunghyun Kang <code.57x53@gmail.com>


CODe1995:
from draft #179
from issue #66.

package-lock.json
```diff
@@ -1,8 +1,2853 @@
 {
   "name": "one-vscode",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
seanshpark, line 4: why is this changed?

CODe1995: [lockfileVersion]
1: The lockfile version used by npm v5 and v6.
2: The lockfile version used by npm v7, which is backwards compatible to v1 lockfiles.

What version of npm are you using? I'm using v7.12.1

seanshpark:
```
$ npm --version
6.14.10
```

CODe1995: Thanks for checking😀
Even if npm is version 6, `lockfileversion` 2 is [compatible](https://stackoverflow.com/questions/66150813/is-lockfileversion-2-in-package-lock-json-from-npm-7-compatible-with-older-ve).
`lockfileversion` is an option that is automatically converted when using npm v7.
How about using `lockfileversion` to 2?

seanshpark: As I'm not used to `npm` domain, I don't know it's OK or not.
I'll just follow your words.

seanshpark:
Please resolve check-format failure.

CODe1995:
> Please resolve check-format failure.

Checking `format.patch`, there is no problem other than adding the package contents while adding flatbuffers. Are there any other issues I'm missing out on?

seanshpark:
Please check `Details` link and look at the format.patch file

CODe1995:
> Please check `Details` link and look at the format.patch file

I confirmed with the team that there was nothing wrong with npm7. Perhaps it is because the npm version of the format checker is 6.

llFreetimell:
Interesting... the `format.patch` includes reverting all the changes in this PR @.@ (Except `package.json`)

seanshpark:
This is script to run check format in CI
```
  check-format:
    runs-on: ubuntu-20.04

    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: '12'
```
node version 12 be the reason?

seanshpark:
I've reopened #21

CODe1995:
If the npm version is lower, it is recommended to manually update it because there are security issues. In general, there is an npm version that is automatically installed for each nodejs, but we update it manually.

There are two workarounds.

1. Update the format checker's npm
```npm i -g npm```

2. I lower the npm version to version 6 and modify the code as required by the format checker.

what would be better?

seanshpark:
> I lower the npm version to version 6 and modify the code as required by the format checker.
> what would be better?

As to land this change, this would be faster :)

CODe1995:
> As to land this change, this would be faster :)

Done. Thank you :)

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

PR #283, opened 2022-02-14
Import a method rather than 'import *'

hyunsik-yoon:
Previously, `import *` was used and this PR refactors this import statement so that 
only the method used in this file is imported.

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

### DIFF
--- a/src/Circletracer.ts
+++ b/src/Circletracer.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import * as fs from 'fs';
+import {readFileSync} from 'fs';
 import * as vscode from 'vscode';
 
 export class Circletracer {
@@ -147,7 +147,7 @@ export class Circletracer {
 
     // import html
     const htmlPath = this.getMediaPath('index.html');
-    let html = fs.readFileSync(htmlPath.fsPath, {encoding: 'utf-8'});
+    let html = readFileSync(htmlPath.fsPath, {encoding: 'utf-8'});
 
     // Apply js and css to html
     html = html.replace(/_nonce/g, `${nonce}`);


src/Circletracer.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import * as fs from 'fs';
```
mhs4670go, line 17: Note to use asterisk, should we not use the descriptor(got from `as`) fs? Then, it's gonna be hard to know where the method came from.

YongseopKim (commented):
IHMO, it looks good if `Circletracer` doesn't use anything from `fs` in the future. But, if possible, it could be always fixing points whenever adding importing functions from `fs`.

src/Circletracer.ts
hyunsik-yoon: Unfortunately _named import_ such as `import { some_name} from somewhere` does not seem to allow `as`. T.T

https://www.typescriptlang.org/docs/handbook/modules.html#import

hyunsik-yoon:
@YongseopKim,
the change in this PR was actually from my coding style in Python. :-P
So, I tried to find information about Javascript and I found the following: 
- https://stackoverflow.com/questions/47294904/why-shouldnt-you-use-import-all-in-es6
- https://stackoverflow.com/questions/40061042/import-vs-import-specificname-in-typescript-es6
- 
Also, the following is the summary of my survey: 
- The code is clean only with required functions (not all functions) in `import`.
- If a file `abc.ts` contains lots of exported objects and functions, the file size after optimization (by some bundling tool like webpack 2) could better if we use `import { ... } from abc` rather than `import * from abc`. However, if there are not many exported objects and functions, it is OK to use `import *`.

mhs4670go (approved):
LGTM

YongseopKim (approved):
LGTM

PR #284, opened 2022-02-14
Add an execution permission to `infra/format` file

hyunsik-yoon:
This adds an executable permission into `infra/format` file.

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>


seanshpark:
Could you please describe the problem for this change?

hyunsik-yoon:
Oh. After your question, I found `$ bash infra/format` in github action script.)
I am fine with running `$ bash infra/format` without changing permission. :-) 
Just one curiosity. I wonder the reason of using `$ bash infra/format` instead of `$ ./infra/format`.

seanshpark:
> I wonder the reason of using $ bash infra/format instead of $ ./infra/format.

I hoped to lower the possibility for unintended execution.

Issue #285, opened 2022-02-15
[CI] Enable License Checker


llFreetimell:
### What?

If we use [`license-checker`](https://www.npmjs.com/package/license-checker) in `npm`, we can check the license of each packages.
With this tool, let's enable license checker in Github Action and make it easy to check each PR about license :D

### Why?

Some packages may have unacceptable license for ONE.
However, when PRs will be created a lot, it's really hard to verify each PR manually and there may be some mistakes.
To reduce such risk and labor, automatic license checker would be good solution!

### How?

I am planning to make a process as following flow!

![diagrams](https://user-images.githubusercontent.com/24720192/154626868-5eb9f7cc-7118-45bb-ad87-c66719bd959c.png)

- Run on `ubuntu`, `windows` and `macos`
  - Installed packages can be different according to OS
- `PASS` and `FAIL` is the result of self verification, not the result of github action running.
  - This is because we may need to merge the PR even there are some warnings or criticals.
- (Stretch) Find which package installs forbidden packages

### Related PR/Issue

- Draft : #288

### Todo

- [x] Add README.md #295
- [x] Add `license-checker` to `package.json` #335
- [x] Add blocklist for package and licenses #292
- [x] Add script for license verification #323 
- [x] Add script for generating comment with verification result #333
- [x] Add script for applying license verification to Github Action #342
- [x] Bug fix
  - [x] #347 
  - [x] #350

llFreetimell:
### self-diagnosis before this issue is done

/cc @hyunsik-yoon 

```bash
$ cd ONE-vscode
ONE-vscode$ npm install license-checker
ONE-vscode$ npx license-checker
├─ @babel/code-frame@7.12.11
│  ├─ licenses: MIT
│  ├─ repository: https://github.com/babel/babel
│  ├─ publisher: Sebastian McKenzie
│  ├─ email: sebmck@gmail.com
│  ├─ path: /home/seok/HDD/repos/ONE-vscode/node_modules/@babel/code-frame
│  └─ licenseFile: /home/seok/HDD/repos/ONE-vscode/node_modules/@babel/code-frame/LICENSE
├─ @babel/helper-validator-identifier@7.14.9
│  ├─ licenses: MIT
...
ONE-vscode$ npx license-checker --summary
├─ MIT: 207
├─ ISC: 43
├─ BSD-2-Clause: 11
├─ BSD-3-Clause: 9
├─ Apache-2.0: 8
├─ MIT*: 3
[...]

PR #333, opened 2022-03-11
[infra] Add License Verification Result Formatter script

llFreetimell:
This commit adds formatter script for license verification result.

---

Parent Issue : #285
Draft : #288

### DIFF
--- a/infra/license/check-result-formatter.ts
+++ b/infra/license/check-result-formatter.ts
@@ -21,21 +21,21 @@ import {EOL} from 'os';
  * Argument List
  *  - Input paths for each of license verification results
  */
-var args = process.argv.slice(2)
+const args = process.argv.slice(2);
 const resultPaths = args;
 
-var commentBody = [] as string[];
-var totalWarnCount = 0;
-var totalDenialCount = 0;
+let commentBody = [] as string[];
+let totalWarnCount = 0;
+let totalDenialCount = 0;
 for (let i = 0; i < resultPaths.length; ++i) {
   const result = JSON.parse(readFileSync(resultPaths[i], 'utf-8'));
 
-  commentBody.push(result['resultMsg']);
+  commentBody.push(result['resultComment']);
   totalWarnCount += Number(result['warnCount']);
   totalDenialCount += Number(result['denialCount']);
 }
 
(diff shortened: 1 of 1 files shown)


yunjayh (dismissed):
LGTM

infra/license/check-result-formatter.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Argument List
+ *  - Input paths for each of license verification results
+ */
+var args = process.argv.slice(2)
+const resultPaths = args;
+
+var commentBody = [] as string[];
+var totalWarnCount = 0;
+var totalDenialCount = 0;
+for (let i = 0; i < resultPaths.length; ++i) {
```
YongseopKim, line 30: Q. why does this `i` is declared with `let`? 

other variables are `var` or `const`. AFAIK, ES6 recommends using `let` and `const`. 

Can I know your  principal line to distinuish `var` and `let`?

llFreetimell: I used them with following principles.

- `const` : variable whose value should/will not be changed
- `let` : temporary variables which is used in for-loop, while-loop, etc.
- `var` : Other than above

But as you said, it would be better to use `let` instead of `var`!
I will change `var` to `let`, thanks for knowledge :)
And I will create another PR to fix those things in another files :D

YongseopKim: I was just curious. You don't have to fix them. :-) 

We need to have a style guide for ts.

YongseopKim (approved):
LGTM

infra/license/check-result-formatter.ts
llFreetimell: I applied your suggestion because another trivial mistakes are found 😅

YongseopKim (approved):
LGTM

infra/license/check-result-formatter.ts
hyunsik-yoon: I found this.
https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

hyunsik-yoon (approved):

yunjayh (approved):

PR #346, opened 2022-03-14
[CI] Fix license checker on push event

llFreetimell:
`report_verification_result` cannot be run on `push` event.
But current codes did not care about this.
This commit will fix it.


mhs4670go (commented):
LGTM

llFreetimell:
Ah... this PR should be created at upstream -_-
Let me create another PR...

PR #349, opened 2022-03-14
[CI] Add publish-extension workflow

jyoungyun:
This commit adds publish extension workflow.
This workflow builds vsix package and publish to vscode market place.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

### DIFF
--- a/.github/workflows/publish-extension.yml
+++ b/.github/workflows/publish-extension.yml
@@ -1,0 +1,36 @@
+name: Publish extension
+
+on:
+  repository_dispatch:
+    types: [ publish-extension ]
+  schedule:
+    # 05:00 AM (KST) Mon-Fri
+    - cron: '00 20 * * 0-4'
+
+jobs:
+  publish-extension:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+      - name: Install node packages
+        run: |
+          npm install
+      - name: Install vsce tool
+        run: |
+          npm install -g vsce
+      - name: Build vsix package
+        run: |
+          vsce package
+      # TODO Enable publish job
+      - name: Publish vsix package
+        if: ${{ false }}
+        run: |
+          vsce publish
+      - name: Upload vsix package
+        uses: actions/upload-artifact@v3
+        with:
+          name: one-vscode-extension
+          path: ./*.vsix
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "mocha": "^9.2.1",
         "ts-node": "^10.7.0",
         "typescript": "^4.1.2",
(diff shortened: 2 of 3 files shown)


jyoungyun:
Related workflow: https://github.com/Samsung/ONE-vscode/actions/runs/1979356031

.github/workflows/publish-extension.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  publish-extension:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+      - name: Install vsce tool
+        run: npm install -g vsce
```
llFreetimell, line 19: Would it be better to include `vsce` to `devDependencies` in `package.json`?
I am not sure which one is better... @.@

YongseopKim: IMHO, `vsce` is enough in devDependencies not global. (-g option is meaning global, right?)

jyoungyun: > Would it be better to include `vsce` to `devDependencies` in `package.json`?

This tool is only required to build a vsix package. I think the `package.json` file contains the dev tools associated with the source code. So I didn't put it. :)

jyoungyun: > IMHO, `vsce` is enough in devDependencies not global. (-g option is meaning global, right?)

Yes, -g option means global. And this step is working inside docker container. So in fact installing it globally or installing it in the top level `node_module`, both are not affected. :) I think a general tool can be installed as a global package and the packages that have a dependency with code can be installed in their root directory. So this is why I install it globally. :)

jyoungyun: If the script command for `package` in the `package.json` file, I can add this tool to the `package.json` file. But at this point I didn't consider it, so I will consider it again later.

```
"scripts": {
   "package": "vsce package -o one-vscode.vsix",
   "packagePreRelease": "vsce package --pre-release -o one-vscode.vsix"
}
```

llFreetimell: But I put `ts-node` to `package.json` because of `license-checker.ts` and `copyright-checker.ts` :)
Even they are not source code, they are our own codes so I thought that it is related with `package.json`.
And if you don't mind, adding to `package.json` would be effective for administrating licenses of all related packages :D

llFreetimell: Ah I missed your latest comment :(
Then we may go as it is and then discuss about packages which are only needed for special cases :)

jyoungyun: Oh, `vsce` tool should be installed globally. It installs the executable file in `/usr/bin`. So I will revert to the suggested code before.

YongseopKim (approved):
LGTM

llFreetimell (approved):
LGTM

PR #353, opened 2022-03-15
[Project] Fix add() of ToolArgs if value is undefined

YongseopKim:
Let's fix add() of ToolArgs if value is undefined.

### DIFF
--- a/src/Project/ToolArgs.ts
+++ b/src/Project/ToolArgs.ts
@@ -16,8 +16,8 @@
 
 export class ToolArgs extends Array<string> {
   public add(option: string, value: string|undefined) {
+    this.push(option);
     if (value !== undefined) {
-      this.push(option);
       this.push(value);
     }
   }


llFreetimell (approved):
LGTM

hyunsik-yoon:
How about throwing an error when value is `undefined` to decrease possible error?

Note: throughout code, passing one option uses `push`.
https://github.com/Samsung/ONE-vscode/blob/ce23131b5a9827437da8c37f4ea3f8256b4c4fb9/src/Project/JobImportTF.ts#L55-L58

Note: TS shows an error with `add(option)` with one param but passing `undefined` seems not good either.

jyoungyun (commented):
LGTM

jyoungyun (approved):

YongseopKim:
@hyunsik-yoon 

> How about throwing an error when value is undefined to decrease possible coding error?

That's good idea. I'll fix the `add()` after this PR.
- the name `add` seems ambiguous -> `push` seems better while `push` is there on Array<Object>
  - I need to check whether `ts` supports overloading.
- So I would change `add(string, string|undefined)` -> `push(string, ...)` (internally handles by for-loop)

hyunsik-yoon (approved):

YongseopKim:
@Samsung/one-vscode Sorry but I thought wrong. This PR should be reverted. The `add()` code helps handling the option will be pushed or not.

```ts
    // optional arguments
    args.add('--input_arrays', this.inputArrays); // inputArrays could be string or undefined
    args.add('--output_arrays', this.outputArrays);
    args.add('--input_shapes', this.inputShapes);
    args.add('--converter_version', this.converterVersion);
```

PR #354, opened 2022-03-15
[CI] Remove vsce package

jyoungyun:
This package is not meaningful in devDependencies.
It should be installed globally.
So this PR remove the vsce package dependency.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>


jyoungyun:
#349 was merged. Done.

PR #361, opened 2022-03-16
[CI] Use clang-format-14 on Ubuntu 22.04

jyoungyun:
This commit uses clang-format-14 on Ubuntu 22.04.
clang-format-14 supports to run format-check on json files.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #369, opened 2022-03-16
[CI] Fix fetch ref in Github Action

llFreetimell:
Until now, we used `refs/pull/PR_NUMBER/merge` or `github.event.after` to fetch PR's commit.
But it causes unintened errors.
This commit will fix it with `github.event.pull_request.head.sha`.

### DIFF
--- a/.github/workflows/check-commit.yml
+++ b/.github/workflows/check-commit.yml
@@ -20,7 +20,7 @@ jobs:
 
       - run: |
           ROLL_BACK_COUNT=$(( ${{ github.event.pull_request.commits }} - 1 ))
-          git log --format=%B -n 1 ${{ github.event.after }}~${ROLL_BACK_COUNT} > main_commit_msg
+          git log --format=%B -n 1 ${{ github.event.pull_request.head.sha }}~${ROLL_BACK_COUNT} > main_commit_msg
 
       - run: npx ts-node infra/commit/signed-off-by-checker.ts main_commit_msg
 


llFreetimell:
With this PR, I expect to fix `check-signed-off-by` test.

mhs4670go:
It would be more helpful if you let me know what are unintended errors.

jyoungyun:
@mhs4670go 

>It would be more helpful if you let me know what are unintended errors.

This issue is to fix https://github.com/Samsung/ONE-vscode/pull/361#issuecomment-1068848020 comment.

llFreetimell:
@mhs4670go 

I updated PR and only one thing left, changing `github.event.after` to `github.event.pull_request.head.sha` :)
As @jyoungyun said, this problem is started from https://github.com/Samsung/ONE-vscode/pull/361#issuecomment-1068848020.
According to https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pushevent, `after` means "latest commit **after** push event".
Similarly, I guess meaning of "after" is same even the event is not push.
If so, current problem can be explained.
- After we create PR, `github.event.after` is empty
- After we modify/update PR, `github.event.after` is not empty and test passed.

Please let me know if I am wrong :D
Github action is still difficult for me T.T

YongseopKim (approved):

jyoungyun (approved):
LGTM Thanks!

PR #376, opened 2022-03-18
[Tests] Add unittests for Jobs in Project

YongseopKim:
Let's add unittests for Jobs in Project.

### DIFF
--- a/src/Tests/Project/JobCodegen.test.ts
+++ b/src/Tests/Project/JobCodegen.test.ts
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as chai from 'chai';
+import {Job} from '../../Project/Job';
+import {JobCodegen} from '../../Project/JobCodegen';
+
+const assert = chai.assert;
+
+suite('Project', function() {
+  suite('JobCodegen', function() {
+    suite('#contructor()', function() {
+      test('is contructed with jobtype', function() {
+        let job = new JobCodegen();
+        assert.equal(job.jobType, Job.Type.tCodegen);
+      });
+    });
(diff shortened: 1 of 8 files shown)


YongseopKim:
PTAL @Samsung/one-vscode

src/Tests/Project/JobCodegen.test.ts
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as chai from 'chai';
+import {Job} from '../../Project/Job';
+import {JobCodegen} from '../../Project/JobCodegen';
+
+const assert = chai.assert;
```
mhs4670go, line 21: If you only use `assert` method from `chai`, how about import just that method instead of asterisk?

YongseopKim: good point!

src/Tests/Project/JobCodegen.test.ts
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+import {assert} from 'chai';
+import {Job} from '../../Project/Job';
+import {JobCodegen} from '../../Project/JobCodegen';
+
+suite('Project', function() {
+  suite('JobCodegen', function() {
+    suite('#contructor()', function() {
```
mhs4670go, line 23: (optional) I'm not sure but AFAIK `#` has no meaning but for readability. `suite`s are concatenated when they are printed. Then, Is it okay for second suite not to have `#`? like `Project#JobCodegen#construcotr()`. I mean all suite except top level one should have `#` prefix.

YongseopKim: wow, I didn't consider it at all. That's a good point. How about these?

1. Express levels by using multiple `#`
```ts
suite('Project', function() {
  suite('#JobCodegen', function() {
    suite('##contructor()', function() {
```

or

2. Express classes by `@` and functions by `#`
```ts
suite('Project', function() {
  suite('@JobCodegen', function() {
    suite('#contructor()', function() {
```

mhs4670go: Well, expressing level seems unnecessary so, just `Project#JobCodegen#constructor()` seems okay to me. But, since this is what you are working for, anyway looks good as long as it makes better readability.

YongseopKim: If your intention is just to make better readability, that is not necessary.

`npm test` shows like this.
```
  Project
    JobOptimize
      #contructor()
        ✔ is contructed with jobtype
      #valid()
        ✔ returns true always
      #tool()
        ✔ returns toolname as string
      #toolArgs()
        ✔ returns args as ToolArgs
```

If you don't have any opinion on other reasons, I'll keep this as it is.

mhs4670go (approved):
LGTM

hyunsik-yoon (approved):

PR #337, opened 2022-03-11
Add Setting For Model Directory Path

hyunsik-yoon:
This adds an input box into Setting under [ONE] section,
where end-user can enter directory path for model files.
    
for #330

reference: https://code.visualstudio.com/api/references/contribution-points#contributes.configuration

ONE-vscode-DCO-1.0-Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #377, opened 2022-03-18
[Tests] Add unittests for Helpers in Utils

YongseopKim:
Let's add unittests for Helpers in Utils.

### DIFF
--- a/src/Tests/Utils/Helpers.test.ts
+++ b/src/Tests/Utils/Helpers.test.ts
@@ -14,12 +14,10 @@
  * limitations under the License.
  */
 
-import * as chai from 'chai';
-import * as path from 'path';
+import {assert} from 'chai';
+import {join} from 'path';
 
 import {loadCfgFile, obtainWorkspaceRoot} from '../../Utils/Helpers';
-
-const assert = chai.assert;
 
 suite('Utils', function() {
   suite('Helpers', function() {
@@ -33,7 +31,7 @@ suite('Utils', function() {
     suite('#loadCfgFile()', function() {
       test('returns cfgIni object', function() {
         const workspaceRoot: string = obtainWorkspaceRoot();
-        const cfgFile = path.join(workspaceRoot, 'res', 'samples', 'cfg', 'inception_v3.cfg');
+        const cfgFile = join(workspaceRoot, 'res', 'samples', 'cfg', 'inception_v3.cfg');
         const cfgIni = loadCfgFile(cfgFile);
         assert.isNotNull(cfgIni);
         assert.strictEqual(cfgIni['onecc']['one-import-tf'], 'True');
--- a/src/Tests/Utils/Helpers.test.ts
+++ b/src/Tests/Utils/Helpers.test.ts
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as chai from 'chai';
+import * as path from 'path';
+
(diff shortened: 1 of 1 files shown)


hyunsik-yoon (dismissed):
LGTM

src/Tests/Utils/Helpers.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import {assert} from 'chai';
+import {join} from 'path';
+
+import {loadCfgFile, obtainWorkspaceRoot} from '../../Utils/Helpers';
+
+suite('Utils', function() {
+  suite('Helpers', function() {
+    suite('#obtainWorkspaceRoot()', function() {
+      test('returns workspaceRoot as string', function() {
+        const workspaceRoot: string = obtainWorkspaceRoot();
```
mhs4670go, line 26: I have a one question.

`obtainWorkspaceRoot()` looks like below.
```typescript
export function obtainWorkspaceRoot(): string {
  const workspaceFolders = vscode.workspace.workspaceFolders;
  if (!workspaceFolders) {
    console.log('obtainWorkspaceRoot: NO workspaceFolders');
    // TODO revise message
    throw new Error('Need workspace');
  }
```
As code said, if there is no `workspaceFolders`, it throws an error. This throw doesn't happen in this test?

YongseopKim: This is a positive unittest so that I ignore the case.

YongseopKim:
PTAL

hyunsik-yoon:
Thinking again on this PR, current UI design does not use workspace path. 
Instead, it uses project root dir shown at https://github.com/Samsung/ONE-vscode/pull/337.
In this case, it is possible to use our extension without workspace open.
So I am not sure if this test will always succeed.

YongseopKim:
@hyunsik-yoon but, current ONE-vscode is assumpting that workspace is used.

```ts
// Helpers.ts
export function obtainWorkspaceRoot(): string {
  const workspaceFolders = vscode.workspace.workspaceFolders;
  if (!workspaceFolders) {
    console.log('obtainWorkspaceRoot: NO workspaceFolders');
    // TODO revise message
    throw new Error('Need workspace');
  }

  // TODO support active workspace from multiple workspace
  if (workspaceFolders.length > 1) {
    Balloon.info('Warning: only first Workspace is supported');
  }
  const workspaceRoot = workspaceFolders[0].uri.path;
  if (!workspaceRoot) {
    console.log('obtainWorkspaceRoot: NO workspaceRoot');
    // TODO revise message
    throw new Error('Need workspace');
  }
  console.log('obtainWorkspaceRoot:', workspaceRoot);

  return workspaceRoot;
}
```

```ts
// Builders.ts

  // called from user interface
  public build(context: vscode.ExtensionContext) {
    try {
      this.currentWorkspace = helpers.obtainWorkspaceRoot();
    } catch (e: unknown) {
      let errmsg = 'Failed to obtain workspace root';
      if (e instanceof Error) {
        errmsg = e.message;
      }
      // TODO add more type for e if changed in obtainWorkspaceRoot
      Balloon.error(errmsg);
      return;
    }

    this.workFlow.start(this.currentWorkspace);
  }
```

And any features which are supported from ONE-vscode need workspace, I think because it will use some modes in workspace. 

Or I may be confusing btw folder and workspace.

I'll investigate it more.

hyunsik-yoon:
@YongseopKim Current UI works on any workspace.

At the beginning of this year, we thought a workspace containing model files and data files (input, test, etc).
However, the answer of the following questions has not been found yet.
"Does user maintain all model files and data files in one folder?"
"If some user maintains files in different machine, is it OK with them to put all files into one folder?"
(Note there's an assumption that data and model files are sometimes large and some people do not want to maintain those files in local machine)

Writing code with workspace means that user must put all files into one folder. I guess that this answer is not known yet. So current UI of ONE-vscode works on _any_ workspace by setting `project root directory` in setting.
Of course, `project root directory` is currently a local directory but we could _virtually_ add remote files into the folder. 

I guess we could go with `project root directory` for now. If we found that end-user prefers workspace, we could switch to workspace later.

Maybe I could be wrong, any comment will be welcome.

BTW, how to get project root dir is simple:

https://github.com/Samsung/ONE-vscode/blob/c0f5c4f95d44ff1934cb5455f34f2834fd8677eb/src/Views/FileTreeViewActivator.ts#L25-L26

src/Tests/Utils/Helpers.test.ts
YongseopKim: `workspace` should be there so that we can assume that the `workspaceRoot` should be processed properly. It's meaning that it throws error if not and then test would be failed.

mhs4670go (approved):
LGTM

hyunsik-yoon (approved):

YongseopKim:
PTAL @Samsung/one-vscode_committers

PR #378, opened 2022-03-18
[Tests] Add MockupJob for unittests of Jobs

YongseopKim:
Let's add MockupJob for preparing unittests of Jobs such as JobCodegen.

### DIFF
--- a/src/Tests/MockupJob.ts
+++ b/src/Tests/MockupJob.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {JobBase} from '../Project/JobBase';
+import {ToolArgs} from '../Project/ToolArgs';
+
+// Job for mock-up
+export class MockupJob extends JobBase {
+  constructor(name: string) {
+    super();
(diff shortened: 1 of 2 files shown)


llFreetimell (dismissed):
LGTM

src/Tests/MockupJob.ts
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {JobBase} from '../Project/JobBase';
+import {ToolArgs} from '../Project/ToolArgs';
+
+// Job for mock-up
+export class MockupJob extends JobBase {
```
mhs4670go, line 21: (optional) How about just use `Mock` prefix? like `MockJob`?

src/Tests/MockupJob.ts
```diff
@@ -0,0 +1,42 @@
...
+    super();
+    this.name = name;
+  }
+
+  public get valid() {
+    return true;
+  }
+
+  public get tool() {
+    // Just `onecc -h` as dummy
```
mhs4670go, line 32:
```suggestion
    // Just `onecc` as dummy
```
Maybe?

src/Tests/MockupJob.ts
```diff
@@ -0,0 +1,42 @@
...
+// Job for mock-up
+export class MockupJob extends JobBase {
+  constructor(name: string) {
+    super();
+    this.name = name;
+  }
+
+  public get valid() {
+    return true;
+  }
```
mhs4670go, line 29: Could you let me know what it does? It always returns `true`. What is it for?

YongseopKim: It is the JobBase's method so that we should impl it. And also,

```ts
// WorkFlow.ts
  private validateJobs(): boolean {
    let isValid = true;
    this.jobs.forEach((job) => {
      if (!job.valid) {
        Balloon.error('Job ' + job.name + ' is not valid');
        isValid = false;
      }
    });
    return isValid;
  }
```

src/Tests/MockupJob.ts
YongseopKim: good opinion!

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #388, opened 2022-03-21
[Tests] Add unittest for Workflow

YongseopKim:
This commit adds unittest for Workflow.

### DIFF
--- a/src/Tests/Project/Workflow.test.ts
+++ b/src/Tests/Project/Workflow.test.ts
@@ -22,6 +22,11 @@ import {MockJob} from '../MockJob';
 suite('Project', function() {
   suite('WorkFlow', function() {
     const logger = new Logger();
+    // jobs for WorkFlow
+    const name0 = 'job0';
+    const name1 = 'job1';
+    const job0 = new MockJob(name0);
+    const job1 = new MockJob(name1);
     suite('#contructor()', function() {
       test('is constructed as WorkFlow', function() {
         let workFlow = new WorkFlow(logger);
@@ -33,11 +38,7 @@ suite('Project', function() {
     });
     suite('#addJob()', function() {
       test('adds jobs', function() {
-        const name0 = 'job0';
-        const name1 = 'job1';
         let workFlow = new WorkFlow(logger);
-        let job0 = new MockJob(name0);
-        let job1 = new MockJob(name1);
         workFlow.addJob(job0);
         workFlow.addJob(job1);
         assert.strictEqual(workFlow.jobs.length, 2);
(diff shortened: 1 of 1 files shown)


src/Tests/Project/Workflow.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        let job1 = new MockJob(name1);
+        workFlow.addJob(job0);
+        workFlow.addJob(job1);
+        assert.strictEqual(workFlow.jobs.length, 2);
+        assert.strictEqual(workFlow.jobs[0].name, name0);
+        assert.strictEqual(workFlow.jobs[1].name, name1);
+        workFlow.clearJobs();
+        assert.strictEqual(workFlow.jobs.length, 0);
+      });
+    });
```
mhs4670go, line 63: Some codes are duplicated. Actually, there's no problem but it can decrease the score of static analysis tool. If it is not intentional, how about reuse them? Maybe same thing can be applied to other tests.
```suggestion
    const name0 = 'job0';
    const name1 = 'job1';
    let workFlow = new WorkFlow(logger);
    let job0 = new MockJob(name0);
    let job1 = new MockJob(name1);
    workFlow.addJob(job0);
    workFlow.addJob(job1);
    suite('#addJob()', function() {
      test('adds jobs', function() {
        assert.strictEqual(workFlow.jobs.length, 2);
        assert.strictEqual(workFlow.jobs[0].name, name0);
        assert.strictEqual(workFlow.jobs[1].name, name1);
      });
    });
    suite('#clearJobs()', function() {
      test('clears jobs', function() {
        assert.strictEqual(workFlow.jobs.length, 2);
        workFlow.clearJobs();
        assert.strictEqual(workFlow.jobs.length, 0);
      });
    });
```

YongseopKim: I agree with your comment about removing duplicated code. 👍 
But, I think that each test has to handle own each workflow. I'll show it.

src/Tests/Project/Workflow.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        workFlow.addJob(job0);
+        workFlow.addJob(job1);
+        assert.strictEqual(workFlow.jobs.length, 2);
+        assert.strictEqual(workFlow.jobs[0].name, name0);
+        assert.strictEqual(workFlow.jobs[1].name, name1);
+      });
+    });
+    suite('#clearJobs()', function() {
+      test('clears jobs', function() {
+        let workFlow = new WorkFlow(logger);
```
YongseopKim, line 51: @mhs4670go , my intuition is to test each method of each WorkFlow.

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #389, opened 2022-03-21
[Project] Do clearJobs when cfg is imported

### LINKED ISSUE #319: Do clearJobs() of Builder after building jobs
After building jobs, Builder doesn't do clearJobs() so that `import A model config -> build` and then `import B model config -> build` doesn't work. But there is a problem. The building jobs run async. Can Builder get the exact timing when the building jobs is done?

YongseopKim:
Let's do clearJobs when cfg is imported to fix #319.


hyunsik-yoon (approved):

llFreetimell (approved):
LGTM

PR #406, opened 2022-03-25
Add `npm run clean`

YongseopKim:
Let's add npm run clean to remove output directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,8 @@
     "lintcss": "csslint media src",
     "linthtml": "htmlhint media src",
     "unittest": "node ./out/Tests/runTest.js",
-    "test": "npm run unittest"
+    "test": "npm run unittest",
+    "clean": "rm -rf out"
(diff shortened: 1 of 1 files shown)


YongseopKim:
Usage
```
$ npm run clean
```

YongseopKim:
FYI, and now I also found out that some commands including this by introducing me are too based on linux-os. This should be fixed.

mhs4670go (approved):
LGTM

mhs4670go:
> FYI, and now I also found out that some commands including this by introducing me are too based on linux-os. This should be fixed.

There is a npm package called [run-sciprt-os](https://www.npmjs.com/package/run-script-os). I think this can be applied later:)

YongseopKim:
@mhs4670go thanks, I'll try soon! 👍

hyunsik-yoon (approved):

PR #407, opened 2022-03-25
[Tests] Add unittest for ToolRunner

YongseopKim:
Let's add unittest for ToolRunner.

### DIFF
--- a/src/Tests/Project/ToolRunner.test.ts
+++ b/src/Tests/Project/ToolRunner.test.ts
@@ -42,7 +42,6 @@ suite('Project', function() {
         // oneccPath could be string or undefined. Avoid compiling error
         if (oneccPath === undefined) {
           assert.fail('oneccPath should be string type');
-          return;  // for successful ts-compiling
         }
         const workspaceRoot: string = obtainWorkspaceRoot();
         const runner =
--- a/src/Tests/Project/ToolRunner.test.ts
+++ b/src/Tests/Project/ToolRunner.test.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {assert} from 'chai';
+import {join} from 'path';
+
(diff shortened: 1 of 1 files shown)


src/Tests/Project/ToolRunner.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import {MockJob} from '../MockJob';
+
+suite('Project', function() {
+  suite('ToolRunner', function() {
+    const logger = new Logger();
+    let job = new MockJob('mockup');
+    suite('#getOneccPath()', function() {
+      test('returns onecc path as string', function() {
+        let toolRunner = new ToolRunner(logger);
+        let actual = toolRunner.getOneccPath();  // string or undefined
```
hyunsik-yoon, line 32: Shouldn't we also need to check something like the following?
```
assert(actual !== undefined);
```

src/Tests/Project/ToolRunner.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      });
+    });
+    suite('#getRunner()', function() {
+      test('returns runner as Promise<string>', function(done) {
+        let toolRunner = new ToolRunner(logger);
+        const oneccPath = toolRunner.getOneccPath();
+        // oneccPath could be string or undefined. Avoid compiling error
+        if (oneccPath === undefined) {
+          assert.fail('oneccPath should be string type');
+          return;  // for successful ts-compiling
```
mhs4670go, line 45: I removed `return` statement and compiled successfully. Which error have you met?

YongseopKim: Oh, now I know the only conditional expression `oneccPath === undefined` is necessary.

src/Tests/Project/ToolRunner.test.ts
YongseopKim: AFAIK, `===` checks object instance and its value so that we can skip it.

YongseopKim: I mean that `actual === '/usr/share/one/bin/onecc'` checks `actual is not undefined` implicitly.

src/Tests/Project/ToolRunner.test.ts
YongseopKim: Fixed

src/Tests/Project/ToolRunner.test.ts
YongseopKim: Please note me if I'm wrong. (You know, I'm not good at ts/js.)

hyunsik-yoon: I am also ts newbie. :-)

I checked the following and found that `?` does the magic.

```typescript
let bar = (n: number) => {
    if (n % 2 === 0) return undefined; 
    else return 'bar';
} 

let foo = bar(2);
console.log((foo === 'bar') || foo?.endsWith('bar'));   // "undefined" is printed
console.log(foo?.endsWith('bar') || (foo === 'bar'));   // "false" is printed

foo = bar(1);
console.log((foo === 'bar') || foo?.endsWith('bar'));   // 'true' is printed
```

hyunsik-yoon (approved):

mhs4670go (approved):
LGTM

PR #408, opened 2022-03-25
[Tests] Add unittest for JobRunner

YongseopKim:
Let's add unittest for JobRunner.

### DIFF
--- a/src/Tests/Project/JobRunner.test.ts
+++ b/src/Tests/Project/JobRunner.test.ts
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {assert} from 'chai';
+
+import {JobRunner} from '../../Project/JobRunner';
+import {WorkJobs} from '../../Project/WorkJobs';
+import {obtainWorkspaceRoot} from '../../Utils/Helpers';
+import {Logger} from '../../Utils/Logger';
+import {MockJob} from '../MockJob';
+
+suite('Project', function() {
+  suite('JobRunner', function() {
+    const logger = new Logger();
+    suite('#start()', function() {
+      test('jobs are done', function(done) {
+        let jobRunner = new JobRunner(logger);
+        const workspaceRoot: string = obtainWorkspaceRoot();
+        let workJobs = new WorkJobs();
+        workJobs.push(new MockJob('mockup'));
+        workJobs.push(new MockJob('mockup'));
+        assert.strictEqual(workJobs.length, 2);
+
+        // overwrite events so that multiple events will be emitted
+        jobRunner.on('cleanup', function() {
+          assert.strictEqual(workJobs.length, 0);
+          done();
+        });
+        jobRunner.start(workspaceRoot, workJobs);
+      });
+    });
+  });
+});


src/Tests/Project/JobRunner.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      test('jobs are done', function(done) {
+        let jobRunner = new JobRunner(logger);
+        const workspaceRoot: string = obtainWorkspaceRoot();
+        let workJobs = new WorkJobs();
+        workJobs.push(new MockJob('mockup'));
+        workJobs.push(new MockJob('mockup'));
+        assert.strictEqual(workJobs.length, 2);
+
+        // overwrite events so that multiple events will be emitted
+        jobRunner.on('cleanup', function() {
```
mhs4670go, line 38: Maybe?
```suggestion
        jobRunner.on(K_CLEANUP, function() {
```

src/Tests/Project/JobRunner.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          assert.strictEqual(workJobs.length, 0);
+          done();
```
mhs4670go, line 40: Just curiosity, what does `done()` do?

YongseopKim: `done()` is a method to notify that this promiss function is done.

src/Tests/Project/JobRunner.test.ts
YongseopKim: `K_CLEANUP` cannot be accessed from the outer. The only `JobRunner` class is exposed.

YongseopKim: If anyone fixes `K_CLEANUP`, how to do it? Unit tests would not pass so the person should also fix the unit test.

mhs4670go: @YongseopKim 

> `K_CLEANUP` cannot be accessed from the outer. The only `JobRunner` class is exposed.

Ah, now I've become realized that it was.

> If anyone fixes `K_CLEANUP`, how to do it? Unit tests would not pass so the person should also fix the unit test.

That is what I'm worrying about.

@seanshpark I think `K_` consts need to be exported. What do you think about this?

src/Tests/Project/JobRunner.test.ts
mhs4670go: Could you let me know what is promiss function in this context? Becuase IIUC, "promise" that I know is an `Promise` object. 

Well, what I found out is [this section](https://mochajs.org/#asynchronous-code) in mocha official web. 

> By adding an argument (usually named `done`) to `it()` to a test callback, Mocha will know that it should wait for this function to be called to complete the test.

So, `done` is just an aribtrary name of that exmaple and the name also should be the argument of `it()`. It seems the line should be `jobRunner.on('cleanup', function(done) {`. But you posted this PR after passing all tests.. Please let me know if I am wrong.

YongseopKim: `done()` is from `chai`

src/Tests/Project/JobRunner.test.ts
YongseopKim: However, the const vars are used only in the JobRunner. To test these, we need them now. Therefore, I don't think the vars should be exported because I commented above.

> Unit tests would not pass so the person should also fix the unit test.

YongseopKim:
PTAL @mhs4670go

src/Tests/Project/JobRunner.test.ts
mhs4670go: Oh, I missed that,,

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

src/Tests/Project/JobRunner.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    suite('#start()', function() {
+      test('jobs are done', function(done) {
+        let jobRunner = new JobRunner(logger);
+        const workspaceRoot: string = obtainWorkspaceRoot();
+        let workJobs = new WorkJobs();
+        workJobs.push(new MockJob('mockup'));
+        workJobs.push(new MockJob('mockup'));
+        assert.strictEqual(workJobs.length, 2);
+
+        // overwrite events so that multiple events will be emitted
```
dayo09, line 37: It seems this test does several checks within one test,
(1) adding two mockup jobs and cehck if workJobs has length 2
(2) cleaning up the workJob queue and check if workJobs has length 0 as a result

Do I understand alright? If I do, what about separating them into two tests?

YongseopKim: 1. This test is for `JobRunner`.
2. `adding two mockup jobs and cehck` is a sub-process of JobRunner so that I added
3. If new test for adding jobs is introduced, it would be in `WorkJobs`.
4. But, WorkJobs is a Array<Job> so that we don't need to test it.

dayo09: @YongseopKim 

> 2. adding two mockup jobs and cehck is a sub-process of JobRunner so that I added

It sounds to be optional. It seems it aims to test mockUp addition and checking 'cleaning' working. If it's not about JobRunner's test, separating them and renaming as mockup would seems to make this test much clearer, IMO.

YongseopKim: Is it resolved?

dayo09: I looked up the code and it still looks like a testing code for a 'EventEmitter' rather than the JobRunner class itself. I find this test still vague in purpose. I'd rather recommend another test, but as you may. I am not strongly against it.

YongseopKim: If so, let's have time to talk about this test soon. 😁

YongseopKim:
Anyway, 2 approvals. I think it can be merged now. @Samsung/one-vscode_committers

dayo09 (approved):

PR #411, opened 2022-03-25
[Draft] Enable to run tflite model

yunjayh:
This commit enables to run tflite model via `python-shell` package.
Python script from `src/` is copied by `copyfiles` package to `out/`.


yunjayh:
The python code for running model has been excluded with intention. I'll add it after discussion following week.

YongseopKim:
Q. @yunjayh 

> The python code for running model has been excluded with intention. I'll add it after discussion following week.

So, when will you discuss it?

yunjayh:
Oh, the discussion was about the source of python code which is mostly written by @seanshpark, and I discussed with him personally. I'll rebase this pr including the python code.

package.json
```diff
@@ -147,7 +153,9 @@
     "linthtml": "htmlhint media src",
     "unittest": "node ./out/Tests/runTest.js",
     "test": "npm run unittest",
-    "clean": "rm -rf out"
+    "clean": "rm -rf out",
+    "copyfile": "copyfiles --error --up 1 src/**/*.py out/",
+    "postcompile": "npm run copyfile"
```
YongseopKim, line 158: These scripts need to be considered more. That could be
```json
{
  "postcompile": "node ./out/runCopyFiles.js" // runCopyFiles describes what files are copied.
}
```

But, what else way? like ... without copying python files?

src/Runner.ts
```diff
@@ -0,0 +1,24 @@
+/*
```
YongseopKim, line 1: `src/Runner.ts` and `src/Runner/Runner.ts` are too word-iterated. 😁

src/Runner/TFLiteRunner.ts
```diff
@@ -0,0 +1,48 @@
...
+    console.log('TFLiteRunner.runModel()');
+
+    let options = {
+      pythonPath: '/usr/share/one/bin/venv/bin/python',  // TODO: change to docker's
+      pythonOptions: ['-u'],                             // get print results in real-time
+      scriptPath: __dirname,                             // Locates with this file
+      args: [this.modelPath]
+    };
+
+    PythonShell.run('execTflite.py', options, function(err, results) {
```
YongseopKim, line 40: 1. Can you use just `shell` instead of `PythonShell`? (Just curious.)
2. `execTflite.py` would be in ONE? or ONE-vscode?

src/Runner/Runner.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import {Balloon} from '../Utils/Balloon';
+import * as helpers from '../Utils/Helpers';
+import {Logger} from '../Utils/Logger';
+
+const K_MODEL_SELECT: string = 'model-select'
+
+export type modelExt = 'pb'|'tflite'|'onnx';
+
+export class Runner extends EventEmitter implements helpers.FileSelector {
```
YongseopKim, line 28: IHMO, this could be divided to two roles?
1. sth to load model: load models from an event
1. runner: just runs a model loaded

package.json
yunjayh: I haven't considered about this approach. It could make the package more lighter. (right? sorry I'm so newbie in ts/js project)

src/Runner.ts
yunjayh: Totally understand about it.. Moreover, we have `Runner` in `JobRunner`, which makes using `Runner` twice in different ways.. How about renaming the module to `Executor` .. or do you have any idea?

src/Runner/TFLiteRunner.ts
yunjayh: > * Can you use just `shell` instead of `PythonShell`? (Just curious.)

I applied `PythonShell` because it helps to specify the python path.
But, since we are aiming to include docker container for execution environment, `shell` could be better to run docker bash script.. 



> * `execTflite.py` would be in ONE? or ONE-vscode?

In this case maybe only for ONE-vscode.

src/Runner/execTflite.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 1: how about using `execTFLite.py` or `exec_tflite.py` ?

yunjayh: Every filename in this repo is following PascalCase.. I should change to `ExecTFLite.py` in my thought of naming rule. Any reason to use snake case?

seanshpark: `ExecTFLite.py` looks good :)

src/Runner/TFLiteRunner.ts
YongseopKim: I'm not sure `In this case maybe only for ONE-vscode.`. If it is in ONE, users who want to run it are easy for running it and `copyFiles` also is unnecessary. Any reason to place it in ONE-vscode?

YongseopKim: Ah, my word was...
- ONE's toolchains will be supported by docker
- If the `execTflite.py` is supported by ONE, ONE-vscode would use the `execTflite.py` smoothly.

yunjayh: Introducing `execTflite.py` to ONE can satisfy both `ONE` toolchain user and `ONE-vscode` users if they are trying to run value test. 
And, I felt your suggestion is related with https://github.com/Samsung/ONE/issues/8743. Let me and our team think about it more carefully. :smile:

hyunsik-yoon: @yunjayh,
How about creating an issue in ONE repo to discuss this issue?

PR #423, opened 2022-03-29
[Tests] Add `@Use-onecc` suite name

YongseopKim:
This commit adds `@Use-onecc` suite name. Current github action doesn't
install the latest ONE so filter the tests which use onecc.

### DIFF
--- a/src/Tests/README.md
+++ b/src/Tests/README.md
@@ -45,6 +45,18 @@ src/
       ...
 ```
 
+## Misc in tests
+
+Some names are used to filter specific tests.
+
+```ts
+suite('@Use-onecc', function() {
+  // ...
+});
+```
+
+Like `@Use-onecc` is expressing that these tests use `onecc` commands.
+
 ## scripts
 
 There are three scripts for tests.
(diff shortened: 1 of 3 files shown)


YongseopKim:
It looks big but I just add `@Use-onecc`. I considered `clang-format off` but this suite name could be useful, I think.

llFreetimell (approved):
LGTM

src/Tests/README.md
```diff
@@ -45,6 +45,18 @@ src/
...
+
+Some names are used to filter specific tests.
+
+```ts
+suite('@Use-onecc', function() {
+  // ...
+});
+```
+
+Like `@Use-onecc` is expressing that these tests use `onecc` commands.
```
hyunsik-yoon, line 58: I wonder the purpose of this naming. 
It would be great if we add the reason why we need to use this naming. What happens if we don't use this name by mistake?

YongseopKim: Nothing happens. However, if we use this name by mistake when we don't use `onecc`, the test suite would be not tested.

YongseopKim: https://github.com/Samsung/ONE-vscode/pull/424/files#r838176015

hyunsik-yoon:
In future if we have another reason to exclude some tests in CI because, e.g, `Foo`, we have to introduce more names, e.g., `@Use-Foo`. My worry is that it is difficult to remember all cases (onecc,  Foo, ...).
However, I don't know have better answer. 😹

hyunsik-yoon (approved):

Issue #435, opened 2022-03-31
Find out how to handle multiple backends


hyunsik-yoon:
Let's find out how to handle multiple backends.

Requirements:
- There can be multiple backends
- Some backends of some companies could be _confidential_. They might keep their source code in their private repo. ONE-vscode should provide way for such companies to use ONE-vscode to use their backends. For example,
  - ONE-vscode can be able to download their backend toolchain package somewhere.
  - ONE-vscode can be able to install  their backend toolchain package.
  - ONE-vscode can be able to compile, e.g., tflite file, with their compiler.
  - ONE-vscode can be able to run value test, latency test, memory test with their tools.

hyunsik-yoon:
We are testing ONE-vode extension + backend extensions model.
Find out details at #441.

PR #436, opened 2022-03-31
Dynamic loading of backend

hyunsik-yoon:
Let's test if we can load backend dynamically by looking at directory name.

Usage:
- We provide ONE-vscode with ONERT backend only.
- Some other companies can clone our repo and copy their backend modules under `./src/Backends/`. After that, they can use their backend in ONE-vscode.

for #435

### DIFF
--- a/src/Backends/Backend.ts
+++ b/src/Backends/Backend.ts
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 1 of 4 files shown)


hyunsik-yoon:
Log:
```
one-vscode activate OK
NPUBackend: downloaded
NPUBackend: installed
NPUBackend: runValueTest
NPUBackend: runLatencyTest
ONERTBackend: downloaded
ONERTBackend: installed
ONERTBackend: runValueTest
ONERTBackend: runLatencyTest
```

hyunsik-yoon:
Successfully done. let me close.

Issue #331, opened 2022-03-11
Prepare HTML/CSS Template


hyunsik-yoon:
Let's prepare html and css template file.

@Samsung/one-vscode team is not familiar with HTML/CSS and having a template page will help the team.
With this template resource, I hope that other sub-teams would create their web pages in more easier way.

- [x] Design UI/GUI for a specific screen(e.g., compilation screen)
- [ ] Add the screen resource as template

hyunsik-yoon:
Compilation : basic 

![image](https://user-images.githubusercontent.com/63768207/158944648-3d6be21b-c13f-466c-81bc-67f8313ce6ad.png)

hyunsik-yoon:
I found this: https://code.visualstudio.com/blogs/2021/10/11/webview-ui-toolkit

hyunsik-yoon:
working code in draft #297 

![image](https://user-images.githubusercontent.com/63768207/159473810-0c70e55b-d612-4a23-8304-09e8ef65bbb5.png)

I will extract css now.

hyunsik-yoon:
Now #297 has a separate css file.

Time to push PR.

hyunsik-yoon:
UI Animated GIF: [one-vscode.zip](https://github.com/Samsung/ONE-vscode/files/8348851/one-vscode.zip)

hyunsik-yoon:
Let me move to the next milestone and finish submitting PRs.

hyunsik-yoon:
I'd like to share an experience while adding [compile] menu when a file (e.g., tflite) or dir (e.g., saved model) was right-clicked.

My dev environment comprises of vscode on **Windows** and workspace on remote **Ubuntu**.

With my draft with #419 and #422, what I wanted to implement in a draft was to see the possibility of those cases:

1. To enable [Compile] context menu when a file with some extensions (e.g. *.tflite, *.onnx, etc)  is right-clicked
2. Search directories containing saved model or Keras model and put them into some list. Then enable [Compile] contextual menu when one of those directories is right-clicked

For 2. I expected the following vscode context with, e.g., `/tmp/saved_model` directory.
- `resourcePath` === `'/tmp/saved_model'`

However, I found the following in [_Developer Tool_](https://code.visualstudio.com/api/references/when-clause-contexts#inspect-context-keys-utility).
- `resourcePath` === `'\\tmp\\saved_model'`

The path looked like a Ubuntu path but the format was Windows format.
So, the following in draft worked:

```ts
[...]

PR #297, opened 2022-03-04
[DRAFT] Show Compile Screen

hyunsik-yoon:
In this draft, the following will be experimented.

- [x] show icon in ViewContainer (ref: https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome)
- [x] show welcome view (ref: https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome)
- [x] user clicks [Open Folder] button in welcome view
- [x] Let our file tree view shows dir user entered.
- [x] once dir is entered, it needs to be persistently stored.
- [x] User can change dir.

PR #437, opened 2022-04-01
Add basic event handlers for HTML element in [Compile] webview

hyunsik-yoon:
This adds basic event handlers for HTML element in [Compile] webview.

The purpose of this PR is to provide basic way to handle events in
JS and vs extension.

Later, "Compile" modeum needs to enable compilation by ONE by adding
more code into this JS code or `_setWebviewMessageListener` in
`CompilePanel.ts` file.

For https://github.com/Samsung/ONE-vscode/issues/331
Draft https://github.com/Samsung/ONE-vscode/pull/297

### DIFF
--- a/src/Compile/CompilePanel.ts
+++ b/src/Compile/CompilePanel.ts
@@ -140,7 +140,7 @@ export class CompilePanel {
       const text = message.text;
 
       switch (command) {
-        case 'compile-completed':  // Called when [Compile] button was pressed
+        case 'request-compile':  // Called when [Compile] button was pressed
           //
           // TODO Run ONE compile command
           //
--- a/src/Compile/CompilePanel.ts
+++ b/src/Compile/CompilePanel.ts
@@ -132,13 +132,34 @@ export class CompilePanel {
   /**
    * Sets up an event listener to listen for messages passed from the webview context and
    * executes code based on the message that is recieved.
-   *
-   * @param webview A reference to the extension webview
-   * @param context A reference to the extension context
    */
   private _setWebviewMessageListener() {
-    /* NYI */
-    vscode.window.showInformationMessage('NYI');
+    const webview = this._panel.webview;
+    webview.onDidReceiveMessage((message: any) => {
+      const command = message.command;
+      const text = message.text;
+
+      switch (command) {
+        case 'compile-completed':  // Called when [Compile] button was pressed
+          //
+          // TODO Run ONE compile command
+          //
+          vscode.window.showInformationMessage(text);
+          return;
+        case 'set-output-dir':  // Called to set output directory
+          const options = {canSelectMany: false, canSelectFiles: false, canSelectFolders: true};
+
+          vscode.window.showOpenDialog(options).then((val: vscode.Uri[]|undefined) => {
+            if (val !== undefined) {
+              this._panel.webview.postMessage({command: 'set-output-dir', data: val[0].fsPath});
+            }
+            //
+            // TODO handle directory user entered
+            //
+          });
+          return;
+      }
+    }, undefined, this._disposables);
   }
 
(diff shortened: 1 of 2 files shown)


hyunsik-yoon:
Please refer to animated GIF of this PR
--> [one-vscode-pr.zip](https://github.com/Samsung/ONE-vscode/files/8395014/one-vscode-pr.zip)

hyunsik-yoon:
For material of event passing between [webview <--> vs code],
please refer to https://code.visualstudio.com/api/extension-guides/webview#scripts-and-message-passing

src/Compile/CompilePanel.ts
```diff
@@ -132,13 +132,34 @@ export class CompilePanel {
...
   private _setWebviewMessageListener() {
-    /* NYI */
-    vscode.window.showInformationMessage('NYI');
+    const webview = this._panel.webview;
+    webview.onDidReceiveMessage((message: any) => {
+      const command = message.command;
+      const text = message.text;
+
+      switch (command) {
+        case 'compile-completed':  // Called when [Compile] button was pressed
```
jyoungyun, line 143: The `compile-completed` looks like a notification that occurs when compilation is completed. But in this switch phrase, it tries to compile here, so I'm a little confused.

(Q) When is this post message called? After compilation or Trying to compile?

hyunsik-yoon: @jyoungyun,
[Compile] button in webview exits in HTML/JS sandbox.
When the button was clicked in HTML/JS sandbox, HTML/JS sandbox passes 'compile-completed' message to _extension_  (*.ts) sandbox, where the actual compilation by ONE can be done.

JS code that sends a message in this PR:
https://github.com/hyunsik-yoon/ONE-vscode/blob/31757c34798b70ffe1272f028ecddcd10e6e0f34/media/Compile/compile.js#L91

I think JS sandbox should handle UI related stuff. 
It's better to handle actual behavior in _extension_ sandbox because less security limitation and easier to debug with vscode.

jyoungyun: I'm sorry but I don't understand.

I saw the `TODO Run ONE compile command` under this function. So I think this request name would be better to be `request-compile` instead of `compile-completed`. Did I misunderstand what you were saying?

hyunsik-yoon: `request-compile` would be better. I will rename.

src/Compile/CompilePanel.ts
```diff
@@ -132,13 +132,34 @@ export class CompilePanel {
...
+          //
+          vscode.window.showInformationMessage(text);
+          return;
+        case 'set-output-dir':  // Called to set output directory
+          const options = {canSelectMany: false, canSelectFiles: false, canSelectFolders: true};
+
+          vscode.window.showOpenDialog(options).then((val: vscode.Uri[]|undefined) => {
+            if (val !== undefined) {
+              this._panel.webview.postMessage({command: 'set-output-dir', data: val[0].fsPath});
+            }
```
yunjayh, line 155: Q. When `val` is given as `undefined`?

hyunsik-yoon: Actually, I am not sure. :-) 
At first, I thought that user could enter some non-existing path... and such could make `val === undefined`. However, it turned out that vscode UI (`vscode.window.showOpenDialog`) shows error UI when wrong path was entered, making user enter correct path. Therefore `val` should not be `undefined`.

However, I must check if `val !== undefined` in typescript because `val`'s type is `vscode.Uri[]|undefined`. So I added L153.

yunjayh (approved):
LGTM

jyoungyun (approved):
LGTM

PR #475, opened 2022-04-12
Add 'test-no-screen' script

hyunsik-yoon:
This adds `test-no-screen` script in package.json, which is used
when dev environment cannot run vscode GUI. (e.g., Windows connecting
Linux machine via SSH extension).

This is originated from [ .github/workflows/check-pr-format.yml.](https://github.com/Samsung/ONE-vscode/blob/e6e52be227ae83e1622a303646b3aad14047dbdd/.github/workflows/check-pr-test.yml#L18)


yunjayh (approved):
LGTM

Issue #395, opened 2022-03-22
[UX] Memory management visualization


ai-moiseev:
Let's support memory measurement visualization tools.
Parent issue: #309

This feature depends on ONE frontend support and also on backend implementations.

## Required steps

Required external steps:
* [ ] Provide unified format for memory allocation description file (CSV or JSON)
* [ ] Implement memory allocation visualization support in ONE
* [ ] Provide support in backends

ONE-vscode support:
* [ ] Provide means for allocation table generation during one of compilation steps
* [ ] Implement UI for allocation-over-time display (with support for specification of cycle range)

hyunsik-yoon:
Thanks for this work. 

> Provide support in backends

Can you also discuss in the repo of backends you're considering?

ai-moiseev:
> Can you also discuss in the repo of backends you're considering?

Sure, there's already a discussion in backend repository, I've added this item for external progress status tracking :)

dayo09:
@ai-moiseev 

I just checked the spawned issue in ONE repository, it seems you are planning to create `circle` memory visualization.
I hope you note that ONE-vscode targets mostly on npu file. We do plan on extensive support on other file formats(circle, ...) but our initial goal this year is to support npu target models.

ai-moiseev:
@dayo09 The idea was to introduce common format in ONE to provide support for more than one backend, but current focus is on NPU target. I've recently added support for export in JSON format in backend's `compile` command, and I plan to add this new format specification to markdown documentation.

This JSON format comes as an update on `mondrian` CSV files with support for versioning and additional metadata, so this effort is part of #309 (`C Support view for mondrian csv file`).

hyunsik-yoon:
@ai-moiseev 
How about discussing UI you're in mind?
E.g., sharing some sketch could help.

ai-moiseev:
@hyunsik-yoon Here's UI concept which I currently have in mind:
It closely follows output of existing `mondrian` utility, but as addition it provides a way to interactively select cycle range, scroll to begin or end of model, show additional info on mouse hover and switch between memory segments.

![image](https://user-images.githubusercontent.com/67966333/162838886-dde4bbb6-079a-475e-af18-06c6bd587662.png)

PR #483, opened 2022-04-21
[Mondrian] Implement viewer for Mondrian files

ai-moiseev:
This commit implements custom editor which provides graphical view of Mondrian JSON allocation traces (`.tracealloc.json`).

Issue: #395 
Draft: #452 

## Notes

Currently implemented features: 
* JSON input data parsing
* round-robin colorized memory blocks output
* ability to scale horizontally/vertically
* switching between memory segments.

More features such as model trim, mouse hover info, rulers, more coloring modes etc. will be added in new PRs, #395 description contains feature roadmap.

Example data: 
[example.tracealloc.json.txt](https://github.com/Samsung/ONE-vscode/files/8531449/example.tracealloc.json.txt) (remove .txt extension for plugin to properly detect .json)

## Screenshots

Scaled up memory zone view:

image

Model "bird's eye" overview:

image

Light theme support:

image

### DIFF
--- a/media/Mondrian/mondrianViewer.js
+++ b/media/Mondrian/mondrianViewer.js
@@ -124,7 +124,10 @@ function updateContent(data, viewer) {
 
   segmentSelect.replaceChildren();
   for (const [index, segment] of data.segments.entries()) {
-    segmentSelect.appendChild(new Option(segment.name, index));
+    let option = document.createElement('vscode-option');
+    option.innerText = segment.name;
+    option.value = index;
+    segmentSelect.appendChild(option);
   }
   segmentSelect.value = viewer.activeSegment;
 
--- a/media/Mondrian/mondrianViewer.js
+++ b/media/Mondrian/mondrianViewer.js
@@ -209,6 +209,11 @@ function updateViewport(data, viewer) {
     box.style.left = (alloc.alive_from / viewportCycles * 100) + '%';
     box.style.right = ((viewportCycles - alloc.alive_till) / viewportCycles * 100) + '%';
     box.style.backgroundColor = boxColors[i % boxColors.length];
+    box.addEventListener('mouseover', (event) => {
+      statusLineContainer.innerHTML =
+          `<b>Origin:</b> ${alloc.origin} | <b>Size:</b> ${alloc.size} | <b>Offset:</b> ${
+              alloc.offset} | <b>Lifetime:</b> ${alloc.alive_till - alloc.alive_from}`;
+    });
     viewerContainer.appendChild(box);
   }
 }
--- a/media/Mondrian/mondrianViewer.js
+++ b/media/Mondrian/mondrianViewer.js
@@ -1,0 +1,265 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+(function() {
+const vscode = acquireVsCodeApi();
+
+const viewerContainer =
+    /** @type {HTMLElement} */ document.querySelector('.mondrian-viewer-bounds');
+const statusLineContainer =
+    /** @type {HTMLElement} */ document.querySelector('.mondrian-statusline');
+const memorySizeContainer =
+    /** @type {HTMLElement} */ document.querySelector('.mondrian-info-memory-size');
+const cycleCountContainer =
+    /** @type {HTMLElement} */ document.querySelector('.mondrian-info-cycle-count');
+const segmentSelect = /** @type {HTMLElement} */ document.querySelector('.mondrian-segment-picker');
+const viewerHScale = /** @type {HTMLElement} */ document.querySelector('.mondrian-viewer-h-scale');
+const viewerVScale = /** @type {HTMLElement} */ document.querySelector('.mondrian-viewer-v-scale');
+
+class Viewer {
+  constructor() {
+    this.activeSegment = 0;
+    this.viewportMinCycle = 0;
+    this.viewportMaxCycle = 0;
+    this.viewportHScale = 5;
+    this.viewportVScale = 5;
+  }
+}
+
+let viewportMemory = 0;
+
+const boxColors = [
+  '#e25935',
+  '#ee7a0b',
+  '#facb35',
+  '#56571b',
+  '#1791c2',
+  '#5453b1',
+  '#77455e',
+];
+
+// Handle messages sent from the extension to the webview
+window.addEventListener('message', event => {
+  const message = event.data;  // The json data that the extension sent
+  switch (message.type) {
+    case 'update': {
+      const data = parseText(message.text);
+      if (!data) {
+        return;
+      }
+
+      const viewer = new Viewer();
+
+      // Update our webview's content
+      updateContent(data, viewer);
+
+      // Persist state information.
+      // This state is returned in the call to `vscode.getState` below when a webview is reloaded.
+      vscode.setState({data, viewer});
+
+      return;
+    }
+  }
+});
+
+segmentSelect.addEventListener('change', event => {
+  let state = vscode.getState();
+  state.viewer.activeSegment = parseInt(segmentSelect.value);
+
+  updateContent(state.data, state.viewer);
+  vscode.setState(state);
+});
+
+function parseText(/** @type (string) */ text) {
+  if (!text) {
+    text = '{}';
+  }
+  let data;
+
+  /* Parse data JSON */
+  try {
+    data = JSON.parse(text);
+  } catch {
+    statusLineContainer.innerText = 'Error: Document is not a valid JSON';
+    return null;
+  }
+
+  /* Check schema version */
+  if (data.schema_version !== 1) {
+    statusLineContainer.innerText = 'Error: Invalid JSON schema version';
+    return null;
+  }
+
+  /* Check if data has any memory segments */
+  if (data.segments === undefined || data.segments.length === 0) {
+    statusLineContainer.innerText = 'Error: Document has no memory segments';
+    return null;
+  }
+
+  return data;
+}
+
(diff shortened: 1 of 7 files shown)


media/Mondrian/mondrianViewer.js
```diff
@@ -0,0 +1,265 @@
...
+  boxTemplateLabel.classList.add('mondrian-allocation-label');
+
+  boxTemplate.appendChild(boxTemplateLabel);
+
+  viewerContainer.replaceChildren();
+  scaleViewport(viewer);
+
+  const viewportCycles = viewer.viewportMaxCycle - viewer.viewportMinCycle;
+  for (const [i, alloc] of data.segments[viewer.activeSegment].allocations.entries()) {
+    if (alloc.alive_from > viewer.viewportMaxCycle) {
```
llFreetimell, line 199: As I am not good at javascript,  please understand me even I ask some stupid questions :)
( I am glad that you created a good example with custom editor because I am working with custom editor too :D )

---

From line 131, following codes can be found.

```js
  for (alloc of data.segments[viewer.activeSegment].allocations) {
    if (alloc.alive_till > totalCycles) {
      totalCycles = alloc.alive_till;
    }
    // ...
viewer.viewportMaxCycle = totalCycles;
```

Then... is this case can be occurred?

ai-moiseev: Oh... this check was added in advance before implementation of the next feature - the ability to show only a part of the model (large models with 100k+ cycles can slow things down quite a bit... that's the purpose of `viewportMinCycle` / `viewportMaxCycle` properties - this check skips allocations outside of selected range). I haven't implemented it yet (I need to implement interactive range selector first), so I plan to add this feature in the next PR :)

src/Mondrian/MondrianEditor.ts
```diff
@@ -0,0 +1,105 @@
...
+        <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src ${
+        webview.cspSource} data:;
+          style-src ${webview.cspSource}; script-src 'nonce-${nonce}';" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <link href="${styleUri}" rel="stylesheet" />
+        <title>Mondrian Viewer</title>
+      </head>
+      <body>
+        <div class="mondrian-layout">
+          <div class="mondrian-scrollbar"></div>
```
llFreetimell, line 76: I cannot find any information about this div from draft PR :(
I think this is some kind of scrollbar with preview but not sure... is it NYI?

src/Mondrian/MondrianEditor.ts
```diff
@@ -0,0 +1,105 @@
...
+          <div class="mondrian-viewer-area">
+            <div class="mondrian-viewer-bounds"></div>
+          </div>
+          <div class="mondrian-statusbar">
+            <div class="mondrian-statusline"></div>
+            <div class="mondrian-info">
+              <b>Memory:</b> <span class="mondrian-info-memory-size">0</span> |
+              <b>Cycles:</b> <span class="mondrian-info-cycle-count">0</span> |
+              <b>Segment:</b> <select class="mondrian-segment-picker"></select>
```
llFreetimell, line 85: Just FYI, there is `vscode-dropdown` for `select` in VSCode :D

src/Mondrian/MondrianEditor.ts
ai-moiseev: Yes, this is a placeholder :(
I've added NYI note :)

src/Mondrian/MondrianEditor.ts
ai-moiseev: `vscode-dropdown` looks nice! I've added vscode toolkit dependency and replaced `<select>` with `<vscode-dropdown>` :)

hyunsik-yoon:
Could you also add json file into, e.g., `res/sample`? Thanks.

hyunsik-yoon:
@ai-moiseev Thanks for your work!
I am trying to understand the whole flow. (based on https://github.com/Samsung/ONE/issues/8743)

1. vscode runs `onecc` to request memory measurement for backend Z.
2. `onecc` calls some tool defined in backend Z.
3. some tool defined in backend Z produces memory measurement result in format of X.
4. `onecc` produces the result in json format in this PR.
5. vscode renders the json.

Some part of step 1 (`onecc` param), step 2, 3, and 4 are not a scope of _ONE-vscode_ but a scope of _ONE_. 
And, step 5 (this PR, in scope of ONE-vscode) is meaningful only if step 1, 2, 3, and 4 are defined and implemented.
If there's discussion in ONE repo regarding those steps, could you let us know?

ai-moiseev:
@llFreetimell As you know, there was a discussion here: https://github.com/Samsung/ONE/issues/8723
It was stated that support for memory allocation format handling is outside of ONE's scope :( So I've decided to provide implementation in backend and ONE-vscode and to continue discussion on this topic.

My ONE's proposal initially included JSON file processing and conversion into SVG format, but with introduction of custom editor/viewer in ONE-vscode, we only need to add `onecc` compile option. I'll prepare PR for it and try to get it merged :-)

hyunsik-yoon:
@ai-moiseev 
I think there was some gap of approach between our opinions.

What I have been thought was https://github.com/Samsung/ONE-vscode/pull/483#issuecomment-1109210443.

What you've though was (I guess)
- support for memory allocation format handling is outside of ONE's scope

I am OK with your way. I guess that we can have short-term and long-term approach:
- short term: introduce mondrian map of backend in ONE-vscode and backend (not ONE)
- long term: _check_ introduction of mondrian to ONE (ONE is a bridge between ONE-vscode and backends)
  - to align with https://github.com/Samsung/ONE/issues/8743

In this view, I am OK with this PR. :-) Let me know your opinion.

hyunsik-yoon:
I would be helpful to have a explanation of schema of json as, e.g., README.md file. Could you write one in, e.g., `src/mondrian`?

ai-moiseev:
> I would be helpful to have a explanation of schema of json as, e.g., README.md file. Could you write one in, e.g., `src/mondrian`?

I've added data format specification in [json-schema](https://json-schema.org/) format.

hyunsik-yoon:
Oh. Maybe my comment was not complete enough. 😹 
What I am curious is the meaning of each key and value.

E.g., If I want to write json in an editor, 
I have to understand the meaning of "segments", or "size", "allocations", "offset" and so on.
Since this repo is public, anyone who access this repo also needs to understand the meaning of those keys and values.
So could you write down the meaning in README.md ?

If this explanation requires some backend knowledge, you may want to discuss this in [other repo](https://github.sec.samsung.net/RS7-RuntimeNTools/ONE-vscode-TRIV/issues).

```
{
  "schema_version": 1,
  "segments": [
    {
      "name": "Segment1",
      "size": 32768,
      "allocations": [
        { "offset": 0, "size": 8192, "alive_from": 0, "alive_till": 8, "origin": "Node1" },
....
```

ai-moiseev:
@hyunsik-yoon I've updated `README.md` and added text description of all fields, PTAL :) This format is intended to be as backend agnostic as possible (although it's usage limited to backends with memory scheduling; not all backends have precise cycle or timing control, so there's abstract 'lifetime' attribute, which can be relative, which is useful for `luci-micro` backend).

hyunsik-yoon:
@ai-moiseev Thanks for the README.md.

I tried to write down some suggestion or question of my UI experience. 
Some could be done in this PR or in next PR, depending on your opinion. :-)

- Some people may not know what this json file is. So , when they click the json file, how about showing some title like `Memory map over time` and some description on top of graph?
- After opening json file, right-bottom area shows `memory`, `cycle`, and `segment`. I guess `memory` here means total memory required through whole times? If so, how about using `Total memory` label?
  -  When I select `segment 2`, `memory` is 16384 and I am not sure why. From the graph, it has one segment of 8K size. So would 8K be right for `memory`? (So does segment 1)
- I guess X axis would be time (cycle??) and Y axis would be offset but I am not sure. How about showing the meaning of X and Y axis?
- When I maximize the size of vscode to full screen and click the json file, initial graph is shown in small top left area. Maybe when user clicks json, how about fitting the graph to screen size?
- Some button like 'Fit to screen size' would provide good usability.
- In the json file, segment 1 has 24 cycles and segment 2 has 32 cycles. When I enlarge segment 1 to fit all cycle in width of screen and then select segment 2, segment 2 also shows 24 cycles in screen. So I fit 32 cycles of segment 2 to the width of screen. Then I select segment 1 again and found that graph is now shrunken and I had to enlarge again. I hope zoom level of each segment should be kept separately.
- zooming granularity of '+' and '-' seems kinda large, I guess.

ai-moiseev:
@hyunsik-yoon Thank you for a feedback! Yes, explanation UI elements are definitely needed (I planned to add rulers on top and the right and label them as 'time' and 'memory'.

Fit to screen buttons are planned (for X and Y axes) as well as ability to remember zoom and position for each memory segment (I wanted to implement submodel view functionality first and make sure these two features will not conflict with each other).

Zooming step can easily be adjusted :) (now it's 1.0, but could be anything). I'll send update to this PR which will speed it down 2x shortly.

Thank you again for your notes! I would like to quote you in parent issue to keep track of things to implement.

res/samples/traces/sample.tracealloc.json
```diff
@@ -0,0 +1,22 @@
...
+      "allocations": [
+        { "offset": 0, "size": 8192, "alive_from": 0, "alive_till": 8, "origin": "Node1" },
+        { "offset": 8192, "size": 8192, "alive_from": 8, "alive_till": 12, "origin": "Node2" },
+        { "offset": 0, "size": 8192, "alive_from": 12, "alive_till": 16, "origin": "Node3" },
+        { "offset": 8192, "size": 16384, "alive_from": 16, "alive_till": 24, "origin": "Node4" }
+      ]
+    },
+    {
+      "name": "Segment2",
+      "size": 16384,
```
hyunsik-yoon, line 16: Shouldn't this be 8192?
Could you also check whether total memory at Line 6 is correct?

ai-moiseev: I've reduced segment sizes to precisely fit all the blocks. Anyway, the purpose of those samples is to demonstrate syntax rather than real data, I guess, so the values may be a little weird :)

ai-moiseev:
> Some part of step 1 (`onecc` param), step 2, 3, and 4 are not a scope of _ONE-vscode_ but a scope of _ONE_.
> And, step 5 (this PR, in scope of ONE-vscode) is meaningful only if step 1, 2, 3, and 4 are defined and implemented.
> If there's discussion in ONE repo regarding those steps, could you let us know?

@hyunsik-yoon I've created PR which adds `save-allocations` option to `onecc codegen` which should provide a bridge between `ONE-vscode`, `ONE` and backends with allocation trace export support.

hyunsik-yoon (approved):

llFreetimell (approved):
LGTM

PR #489, opened 2022-04-24
[Backend] Introduce Version class

YongseopKim:
Let's introduce Version class.

ONE-vscode-DCO-1.0-Signed-off-by: Yongseop Kim <<yons.kim@samsung.com>>

### DIFF
--- a/src/Tests/Backend/Version.test.ts
+++ b/src/Tests/Backend/Version.test.ts
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


src/Tests/Backend/Version.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import {Version} from '../../Backend/Version';
+
+suite('Backend', function() {
+  suite('Version', function() {
+    suite('#contructor()', function() {
+      test('is contructed with values', function() {
+        const major: number = 1;
+        const minor: number = 0;
+        const patch: number = 2;
+        const option: string = '~220424nightly';
```
jyoungyun, line 27: IMHO, there is no need to use `nightly` postfix on version string.
If the version has `~`, it means that it is a nightly version.

jyoungyun (approved):
LGTM

hyunsik-yoon (approved):

src/Tests/Backend/Version.test.ts
hyunsik-yoon: Note: I wonder if there's any standard (?) versioning scheme for nightly version.
For example, Tensorflow has format like `2.10.0.dev20220424`.
It seems that semantic versioning (https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) accepts only `+` or `-` after `major.min.patch`.
So, I guess that vendors might decide strings after `major.min.patch`, which could be `~`, or `.` (Tensorflow case), or `-` (semantic versioning scheme)

jyoungyun: @hyunsik-yoon 
We had a lot of discussions about the versioning rule. And with the result of discussion, the current rule was selected.
You can see discussions detail [here](https://github.com/Samsung/ONE/issues/7421).

PR #491, opened 2022-04-24
[Project] Take Job as param by invoke()

YongseopKim:
Let's take Job as param by invoke().

ONE-vscode-DCO-1.0-Signed-off-by: Yongseop Kim <<yons.kim@samsung.com>>

### DIFF
--- a/src/Project/ToolRunner.ts
+++ b/src/Project/ToolRunner.ts
@@ -17,6 +17,7 @@
 import * as cp from 'child_process';
 import * as fs from 'fs';
 import {Logger} from '../Utils/Logger';
+import {Job} from './Job';
 import {ToolArgs} from './ToolArgs';
 
 const path = require('path');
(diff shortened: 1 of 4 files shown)


YongseopKim:
I don't yet test this with `onecc`. This is necessary for the test.

src/Project/ToolRunner.ts
```diff
@@ -83,12 +84,10 @@ export class ToolRunner {
     return oneccRealPath;
   }
 
-  public getRunner(name: string, onecc: string, tool: string, toolargs: ToolArgs, path: string) {
+  public getRunner(job: Job, path: string) {
```
hyunsik-yoon, line 87: In `JobRunner.ts`, it is reasonable to change previous code to getRunner(job, path) but
in `ToolRunner.ts`, I am not sure this needs to take `job` as parameter. 
IMHO, I guess `ToolRunner` handles level lower underneath `JobRunner`, and `ToolRunner` is used by `JobRunner` or some other classes which want to run some tool with string params.

YongseopKim: > I guess ToolRunner handles level lower underneath JobRunner, and ToolRunner is used by JobRunner or some other classes which want to run some tool with string params.

That's very reasonable. I'll fix it. And FYI the reason why I changed is to ease make it call the callback function.

YongseopKim: Now I see that fixing `toolrunner` is not necessary. Thanks!

src/Project/ToolRunner.ts
```diff
@@ -83,12 +83,10 @@ export class ToolRunner {
     return oneccRealPath;
   }
 
-  public getRunner(name: string, onecc: string, tool: string, toolargs: ToolArgs, path: string) {
+  public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string) {
```
YongseopKim, line 86: This change is necessary because now this `ToolRunner` is very dependent on `onecc` as a driver.

YongseopKim:
I confirm the tests are passed with onecc.

hyunsik-yoon (approved):
LGTM

jyoungyun (approved):
LGTM

Issue #494, opened 2022-04-25
Add initial version of file view


hyunsik-yoon:
Let's add an initial version of file view.

Features to be implemented are as follows:
- Basic 
  - [x] Show tflite files and their directory #505
  - [x] Show cfg file under tflite file #505
  - [x] Clicking cfg file will open cfg editor panel
- Run onecc
  - [x] #524
    - [x] Clicking 'Run' will run onecc
  - [ ] Newly created tvn file after compilation will show inside view.
- Creating a new cfg
  - [ ] show quickpick for path. User enters path for new cfg
    - Note: cfg in the view is shows under tflite. So path should be entered by user. (default path is the path where tflite is located)
  - [ ] show cfg under tflite and show cfg edit panel of which input tflite path is already entered.

hyunsik-yoon:
/cc @Samsung/one-vscode  @llFreetimell 
Please find the UI for 'Creating a new cfg'. It is a bit _different_ from what we've discussed during online meeting.

@dayo09 is writing `Basic` feature. I believe that some part of `Run onecc` and `Creating a new cfg` can be done. Anyone can join this work. 😸

hyunsik-yoon:
/cc @Samsung/one-vscode  @llFreetimell 
Please find the UI for 'Creating a new cfg'. It is a bit _different_ from what we've discussed during online meeting.

@dayo09 is writing `Basic` feature. I believe that some part of `Run onecc` and `Creating a new cfg` can be done. Anyone can join this work. 😸

dayo09:
@hyunsik-yoon Basic features are now all supported by the draft. I will split them up and make PR.

jyoungyun:
> Please find the UI for 'Creating a new cfg'. It is a bit different from what we discussed during online meeting (4/25 10:30AM).

https://github.sec.samsung.net/RS7-RuntimeNTools/secret/issues/187#issuecomment-1479490

I posted several days ago. I have thought the way to create new cfg file like a gif in my comment.
(Add icons to model file, Click the icon, Create a new file and show the cfg editor.)
Is it not enough?

dayo09:
cc @jyoungyun 
@hyunsik-yoon and I shared an idea offline.
This is a summary.

#### 'Create' : generate config file. path is decided (same directory as model) / name is given by user.
Basically, `create` icon generates cfg file within the same directory. 
Detailed UI follows jyoungyun's idea.

(Sorry for using korean, I attach her origianal comment)
```
[...]

PR #480, opened 2022-04-19
[Draft] [View] Add context explorer view

jyoungyun:
This commit adds context and detail explorers on ONE view container.
The detail explorer is collapsed at the first installed.
Once the user has changed the view state by moving, hiding the view,
the initial state will not be used again.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #495, opened 2022-04-25
DRAFT: Add Context Explorer View

dayo09:
This draft implements context explorer view.

For #494
From #480

### DIFF
--- a/src/contextExplorer.ts
+++ b/src/contextExplorer.ts
@@ -92,7 +92,7 @@ export class ContextTreeDataProvider implements vscode.TreeDataProvider<ContextN
   }
 
   getTreeItem(element: ContextNode): vscode.TreeItem {
-    if (element.node.type === NodeType.directory) {
+    if (element.node.type === NodeType.config) {
       element.command = { command: 'contextExplorer.openConfigFile', title: "Open File", arguments: [element.node] };
     }
     return element;
@@ -205,7 +205,7 @@ export class ContextExplorer {
     vscode.commands.registerCommand('contextExplorer.openConfigFile', (file) => this.openConfigFile(file));
   }
 
-  private openConfigFile(dirnode: DirNode) {
-    vscode.window.showTextDocument(dirnode.uri);
+  private openConfigFile(configNode: ConfigNode) {
+    vscode.window.showTextDocument(configNode.uri);
   }
 }
--- a/src/contextExplorer.ts
+++ b/src/contextExplorer.ts
@@ -79,7 +79,7 @@ export class ContextTreeDataProvider implements vscode.TreeDataProvider<ContextN
   readonly onDidChangeTreeData: vscode.Event<ContextNode|undefined|void> =
       this._onDidChangeTreeData.event;
 
-  private cfgMap: DirNode|undefined;
+  private cfgMap: Node|undefined;
 
   constructor(private workspaceRoot: vscode.Uri|undefined) {
     if (workspaceRoot !== undefined) {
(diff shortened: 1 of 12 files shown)


src/extension.ts
```diff
@@ -53,6 +54,8 @@ export function activate(context: vscode.ExtensionContext) {
 
   setGlobalContext();
 
+  new ContextExplorer(context);
```
jyoungyun, line 57: If we determine the view name, it would be better to change this class and file name.

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+import {config} from 'process';
+import {stringify} from 'querystring';
+import * as vscode from 'vscode';
+import { Model } from './Circlereader/circle-analysis/circle/model';
+
+enum NodeType{
+  directory,
+  model,
+  config,
+  configWrapper
```
jyoungyun, line 28: (Q) Do we need config wrapper directory? How about showing `cfg` file directly without a config wrapper director under the model file?

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+interface Node {
+  type: NodeType, name: string, childs: Node[]
+}
+
+interface DirNode extends Node{
```
jyoungyun, line 35:
```suggestion
interface DirNode extends Node {
```

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+  uri: vscode.Uri
+}
+
+interface ModelNode extends Node {
+  uri: vscode.Uri
+}
+
+interface ConfigNode extends Node {
+  uri: vscode.Uri
+  pairModel: Node
```
jyoungyun, line 45:
```suggestion
  pairModel: ModelNode
```

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+
+export class ContextNode extends vscode.TreeItem {
+  constructor(
+      public readonly label: string,
+      public readonly collapsibleState: vscode.TreeItemCollapsibleState,
+      public readonly node: Node,
+  ) {
+    super(label, collapsibleState);
+
+    this.tooltip = `${this.label}`;
```
jyoungyun, line 59: How about showing full file path like file explorer view?

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+      if (fstat.isDirectory()) {
+        const dirNode = {type: NodeType.directory, name: fn, dir: true, childs: [], uri: vscode.Uri.file(fpath)};
+
+        this.searchConfigs(dirNode, dirpath);
+        if (dirNode.childs.length > 0) {
+          node.childs.push(dirNode);
+        }
+      }
+      else if ((fstat.isFile() 
+        && (fn.endsWith('.tflite')) || fn.endsWith('.onnx'))){
```
jyoungyun, line 149: How about supporting the tensorflow model ?

src/contextExplorer.ts
```diff
@@ -0,0 +1,211 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import {config} from 'process';
+import {stringify} from 'querystring';
+import * as vscode from 'vscode';
+import { Model } from './Circlereader/circle-analysis/circle/model';
```
jyoungyun, line 22:
```suggestion
import * as fs from 'fs';
import * as path from 'path';
import * as vscode from 'vscode';
```

src/contextExplorer.ts
dayo09: Sure. @hyunsik-yoon game me the same opinion, I have removed the wrapper node.

src/contextExplorer.ts
dayo09: Looks like a good idea.

dayo09:
https://github.com/Samsung/ONE-vscode/pull/505 is merged

PR #500, opened 2022-04-26
[UI] Rename Explorer View

dayo09:
This commit renames 'Context Explorer' as 'Explorer'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
     "views": {
       "ONE-Views": [
         {
-          "id": "ContextExplorerView",
-          "name": "Context Explorer"
+          "id": "OneExplorerView",
+          "name": "ONE Explorer"
         },
         {
           "id": "CompilerToolchainView",


hyunsik-yoon:
I think the name is confusing because vscode's view name is Explorer. https://code.visualstudio.com/docs/getstarted/userinterface#_explorer

dayo09:
[hyunsik-yoon](https://github.com/hyunsik-yoon)

Then, how about `ONE explorer` or `Navigator`?

dayo09:
I have separated the deletion of detail explorer view from here, and suggested new 'ONE explorer'

package.json
```diff
@@ -34,13 +34,8 @@
...
         {
-          "id": "ContextExplorerView",
-          "name": "Context Explorer"
-        },
-        {
-          "id": "DetailExplorerView",
-          "name": "Detail Explorer",
-          "visibility": "collapsed"
+          "id": "ExplorerView",
+          "name": "ONE Explorer"
```
jyoungyun, line 38: Um, `ExplorerView` might lead to confusion because vscode already use `Explorer` on the left top view container of vscode. And this is called `fileExplorer` in the code and use this ID when other extension want to attach view on vscode's Explorer.

And when I checked the vscode sample codes, the ID usually starts with lowercase letter.

How about it?
```suggestion
          "id": "oneExplorer",
          "name": "ONE Explorer"
```

dayo09: @jyoungyun I have changed the UI name into ONE Explorer and left the internal class ID as ExplorerView. It concerns only about our internal extension, so I assumed that we don't put 'ONE' there.

dayo09: FYI, this view IDs are only used inside our vscode extension project. :-D

```
      "ONE-Views": [
        {
          "id": "ExplorerView",
          "name": "ONE Explorer"
        },
        {
          "id": "CompilerToolchainView",
          "name": "Compiler Toolchain"
        },
        {
          "id": "ExecutorView",
          "name": "Executor"
        }
      ]
```

jyoungyun: Ah, I misunderstand the way how to add extra view on file explorer. According to vscode sample code, the vscode developer can put the their view to vscode's Explorer using `explorer` ID. [refer](https://github.com/microsoft/vscode-extension-samples/blob/main/tree-view-sample/package.json#L53)

Even though the `ExplorerView` is only used in `ONE-Views`, I think it can still cause confusion. So I prefer to add extra word to id.

However if you want to go as it is, I agree your opinion. :)

dayo09: :thinking:  I understand your concern, I will change it. It would be confusing to call it an explorer even inside our project if we are going to support something about vscode explore.

package.json
```diff
@@ -34,13 +34,8 @@
...
         {
-          "id": "ContextExplorerView",
-          "name": "Context Explorer"
-        },
-        {
-          "id": "DetailExplorerView",
-          "name": "Detail Explorer",
-          "visibility": "collapsed"
+          "id": "OneExplorerView",
+          "name": "ONE Explorer"
```
dayo09, line 38: @jyoungyun

jyoungyun (approved):
LGTM Thank you

dayo09:
@jyoungyun  I have rebased the PR. :-D
@hyunsik-yoon

hyunsik-yoon (approved):
LGTM

jyoungyun (approved):
LGTM

PR #505, opened 2022-04-26
[UI] Implement One Explorer View 

dayo09:
This commit implements One Explorer View.

Co-authored-by: Jiyoung Yun <jy910.yun@samsung.com>

----

From #495 #480
For #494

### DIFF
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -153,7 +153,8 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
    * NOTE It assumes 1-1 relation for model and config
    *
    * TODO(dayo) Support N-N relation
-   * TODO(dayo) Search by parsing config file's model entry
+   * TODO(dayo) Search by parsing config file's model entry (Currently model name and cfg name must
+   * match)
    */
   private searchPairConfig(node: Node, dirPath: string) {
     const dirpath = path.dirname(path.join(dirPath, node.name));
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -25,7 +25,10 @@ enum NodeType{
 }
 
 interface Node {
-  type: NodeType, name: string, childs: Node[], uri: vscode.Uri
+  type: NodeType;
+  name: string;
+  childNodes: Node[];
+  uri: vscode.Uri;
 }
 
 export class OneNode extends vscode.TreeItem {
@@ -89,18 +92,18 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
 
   private getNode(node: Node): OneNode[] {
     const toOneNode = (node: Node): OneNode => {
-      if (node.childs.length > 0) {
+      if (node.childNodes.length > 0) {
         return new OneNode(node.name, vscode.TreeItemCollapsibleState.Collapsed, node);
       } else {
         return new OneNode(node.name, vscode.TreeItemCollapsibleState.None, node);
       }
     };
 
-    return node.childs.map(node => toOneNode(node));
+    return node.childNodes.map(node => toOneNode(node));
   }
 
   private getConfigMap(rootPath: vscode.Uri): Node {
-    const node = {type: NodeType.directory, name: path.parse(rootPath.fsPath).base, dir: true, childs: [], uri: rootPath};
+    const node = {type: NodeType.directory, name: path.parse(rootPath.fsPath).base, dir: true, childNodes: [], uri: rootPath};
     this.searchConfigs(node, path.dirname(rootPath.fsPath));
     return node;
   }
@@ -114,20 +117,20 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       const fstat = fs.statSync(fpath);
 
       if (fstat.isDirectory()) {
-        const dirNode = {type: NodeType.directory, name: fn, dir: true, childs: [], uri: vscode.Uri.file(fpath)};
+        const dirNode = {type: NodeType.directory, name: fn, dir: true, childNodes: [], uri: vscode.Uri.file(fpath)};
 
         this.searchConfigs(dirNode, dirpath);
-        if (dirNode.childs.length > 0) {
-          node.childs.push(dirNode);
+        if (dirNode.childNodes.length > 0) {
+          node.childNodes.push(dirNode);
         }
       }
       else if (fstat.isFile() 
         && (fn.endsWith('.pb') || fn.endsWith('.tflite') || fn.endsWith('.onnx'))) {
-        const modelNode: Node = {type: NodeType.model, name: fn, childs: [], uri: vscode.Uri.file(fpath)};
+        const modelNode: Node = {type: NodeType.model, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
 
         this.searchPairConfig(modelNode, dirpath);
 
-        node.childs.push(modelNode);
+        node.childNodes.push(modelNode);
       }
     }
   }
@@ -157,8 +160,8 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       if (fstat.isFile() 
       && fn.endsWith('.cfg')
       && (extSlicer(fn) === extSlicer(node.name))){
-        const configNode : Node = {type: NodeType.config, name: fn, childs: [], uri: vscode.Uri.file(fpath)};
-        node.childs.push(configNode);
+        const configNode : Node = {type: NodeType.config, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
+        node.childNodes.push(configNode);
       }
     }
   }
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -59,11 +59,11 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   readonly onDidChangeTreeData: vscode.Event<OneNode|undefined|void> =
       this._onDidChangeTreeData.event;
 
-  private cfgMap: Node|undefined;
+  private oneTree: Node|undefined;
 
   constructor(private workspaceRoot: vscode.Uri|undefined) {
     if (workspaceRoot !== undefined) {
-      this.cfgMap = this.getConfigMap(workspaceRoot);
+      this.oneTree = this.getTree(workspaceRoot);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export class OneNode extends vscode.TreeItem {
+  constructor(
+      public readonly label: string,
+      public readonly collapsibleState: vscode.TreeItemCollapsibleState,
+      public readonly node: Node,
+  ) {
+    super(label, collapsibleState);
+
+    this.tooltip = `${this.label}`;
```
dayo09, line 39: How about showing full file path like file explorer view?

_Originally posted by @jyoungyun in https://github.com/Samsung/ONE-vscode/pull/495#discussion_r858158508_

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+      if (fstat.isDirectory()) {
+        const dirNode = {type: NodeType.directory, name: fn, dir: true, childs: [], uri: vscode.Uri.file(fpath)};
+
+        this.searchConfigs(dirNode, dirpath);
+        if (dirNode.childs.length > 0) {
+          node.childs.push(dirNode);
+        }
+      }
+      else if ((fstat.isFile() 
+        && (fn.endsWith('.tflite')) || fn.endsWith('.onnx'))){
```
dayo09, line 125: How about supporting the tensorflow model ?

_Originally posted by @jyoungyun in https://github.com/Samsung/ONE-vscode/pull/495#discussion_r858163175_

dayo09: Done

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+  private searchConfigs(node: Node, dirPath: string) {
+    const dirpath = path.join(dirPath, node.name);
+    const files = fs.readdirSync(dirpath);
+
+    for (const fn of files) {
+      const fpath = path.join(dirpath, fn);
+      const fstat = fs.statSync(fpath);
+
+      if (fstat.isDirectory()) {
+        const dirNode = {type: NodeType.directory, name: fn, dir: true, childs: [], uri: vscode.Uri.file(fpath)};
```
hyunsik-yoon, line 117:
```suggestion
        const dirNode = {type: NodeType.directory, name: fn, dir: true, children: [], uri: vscode.Uri.file(fpath)};
```

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+        this.searchConfigs(dirNode, dirpath);
+        if (dirNode.childs.length > 0) {
+          node.childs.push(dirNode);
+        }
+      }
+      else if (fstat.isFile() 
+        && (fn.endsWith('.pb') || fn.endsWith('.tflite') || fn.endsWith('.onnx'))) {
+        const modelNode: Node = {type: NodeType.model, name: fn, childs: [], uri: vscode.Uri.file(fpath)};
```
hyunsik-yoon, line 126:
```suggestion
        const modelNode: Node = {type: NodeType.model, name: fn, children: [], uri: vscode.Uri.file(fpath)};
```

src/OneExplorer.ts
jyoungyun: You can use the fsPath(?) of node's Uri. 
```suggestion
    this.tooltip = `${this.node.uri.fsPath}`;
```

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+import * as vscode from 'vscode';
+
+enum NodeType{
+  directory,
+  model,
+  config
+}
+
+interface Node {
+  type: NodeType, name: string, childs: Node[], uri: vscode.Uri
```
hyunsik-yoon, line 28:
```suggestion
  type: NodeType, name: string, children: Node[], uri: vscode.Uri
```

src/OneExplorer.ts
dayo09: Done

dayo09:
@hyunsik-yoon May I use `childNodes` instead? The term sounds more familiar

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+export class OneNode extends vscode.TreeItem {
+  constructor(
+      public readonly label: string,
```
hyunsik-yoon, line 33:
```suggestion
      public readonly tooltip: string,
```
`label` can mean many things so it's hard to read code to figure out what's the meaning of label here. But we know what `tooltip` is.

src/OneExplorer.ts
hyunsik-yoon: Why don't you add `TODO` at line 38?

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+        return new OneNode(node.name, vscode.TreeItemCollapsibleState.Collapsed, node);
+      } else {
+        return new OneNode(node.name, vscode.TreeItemCollapsibleState.None, node);
+      }
+    };
+
+    return node.childs.map(node => toOneNode(node));
+  }
+
+  private getConfigMap(rootPath: vscode.Uri): Node {
```
hyunsik-yoon, line 105: 1. I guess that this returns _tree_, not _map_.
2. I guess that this returns tree of not only config but also directory, model, and cfg.

IMHO maybe something like `getTree`(?) would be better name.

src/OneExplorer.ts
jyoungyun: @hyunsik-yoon 
This `label` variable is not same with `tooltip` string. The tooltip string can be changed regardless label string. So I prefer to keep `label` variable. Instead, as you said, it is hard to figure out the meaning, so the name can be changed. How about the name variable? This string means the file or directory name.

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+    const node = {type: NodeType.directory, name: path.parse(rootPath.fsPath).base, dir: true, childs: [], uri: rootPath};
+    this.searchConfigs(node, path.dirname(rootPath.fsPath));
+    return node;
+  }
+
+  private searchConfigs(node: Node, dirPath: string) {
```
hyunsik-yoon, line 111: similar with 2 in https://github.com/Samsung/ONE-vscode/pull/505/files#r858276639.
How about `searchNode`?

dayo09:
@hyunsik-yoon 

> label can mean many things so it's hard to read code to figure out what's the meaning of label here. But we know what tooltip is.

the term 'label' is a member of vscode.TreeItem. It's different from tooltip. tooltip is another member of it.

dayo09:
> similar with 2 in https://github.com/Samsung/ONE-vscode/pull/505/files#r858276639.
How about searchNode?

Thanks for pointing out, I forgot to re-word the 'config's there.

src/OneExplorer.ts
dayo09: > @hyunsik-yoon
> 
> > label can mean many things so it's hard to read code to figure out what's the meaning of label here. But we know what tooltip is.
> 
> the term 'label' is a member of vscode.TreeItem. It's different from tooltip. tooltip is another member of it. `tooltip` defines what to show on a tooltip, I think `label` has different purpose, for example showing the asset label when uri parsing is failed or so.

src/OneExplorer.ts
```diff
@@ -0,0 +1,183 @@
...
+      return fileName.slice(0, fileName.lastIndexOf('.'));
+    };
+
+    for (const fn of files) {
+      const fpath = path.join(dirpath, fn);
+      const fstat = fs.statSync(fpath);
+
+      if (fstat.isFile() 
+      && fn.endsWith('.cfg')
+      && (extSlicer(fn) === extSlicer(node.name))){
```
hyunsik-yoon, line 162: I wonder it assume that cfg file name and model file name are same.

src/OneExplorer.ts
dayo09: I will remove either of `toNode` function or OneNode contstructor, they do the same-purposed jobs in different functions and making confusion.

hyunsik-yoon:
Could you later add test code in other PR? Thanks! :-)

src/OneExplorer.ts
dayo09: I have added 'todo' to parse .cfg file's model file name. I think we need cfg file importer to be implemented for the feature.

dayo09:
@hyunsik-yoon sure :-D

src/OneExplorer.ts
hyunsik-yoon: Could you add more into TODO?

`TODO(dayo) Search by parsing config file's model entry (Currently model name and cfg name must match)`

Without knowing `(Currently model name and cfg name must match)`,
`(extSlicer(fn) === extSlicer(node.name))` was hard to understand since I didn't know your intention. 🐱 

By the way, this approach is quick and reasonable for demo. :-)

dayo09: Applied :-D

src/OneExplorer.ts
dayo09: This TODOs in comments are all resolved.

src/OneExplorer.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  private getTree(rootPath: vscode.Uri): Node {
+    const node = {
+      type: NodeType.directory,
+      name: path.parse(rootPath.fsPath).base,
+      dir: true,
+      childNodes: [],
+      uri: rootPath
+    };
```
jyoungyun, line 111:
```suggestion
    const node: Node = {
      type: NodeType.directory,
      name: path.parse(rootPath.fsPath).base,
      childNodes: [],
      uri: rootPath
    };
```

src/OneExplorer.ts
```diff
@@ -0,0 +1,196 @@
...
+      const fstat = fs.statSync(fpath);
+
+      if (fstat.isDirectory()) {
+        const dirNode = {
+          type: NodeType.directory,
+          name: fn,
+          dir: true,
+          childNodes: [],
+          uri: vscode.Uri.file(fpath)
+        };
```
jyoungyun, line 131:
```suggestion
        const dirNode: Node =
            {type: NodeType.directory, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
```

I have tested `check format` and it changes to this format.

jyoungyun (commented):
Sorry for late feedback.

src/OneExplorer.ts
dayo09: @[jyoungyun](https://github.com/jyoungyun) I have run `infra/format` script locally to resolve our format checker CI. And it has passed the format test.
I think the version of our format tools are different..?

jyoungyun: What I want to say was that the original code didn't explicitly specify the type of `dirNode`, and in my suggestion, I explicitly named `dirNode` as `Node` type.

The `Node` interface does not have a `dir` property, but there is no error because there is no type of `dirNode`.

jyoungyun (approved):
LGTM

hyunsik-yoon (approved):

src/OneExplorer.ts
dayo09: I caught the point and I have applied it. I just wanted to find out our style linter differences :-D

PR #525, opened 2022-04-28
[ONE Explorer] Find model by input_path

dayo09:
This commit changes ONE Explorer's behavior
to search a modelfile specified in a configuration file.
It still assumes they are in the same directory.

### DIFF
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -149,12 +149,6 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
 
   /**
    * Search .cfg files in the same directory of the node
-   *
-   * NOTE It assumes 1-1 relation for model and config
-   *
-   * TODO(dayo) Support N-N relation
-   * TODO(dayo) Search by parsing config file's model entry (Currently model name and cfg name must
-   * match)
    */
   private searchPairConfig(node: Node) {
     const modelPath = node.path;
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { assert } from 'console';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
@@ -27,6 +28,7 @@ enum NodeType {
 interface Node {
   type: NodeType;
   name: string;
+  path: string;
   childNodes: Node[];
   uri: vscode.Uri;
 }
@@ -105,37 +107,41 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
     const node: Node = {
       type: NodeType.directory,
       name: path.parse(rootPath.fsPath).base,
+      path: rootPath.fsPath,
       childNodes: [],
       uri: rootPath
     };
-    this.searchNode(node, path.dirname(rootPath.fsPath));
+    this.searchNode(node);
     return node;
   }
 
-  private searchNode(node: Node, dirPath: string) {
-    const dirpath = path.join(dirPath, node.name);
+  private searchNode(node: Node) {
+    assert(node.type === NodeType.directory);
+
+    const dirpath = node.path;
     const files = fs.readdirSync(dirpath);
 
-    for (const fn of files) {
-      const fpath = path.join(dirpath, fn);
+    for (const fname of files) {
+      const fpath = path.join(dirpath, fname);
       const fstat = fs.statSync(fpath);
 
       if (fstat.isDirectory()) {
         const dirNode: Node =
-            {type: NodeType.directory, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
+            {type: NodeType.directory, name: fname, path: fpath, childNodes: [], uri: vscode.Uri.file(fpath)};
 
-        this.searchNode(dirNode, dirpath);
+        this.searchNode(dirNode);
         if (dirNode.childNodes.length > 0) {
           node.childNodes.push(dirNode);
         }
-      } else if (
-          fstat.isFile() &&
-          (fn.endsWith('.pb') || fn.endsWith('.tflite') || fn.endsWith('.onnx'))) {
+      } 
+      else if 
+      (fstat.isFile() &&
+        (fname.endsWith('.pb') || fname.endsWith('.tflite') || fname.endsWith('.onnx'))) {
+
         const modelNode:
-            Node = {type: NodeType.model, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
+          Node = {type: NodeType.model, name: fname, path: fpath, childNodes: [], uri: vscode.Uri.file(fpath)};
 
-        this.searchPairConfig(modelNode, dirpath);
-
+        this.searchPairConfig(modelNode);
         node.childNodes.push(modelNode);
       }
     }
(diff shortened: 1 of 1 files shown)


src/OneExplorer.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { assert } from 'console';
```
hyunsik-yoon, line 17: How about using https://stackoverflow.com/a/39884185?
strict asserts throws Error when assertion failed.

src/OneExplorer.ts
```diff
@@ -105,67 +107,72 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-      return fileName.slice(0, fileName.lastIndexOf('.'));
+  private searchPairConfig(node: Node) {
+    const modelPath = node.path;
+    const modelName = node.name;
+    const modelExt = path.extname(modelPath).slice(1);
+    
+    const readInputPath = (configPath: string) => {
+      const ini = require('ini');
+      const config = ini.parse(fs.readFileSync(configPath).toString());
+      return config[`one-import-${modelExt}`]['input_path'];
```
hyunsik-yoon, line 161: IMHO we need to handle the case when `config[`one-import-${modelExt}] === undefined`. 😸

src/OneExplorer.ts
```diff
@@ -105,67 +107,72 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-    const extSlicer = (fileName: string) => {
-      return fileName.slice(0, fileName.lastIndexOf('.'));
+  private searchPairConfig(node: Node) {
+    const modelPath = node.path;
+    const modelName = node.name;
+    const modelExt = path.extname(modelPath).slice(1);
+    
+    const readInputPath = (configPath: string) => {
+      const ini = require('ini');
+      const config = ini.parse(fs.readFileSync(configPath).toString());
```
hyunsik-yoon, line 160: I wonder what happens if file contains wrong format of ini file.

src/OneExplorer.ts
```diff
@@ -105,67 +107,72 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-      const fpath = path.join(dirpath, fn);
+      const fpath = path.join(dirPath, fn);
       const fstat = fs.statSync(fpath);
 
-      if (fstat.isFile() && fn.endsWith('.cfg') && (extSlicer(fn) === extSlicer(node.name))) {
-        const configNode:
-            Node = {type: NodeType.config, name: fn, childNodes: [], uri: vscode.Uri.file(fpath)};
-        node.childNodes.push(configNode);
+      if (fstat.isFile() && fn.endsWith('.cfg')) {
+        if (readInputPath(fpath) === modelName) {
```
hyunsik-yoon, line 171: Assuming model name is `"foo.tflite"`,
I guess `readInputPath(fpath)` could return 
1. `"foo.tflite"`
2. `"./foo.tflite"`
3. `"/home/eric/model/inception/foo.tflite"`

maybe its better check basename of return value from `readInputPath(fpath)`.

src/OneExplorer.ts
```diff
@@ -105,67 +107,72 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+          Node = {type: NodeType.model, name: fname, path: fpath, childNodes: [], uri: vscode.Uri.file(fpath)};
 
+        this.searchPairConfig(modelNode);
         node.childNodes.push(modelNode);
       }
     }
   }
 
   /**
    * Search .cfg files in the same directory of the node
```
hyunsik-yoon, line 151: Coudl you add TODO like, `TODO support cfg in different directory` ? 
That is because future reader might be confused with L151, thinking 'is it mandatory to keep tflite and cfg in a same directory? or will this be fixed later?'

src/OneExplorer.ts
```diff
@@ -27,6 +28,7 @@ enum NodeType {
 interface Node {
   type: NodeType;
   name: string;
+  path: string;
```
hyunsik-yoon, line 31: `path` also could be get from `uri.fsPath`.
So if `uri.fsPath` retuns same value with `path`, we may not need to introduce `path`.

dayo09: @hyunsik-yoon I'd thought about that, but I introduced the field to make it explicitly mandatory. I assumed that the original code used uri aside from a container of path, but it doesn't. I'd rather remove uri then :-D.

src/OneExplorer.ts
dayo09: Sure :-D

src/OneExplorer.ts
dayo09: Sure :-D

src/OneExplorer.ts
jyoungyun: `uri` is also used to determine whether this file is a file or directory. And it can be useful to add directory or watch the file changed event. So, I don't want to recommend removing the uri property.

src/OneExplorer.ts
jyoungyun: We need to consider several situations:
- Not defined `one-import-xxx` section
- Not described `input_path` in `one-import-xxx` section
- Relative file path of `input_path`
- The `one-import-xxx` section off status
- Not matched `modelExt` and the described file format on `input_path`
...?

src/OneExplorer.ts
hyunsik-yoon: While looking at many usage of `uri`,
it looks like vscode handles `uri` instead of file path. 
e.g., `export function showOpenDialog(options?: OpenDialogOptions): Thenable<Uri[] | undefined>;` which is defined in `vscode.window` returns `Uri` and `OpenDialogOptions` also gets `Uri` as default path.

vscode Uri has the following comment and I guess that vscode seems to treat local file and remote file with `Uri`.
```
     A universal resource identifier representing either a file on disk
     or another resource, like untitled resources.
```

At the beginning of ONE-vscode, we  also discussed that user might not want to copy a large model file in a _remote_ file server. Not sure how to handle such case but having Uri could help such thing later, I guess.

dayo09:
@hyunsik-yoon @jyoungyun Thanks for your replies. Actually this pr was a bit in a hurry for the demo, and I have rather different draft with more features. As of now we have more time for the demo, I will turn this one into a draft and push more commits. I will take note of your comments and reflect them.

src/OneExplorer.ts
dayo09: Sure :-D

dayo09: Well the thing is, it's right that vscode handles uri instead of its file path. But in this module, we only use 'fspath' property.

> uri is also used to determine whether this file is a file or directory. 

It don't even handle 'wehther this file is a file or directory' its own, it uses `fs` module on 'fspath' property. 
In vscode extension examples, it uses string path instead of uri too.

> vscode Uri has the following comment and I guess that vscode seems to treat local file and remote file with Uri.

Here in this module, I don't use `uri`'s properties nor methods. I tried to look into its methods / properties but it has like 'query' / 'fragment' / 'authority' properties which looks like to  be very much usable in a web.

To make long story short, here we handle with only 'path' information and don't use any useful methods in 'uri', so I'd prefer to use string path.

dayo09: Maybe it's that I don't understand how uri is useful here, but I cannot see its usage as of now. May I use string path and maybe change into uri later when the need is there?

src/OneExplorer.ts
YongseopKim: Yes, please use `assert` from `nodejs`.

src/OneExplorer.ts
YongseopKim: Maybe this can throw some exceptions.

src/OneExplorer.ts
jyoungyun: Okay, I agree with your comments. Let's use file path instead of `vscode.Uri`. I don't know what it will happen when we don't use this, let's think about it later.

jyoungyun:
Please let me know if this pr is ready. :)

src/OneExplorer.ts
dayo09: https://github.com/Samsung/ONE-vscode/pull/543 I implemented Node as a class to resolve this discussion. I think the problem arises as Node manipulates on path more than expected.

dayo09:
Replaced by https://github.com/Samsung/ONE-vscode/pull/554

PR #534, opened 2022-04-29
[OneExplorer] Remove 'oneTree' member

dayo09:
Let's remove 'oneTree' member.
**This change is necessary for a bug fix for refresh command.**
NOTE that OneNodes are generated on the first expansion of tree view.
'oneTree' is not used as getChildren method should returns
freshly generated trees.

### DIFF
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -71,14 +71,14 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   }
 
   getChildren(element?: OneNode): OneNode[]|Thenable<OneNode[]> {
+    if (!this.workspaceRoot) {
+      vscode.window.showInformationMessage('Cannot find workspace root');
+      return Promise.resolve([]);
+    }
+
     if (element) {
       return Promise.resolve(this.getNode(element.node));
     } else {
-      if (!this.workspaceRoot) {
-        vscode.window.showInformationMessage('Cannot find workspace root');
-        return Promise.resolve([]);
-      }
-
       return Promise.resolve(this.getNode(this.getTree(this.workspaceRoot)));
     }
   }
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -57,13 +57,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   readonly onDidChangeTreeData: vscode.Event<OneNode|undefined|void> =
       this._onDidChangeTreeData.event;
 
-  private oneTree: Node|undefined;
-
-  constructor(private workspaceRoot: vscode.Uri|undefined) {
-    if (workspaceRoot !== undefined) {
-      this.oneTree = this.getTree(workspaceRoot);
-    }
-  }
+  constructor(private workspaceRoot: vscode.Uri|undefined) {}
 
   // TODO(dayo): enable refresh command
   refresh(): void {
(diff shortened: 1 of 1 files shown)


src/OneExplorer.ts
```diff
@@ -77,15 +71,15 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-    if (this.oneTree === undefined) {
-      vscode.window.showInformationMessage('No ONE model or config in empty workspace');
-      return Promise.resolve([]);
-    }
-
     if (element) {
       return Promise.resolve(this.getNode(element.node));
     } else {
-      return Promise.resolve(this.getNode(this.oneTree));
+      if (!this.workspaceRoot) {
```
jyoungyun, line 77: How about locating this function outside this `if-else` phrase?
This change does not change the operation, but it's easier to understand the code.

Note.
If the `this.workspaceRoot` is undefined, element is always undefined.

jyoungyun (approved):
LGTM Thank you!

src/OneExplorer.ts
```diff
@@ -77,15 +71,15 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   }
 
   getChildren(element?: OneNode): OneNode[]|Thenable<OneNode[]> {
-    if (this.oneTree === undefined) {
-      vscode.window.showInformationMessage('No ONE model or config in empty workspace');
+    if (!this.workspaceRoot) {
+      vscode.window.showInformationMessage('Cannot find workspace root');
       return Promise.resolve([]);
```
hyunsik-yoon, line 76: (not a part of this PR) but would `Promise.reject()` be correct?

hyunsik-yoon:
I wonder if the approach would fast enough if user has many files, e.g., about 20 models and 4 cfg files per model.

dayo09:
@hyunsik-yoon It doens't change the performance. The number of tree-search operator is the same. The original code makes bug on refresh command(refresh not working).

The child nodes are generated on the first whole-tree generation(getTree). 
getChildren is called when you first expand the node and use the nodes generated on getTree. (it converts nodes into oneNode)
Refresh command expects that getChildren with an empty argument to get the fresh tree, but it just remembers oneTree before. This commit fixes it.

jyoungyun:
> I wonder if the approach would fast enough if user has many files, e.g., about 20 models and 4 cfg files per model.

It takes some minutes at the first loading time, but once it is loaded, there will no issues of speed until pressing Refresh button.
Modules that take longer time to load in the first loading time of other extension modules are showing blue progress bar. And I think this operation will work similarly in our Explorer view.

src/OneExplorer.ts
dayo09: image

jyoungyun: Please refer to this https://github.com/microsoft/vscode-extension-samples/blob/main/tree-view-sample/src/nodeDependencies.ts#L22-L25 codes.

I think we don't need to show error message. Instead we can show the information noti for this one.

hyunsik-yoon:
@dayo09 @jyoungyun Thanks for explanation about `getChildren()`. 😸

hyunsik-yoon (approved):

src/OneExplorer.ts
dayo09: Yes, I took the same reference. Both cases looks fine. You can choose. Or maybe we can discuss another time.

PR #539, opened 2022-04-29
Renaming params of ToolRunner.getRunner()

hyunsik-yoon:
Some params were renamed to have more specific meaning.

### DIFF
--- a/src/Project/ToolRunner.ts
+++ b/src/Project/ToolRunner.ts
@@ -83,10 +83,10 @@ export class ToolRunner {
     return oneccRealPath;
   }
 
-  public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string) {
+  public getRunner(name: string, toolPath: string, toolargs: ToolArgs, cwd: string) {
     return new Promise<string>((resolve, reject) => {
       this.logger.outputWithTime('Running: ' + name);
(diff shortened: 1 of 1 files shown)


YongseopKim:
Hmm just tool is enough to run anything, I think. 
For example, onecc can be run. And like 'apt-get' doesnt have path.

jyoungyun (approved):
LGTM

hyunsik-yoon:
@YongseopKim 

In the previous code,
```
  public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string) {
```

When I tried to use this method for the first time, I assumed that `tool` could be _tool name_ but tool name was already provided as the first param. So I was confused the meaning of this second param. 

Maybe we can rename `tool` to `command` but I thought `toolPath` would be better in `ToolManager`.

The second param is passed to `child_process.spawn()` and `spawn()`'s doc mentioned. :
```
command - The command to run.
```

Anywa, if we prefer `command`, I will rename `toolPath` to `command`.

YongseopKim:
> When I tried to use this method for the first time, I assumed that tool could be tool name but tool name was already provided as the first param. So I was confused the meaning of this second param.

If you mention about `name`, it is just name of `Job`. That could be the name of the tool and also the name with description like `tool where A case`. Therefore, the code doesn't use the `name` as running commands. The code just uses it as a debug log.

PR #544, opened 2022-05-02
[UI]Change Executor view title contents

struss:
This commit will change view name and add some icon on this.

ONE-vscode-DCO-1.0-Signed-off-by: struss rrstrous@nate.com

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
         },
         {
           "id": "TargetDevice",
-          "name": "Target Device"
+          "name": "Target Devices"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
           "name": "Compiler Toolchain"
         },
         {
-          "id": "ExecutorView",
-          "name": "Executor"
+          "id": "TargetDevice",
+          "name": "Target Device"
         }
       ]
     },
@@ -98,6 +98,12 @@
         "icon": "$(desktop-download)"
       },
       {
+        "command": "onevscode.connect-device",
+        "title": "connect",
+        "category": "ONE",
+        "icon": "$(add)"
+      },
+      {
         "command": "onevscode.build",
         "title": "build",
         "category": "ONE"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,8 +43,8 @@
           "name": "Compiler Toolchain"
         },
         {
-          "id": "ExecutorView",
-          "name": "Executor"
+          "id": "TargetDevice",
```
hyunsik-yoon, line 46:
```suggestion
          "id": "TargetDeviceView",
```

package.json
```diff
@@ -43,8 +43,8 @@
...
+          "name": "Target Device"
```
hyunsik-yoon, line 47:
```suggestion
          "name": "Target Devices"
```

I looked into other extensions and they used plural noun.

src/extension.ts
```diff
@@ -67,6 +67,12 @@ export function activate(context: vscode.ExtensionContext) {
   });
   context.subscriptions.push(installCompiler);
 
+  // Target Device view
+  let connectDevice = vscode.commands.registerCommand('onevscode.connect-device', () => {
```
hyunsik-yoon, line 71: Can I ask a question about the naming `connect`?

Is it a command when, e.g, `$ adb connect` is called?

While a user see a target device in a Target Device View,
internally the device can be connected and disconnected.

So I wonder this command is for such connection.

struss: > Is it a command when, e.g, `$ adb connect` is called?

I think it will call `$ adb connect` once to check device is connectable.

> While a user see a target device in a Target Device View,
> internally the device can be connected and disconnected.

but, this case would be right. so I think `register` would be more right for this.
`Register` this device as a `Target Device` on user or one-vscode configuration.

so, what about change this into `register`?

hyunsik-yoon (approved):
LGTM, thanks. 😸

package.json
```diff
@@ -97,6 +97,12 @@
         "category": "ONE",
         "icon": "$(desktop-download)"
       },
+      {
+        "command": "onevscode.register-device",
```
jyoungyun, line 101: (Suggestion) How about using `connect-device`?

jyoungyun (approved):
LGTM

PR #549, opened 2022-05-03
[Backend] Introduce `getExecutableExt` method to executor

yunjayh:
This commit introduces `getExecutableExt` method which returns executable
extension list to Executor class.

---
draft: #522

### DIFF
--- a/src/Backend/Executor.ts
+++ b/src/Backend/Executor.ts
@@ -18,6 +18,9 @@ import {Command} from './Command';
 import {Toolchains} from './Toolchain';
 
 interface Executor {
+  // exetensions of executable files
+  getExecutableExt(): string[];
+
   // defined/available toolchains
   toolchains(): Toolchains;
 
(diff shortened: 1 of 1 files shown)


jyoungyun (approved):
LGTM

src/Backend/Executor.ts
```diff
@@ -18,6 +18,9 @@ import {Command} from './Command';
 import {Toolchains} from './Toolchain';
 
 interface Executor {
+  // exetensions of executable files
+  getExecutableExt(): string[];
```
YongseopKim, line 22: Like ['tflite', 'onxx']?

YongseopKim (approved):
LGTM

src/Backend/Executor.ts
yunjayh: Exactly. IMO, After one-infer implementaion, every ONE toolchain might have tflite and onnx as its exacutable extension.

PR #553, opened 2022-05-03
[OneExplorer] Parse .cfg and display on tree

dayo09:
This commit parses config and link all specified model nodes.


src/OneExplorer.ts
```diff
@@ -152,29 +177,131 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+      {section: 'one-optimize', key: 'input_path', grepper: this.grepTarget},
+      {section: 'one-quantize', key: 'input_path', grepper: this.grepTarget},
+      {section: 'one-quantize', key: 'input_path', grepper: this.grepTarget},
+      {section: 'one-codegen', key: 'command', grepper: this.grepTargetInCommand},
+    ];
+
+    let intermediates: string[] = [];
+    for (let loc of targetLocator) {
+      let confSection = config[loc.section as keyof typeof config];
+      let confKey = confSection ? confSection[loc.key as keyof typeof config] : undefined;
```
llFreetimell, line 232: (Optional) Javascript/Typescript has "optional chaining" :D

```suggestion
      let confKey = confSection?.[loc.key as keyof typeof config];
```

src/OneExplorer.ts
```diff
@@ -152,29 +177,131 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    const config = readIni(configPath);
+    const ext = path.extname(modelPath).slice(1);
+
+    if (ext === undefined || config === null) {
+      return undefined;
+    }
+
+    return config[`one-import-${ext}` as keyof typeof config] ?
+        config[`one-import-${ext}` as keyof typeof config]['input_path'] :
+        undefined;
```
llFreetimell, line 191: (Optional) ditto :D

```suggestion
      let confKey = confSection?.[loc.key as keyof typeof config];
```

```suggestion
    return config[`one-import-${ext}` as keyof typeof config]?.['input_path'];
```

llFreetimell, line 191: (Optional) ditto :D

```suggestion
    return config[`one-import-${ext}` as keyof typeof config]?.['input_path'];
```

src/OneExplorer.ts
```diff
@@ -15,13 +15,34 @@
...
+  try {
+    configRaw = fs.readFileSync(filePath, 'utf-8');
+  } catch (e) {
+    console.error(e);
+    return null;
+  }
+
+  // TODO check if toString() is required
+  return ini.parse(configRaw.toString());
+}
```
yunjayh, line 39: https://github.com/Samsung/ONE-vscode/blob/7eeb7485e08af5642001fb088fbe4c29602f038c/src/Utils/Helpers.ts#L79-L86

It seems that there is a helper which do the same (or pretty similar) usage.

src/OneExplorer.ts
```diff
@@ -152,29 +177,131 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+
+  // TODO(dayo) extract file-relative functions as another module
+  private grepTargetInCommand = (str: string): string[] => {
+    let targets: string[] = [];
+    for (let entry of str.split(' ')) {
+      if (path.extname(entry) === '.tvn' || path.extname(entry) === '.circle') {
+        targets.push(entry);
+      }
+    }
+    return targets;
```
jyoungyun, line 202:
```suggestion
    return str.split(' ').filter(e => path.extname(e) === '.tvn' || path.extname(e) === '.circle');
```

src/OneExplorer.ts
```diff
@@ -106,9 +129,11 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       if (node.type === NodeType.directory) {
         return new OneNode(node.name, vscode.TreeItemCollapsibleState.Expanded, node);
       } else if (node.type === NodeType.model) {
+        return new OneNode(node.name, vscode.TreeItemCollapsibleState.None, node);
```
hyunsik-yoon, line 132: Can we expand all nodes by default?
Every time when I finds cfg file, I have to expand tflite file by clicking the tflite file. By removing those clicking it would provide better usability.

src/OneExplorer.ts
```diff
@@ -152,29 +177,131 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    const files = fs.readdirSync(node.parent);
 
     for (const fname of files) {
       const fpath = path.join(node.parent, fname);
       const fstat = fs.statSync(fpath);
 
-      if (fstat.isFile() && fname.endsWith('.cfg') && (extSlicer(fname) === extSlicer(node.name))) {
-        const pairNode = new Node(NodeType.config, [], vscode.Uri.file(fpath));
-        node.childNodes.push(pairNode);
+      if (fstat.isFile() && (fname.endsWith('.circle') || fname.endsWith('.tvn'))) {
```
hyunsik-yoon, line 296: File extention could be different from backend to backend. Rather than using string `.tvn`, it would be better to use some backend API to get file extention string by asking backend API.
Please refer to the approach in #549.

src/OneExplorer.ts
```diff
@@ -152,29 +177,131 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    }
+    return targets;
+  };
+
+  // TODO(dayo) extract file-relative functions as another module
+  private grepTarget = (str: string): string[] => {
+    let targets: string[] = [];
+    // TODO(dayo) add checks
+    targets.push(str);
+    return targets;
```
hyunsik-yoon, line 210: 1. why don't you use `return [str]` ?
2. out of curiosity, could you write a comment why the name is `grepTarget`?

hyunsik-yoon:
Could you add unit tests into this PR?

src/OneExplorer.ts
dayo09: Yeap. I think that this type of functions are better to be extracted out as a module, I don't expected OneExplorer to parse things, but as it need to look into config file to grep target models, parser is needed. I have told this to @llFreetimell before.

I also plan to extract this as a function, but later.

llFreetimell: Hmm...? What I talked with @dayo09 was about parsing `.cfg`, not about getting file extension list...? @.@

dayo09: > I think that this type of functions are better to be extracted out as a module,

@llFreetimell OneExplorer now looks into ini file and parse [one-codegen] config so on. I say this can be extracted out as a module.

dayo09: I am worried if you are against the idea - making another module to look into filesystem and parse config file and specify file extension list?

src/OneExplorer.ts
dayo09: 1. sure.
2. I thought this function would do grep something as grepTargetInCommand but it actually doesn't. I can replace them with a simple array function, but I didn't want to repeat them. Maybe `grepAll` would do?

src/OneExplorer.ts
llFreetimell: > it would be better to use some backend API to get file extention string by asking backend API.

I understood like following.

```typescript
// before
if (fstat.isFile() && (fname.endsWith('.circle') || fname.endsWith('.tvn'))) {

// after (Executor.getExecutableExt is in #549)
if (fstat.isFile() && Executor.getExecutableExt().includes(fname.substring(fname.lastIndexOf('.')))
```

So I thought that this is irrelevant with ini file or parsing it.

src/OneExplorer.ts
dayo09: Look good. But `loadCfgFile` doesn't handle errors, so may I replace the `readIni` to `loadCfgFile` function later and enhance `loadCfgFile`'s error handling?

yunjayh: :+1:

src/OneExplorer.ts
```diff
@@ -15,13 +15,34 @@
...
+  // TODO check if toString() is required
+  return ini.parse(configRaw.toString());
+}
+
 enum NodeType {
   directory,
   model,
-  config
+  config,
+  baseModel
```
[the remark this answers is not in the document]
dayo09: model mean a target model and baseModel are kinda tflite and onnx. 
They must be distinguished as baseModel is a parent of config and config displays other models as a child.
I name baseModel as they are easy to define but not `model`.
You can suggest better name.

src/OneExplorer.ts
dayo09: I have collapsed basemodel because most users have 10+ configs under their directory. 
Basically it expands all directory. It will be irritating to fold all everytime you open the explorer.
I made this choice based on `cmake` explorer use experience.

src/OneExplorer.ts
dayo09: Currently your code will work but he also implies the case of 'other backends' which uses other ext formats.
The code above won't resolve here, as it gets 'executable ext' which are only .circle and .tvn.
I have temporarily made this to support .circle and .tvn as config child but actually if there are .tflite or .onnx, they should be displayed too.(furthur feature)

As I said, I am planning to make a function like `getChildExt`, of course.

src/OneExplorer.ts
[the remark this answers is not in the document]
hyunsik-yoon: How about `model` -> `outputModel`?
Some comment would help reader. 😸

src/OneExplorer.ts
hyunsik-yoon: @dayo09 What I am expecting is
- ONE-vscode should not know `.tvn` because it is specific to triv backend.
- likewise, if backend dummy's final code has `.dummy` file extension, ONE-vscode should not know `.dummy`.

So as long as we don't have `.tvn` string and as long and this code can handle `.dummy`, it would be great.

hyunsik-yoon: Given timeframe for internal release, I think we can delay this support later. 
@dayo09 Could you add `TODO Get `.tvn` file extension from backend` comment? Thanks.

hyunsik-yoon:
> Could you add unit tests into this PR?

Discussed with @dayo09, it would be better to add unit tests later in future PR, considering timeframe for internal release.

jyoungyun:
Please merge this PR without my approval. :)

src/OneExplorer.ts
dayo09: I understand. I will make a change on another PR.

src/OneExplorer.ts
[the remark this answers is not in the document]
dayo09: I thought outputModel kinda sounds like one-and-only. I used the word 'intermediate' in another functions.

src/OneExplorer.ts
```diff
@@ -172,29 +196,123 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    const config = readIni(configPath);
+    const ext = path.extname(modelPath).slice(1);
+
+    if (ext === undefined || config === null) {
+      return undefined;
+    }
+
+    return config[`one-import-${ext}` as keyof typeof config] ?
+        config[`one-import-${ext}` as keyof typeof config]['input_path'] :
+        undefined;
```
llFreetimell, line 210: (Optional) If possible, maybe better? :D

```suggestion
    return config[`one-import-${ext}` as keyof typeof config]?.['input_path'];
```

src/OneExplorer.ts
```diff
@@ -172,29 +196,123 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+      {section: 'one-optimize', key: 'output_path', grepper: this.grepAll},
+      {section: 'one-quantize', key: 'input_path', grepper: this.grepAll},
+      {section: 'one-quantize', key: 'output_path', grepper: this.grepAll},
+      {section: 'one-codegen', key: 'command', grepper: this.grepTargetInCommand},
+    ];
+
+    let intermediates: string[] = [];
+    for (let loc of targetLocator) {
+      let confSection = config[loc.section as keyof typeof config];
+      let confKey = confSection ? confSection[loc.key as keyof typeof config] : undefined;
```
llFreetimell, line 242: (Optional) :D

```suggestion
      let confKey = confSection?.[loc.key as keyof typeof config];
```

src/OneExplorer.ts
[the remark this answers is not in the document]
dayo09: That's my personal opinion tho. Or we can discuss about this by creating an issue and make a change later.

src/OneExplorer.ts
dayo09: Oh sorry that I missed this comment. 

> Given timeframe for internal release, I think we can delay this support later.
> @dayo09 Could you add TODO Get .tvn file extension from backend comment? Thanks.
> 

I will make immediate chagne about that.

src/OneExplorer.ts
```diff
@@ -172,29 +196,121 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    const files = fs.readdirSync(node.parent);
 
     for (const fname of files) {
       const fpath = path.join(node.parent, fname);
       const fstat = fs.statSync(fpath);
 
-      if (fstat.isFile() && fname.endsWith('.cfg') && (extSlicer(fname) === extSlicer(node.name))) {
-        const pairNode = new Node(NodeType.config, [], vscode.Uri.file(fpath));
-        node.childNodes.push(pairNode);
+      if (fstat.isFile() && (fname.endsWith('.circle') || fname.endsWith('.tvn'))) {
```
dayo09, line 304:
```suggestion
      // TODO(dayo) Get .tvn file extension from backend
      if (fstat.isFile() && (fname.endsWith('.circle') || fname.endsWith('.tvn'))) {
```

src/OneExplorer.ts
```diff
@@ -23,11 +24,30 @@ import {ToolRunner} from './Project/ToolRunner';
 import {Logger} from './Utils/Logger';
 
 const which = require('which');
```
dayo09, line 26:
```suggestion
```

src/OneExplorer.ts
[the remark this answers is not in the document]
hyunsik-yoon: Issue or TODO would sound good.

src/OneExplorer.ts
```diff
@@ -172,29 +195,122 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+  };
+
+  // TODO(dayo) extract file-relative functions as another module
+  private grepAll = (str: string): string[] => {
+    return [str];
+  };
+
+  // TODO(dayo) extract file-relative functions as another module
+  private parseIntermediates = (configPath: string): string[] => {
+    const config = readIni(configPath);
```
hyunsik-yoon, line 222: Maybe is same configfile read more than one time at L200 and then L222?
If so, maybe this could be fixed later.

hyunsik-yoon (approved):

src/OneExplorer.ts
dayo09: I may pass parsed init as argument to the function, but I didn't see the needs as basically our ini file is small that it won't affect performance.

hyunsik-yoon (approved):

llFreetimell (approved):
LGTM

yunjayh (approved):
LGTM

jyoungyun (approved):
LGTM

PR #555, opened 2022-05-03
Add lintcss exclude in package.json

seanshpark:
This will add exclude folders of lintcss in package.json for external css files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,7 +218,7 @@
     "pretest": "npm run compile && npm run lint && npm run lintjs && npm run lintcss && npm run linthtml",
     "lint": "eslint src --ext ts --max-warnings 0",
     "lintjs": "eslint media --ext js --max-warnings 0",
-    "lintcss": "csslint media src",
+    "lintcss": "csslint media src --exclude-list=media/Jsontracer,media/Circletracer",
     "linthtml": "htmlhint media src",
     "unittest": "node ./out/Tests/runTest.js",
     "test": "npm run unittest",


jyoungyun (approved):
LGTM

package.json
```diff
@@ -218,7 +218,7 @@
     "pretest": "npm run compile && npm run lint && npm run lintjs && npm run lintcss && npm run linthtml",
     "lint": "eslint src --ext ts --max-warnings 0",
     "lintjs": "eslint media --ext js --max-warnings 0",
-    "lintcss": "csslint media src",
+    "lintcss": "csslint media src --exclude-list=media/Jsontracer,media/Circletracer",
```
llFreetimell, line 221: According to the commit message, only excluding `external` directory may be correct.

```suggestion
    "lintcss": "csslint media src --exclude-list=media/Jsontracer/external,media/Circletracer/external",
```

BTW, `media/Jsontracer` does not have `external` directory...?

seanshpark: css files showing warnings are in media/Jsontracer and media/Circletracer folder

seanshpark: `external` in commit message is from external source, not `external` folder.

llFreetimell: I could find `media/Jsontracer/style.css` and `media/Circletracer/style.css` in the two directories.
Those files are not external but our source code.
As the comment in both files say, they only referenced the result of external open source.

So warnings for those two files should be fixed, not ignored :)

seanshpark: I do not have any intension to fix these.

PR #556, opened 2022-05-04
Add csslint to exclude CircleGraph/external

seanshpark:
This will add csslint tool to exclude media/CircleGraph/external folder
as files are from external source.


llFreetimell (approved):
LGTM

hyunsik-yoon:
I tried to search if there's any file format like `.gitignore` for csslint. However it seems that they don't support such file.

hyunsik-yoon (approved):

PR #558, opened 2022-05-04
[Execute] Add execute quickInput

yunjayh:
This commit adds execute quickInput which select the backend, model to run,
and input-spec as option. This will be deprecated when one-infer has been landed.

---

draft: #522

### DIFF
--- a/src/Execute/executeQuickInput.ts
+++ b/src/Execute/executeQuickInput.ts
@@ -1,0 +1,166 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {exec} from 'child_process';
+import {appendFileSync} from 'fs';
+import * as vscode from 'vscode';
+
+import {Backend} from '../Backend/API';
+import {globalBackendMap} from '../Backend/Backend';
+import {MultiStepInput} from '../Utils/MultiStepInput';
+
+export async function runInferenceQuickInput(context: vscode.ExtensionContext) {
+  interface State {
+    backend: Backend;
+    modelPath: vscode.Uri;
+    inputSpec: string;
+    selectedItem: vscode.QuickPickItem;
+    error: string|undefined;
+  }
+
+  async function collectInputs() {
+    const state = {error: undefined} as Partial<State>;
+    await MultiStepInput.run(input => pickBackend(input, state));
+    return state as State;
+  }
+
+  async function pickBackend(input: MultiStepInput, state: Partial<State>) {
+    const backends: vscode.QuickPickItem[] = Object.keys(globalBackendMap).map(label => ({label}));
+    state.selectedItem = await input.showQuickPick({
+      title: 'Choose Executor Toolchain',
+      step: 1,
+      totalSteps: 3,
+      placeholder: 'Select a Backend',
+      items: backends,
+      shouldResume: shouldResume
+    });
+    state.backend = globalBackendMap[state.selectedItem.label];
+    return (input: MultiStepInput) => selectInputModel(input, state);
+  }
+
+  // TODO: Use quickPick window with fast grep child process
+  async function selectInputModel(input: MultiStepInput, state: Partial<State>) {
+    if (state.backend === undefined) {
+      state.error = 'Backend to infer is not chosen. Please check once again.';
+      return;
+    }
+    if (state.backend.executor() === undefined) {
+      state.error = 'Backend executor is not set yet. Please check once again.';
+      return;
+    }
+
+    let backendName: string = state.backend.name();
+    const executor = state.backend.executor();
+    let filter: {[name: string]: string[]} = {};
+    // List files which are filtered with executable extensions
+    if (executor) {
+      filter = {backendName: executor.getExecutableExt()};
+    }
+
+    const fileUri = await vscode.window.showOpenDialog({
+      title: `Select Model to Infer (2/3)`,
+      canSelectMany: false,
+      openLabel: 'Select Model to Infer',
+      filters: filter
+    });
+    if (fileUri && fileUri[0]) {
+      state.modelPath = fileUri[0];
+    } else {
+      console.log('No model has been selected');
+      state.error = 'No model has been selected. Please check once again.';
+      state.modelPath = undefined;
+      return;
+    }
+
+    return (input: MultiStepInput) => selectInputSpec(input, state);
+  }
+
+  // TODO: enable the backend-driven option steps by backend extension
+  async function selectInputSpec(input: MultiStepInput, state: Partial<State>) {
+    const inputSpecKeys: vscode.QuickPickItem[] =
+        ['any', 'non-zero', 'positive'].map(label => ({label}));
+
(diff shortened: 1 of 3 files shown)


YongseopKim:
You should run `format-checker`.

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  // TODO: Use quickPick window with fast grep child process
+  async function selectInputModel(input: MultiStepInput, state: Partial<State>) {
+    if (state.backend === undefined) {
+      throw Error('Backend is undefined');
+    }
+
+    let backendName: string = state.backend.name();
+    const executor = state.backend.executor();
```
YongseopKim, line 61: if executor is undefined?

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+      totalSteps: 3,
+      placeholder: 'Select Random Input Spec',
+      items: inputSpecKeys,
+      shouldResume: shouldResume
+    });
+    state.inputSpec = state.infer.label;
+  }
+
+  async function runInference (state: State){
+    let cmd = state.backend.executor()?.runInference(state.modelPath.path, ["--input-spec", state.inputSpec]);
```
YongseopKim, line 103: if executor is undefined?

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+    const ws = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders.map(f => f.uri.fsPath) : process.cwd();
+    const q = process.platform === 'win32' ? '\\' : '/';
```
YongseopKim, line 105: is it impossible to use nodejs's path api?
- https://nodejs.org/api/path.html?

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+    let outFileName: string = ws[0].concat(`${q}${state.modelPath.path.split(q).slice(-1)[0]}.infer.log`);
```
YongseopKim, line 106: For reviewers, you can add comments for this code or use multiple variables.

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+      location: vscode.ProgressLocation.Notification,
+      title: "Inference Running!",
+      cancellable: true
+    }, (progress, token) => {
+      token.onCancellationRequested(() => {
+        console.log("User canceled the log running operation");
+      });
+      const p = new Promise((resolve, reject) => {
+        exec(cmd?.str() + ' > ' + outFileName, (error, stdout, stderr) => {
+          if (error) { return reject(error.message); }
```
YongseopKim, line 118: Q. This returns rejects() and below code doesn't return resolve(). Why does it return differently?

src/Execute/executeQuickInput.ts
yunjayh: I'll add error handler. User may notice the error by notification which you can normally find in right bottom on your vscode window.

```ts
let cmd = state.backend.executor()?.runInference(state.modelPath.path, ["--input-spec", state.inputSpec]);
// For example,
if (cmd === undefined)  { 
  console.log("Backend executor is undefined."); 
  Balloon("Your Backend Executor is undefined!");
  return
}
```

src/Execute/executeQuickInput.ts
yunjayh: It will considered by the code of following comment.( https://github.com/Samsung/ONE-vscode/pull/558/files#r864502593)
Thank you for notifying me!

src/Execute/executeQuickInput.ts
yunjayh: I'm sorry but I couldn't catch your question. What does `below code` mean?

YongseopKim: This

```ts
if (error) {
  return reject(error.message);
}
resolve(stdout);
```

yunjayh: (First of all, I'm not familiar with Promise and related expressions. Please let me know if there is any misunderstood.

`vscode.window.withProgress`'s second parameter `task` is a callback which returns a promise. And, `withProgress` returns the promise which is returned by `task`.
```ts
      const p = new Promise((resolve, reject) => {
        exec(cmd?.str() + ' > ' + outFileName, (error, stdout, stderr) => {
          if (error) { return reject(error.message); }
          resolve(stdout);
        });
      });
      return p;
```
So, each `reject` and `resolve` are handled by `withProgress(...).then(()=>{}, ()=>{})` as you can find in the line after `return p;`.
If the command emits any error, it will show error notification. On the other case of successful execution, it will show **Inference Succeed!**.

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+    infer: vscode.QuickPickItem;
+  }
+
+  async function collectInputs() {
+    const state = {} as Partial<State>;
+    await MultiStepInput.run(input => pickBackend(input, state));
+    return state as State;
+  }
+
+  const title = 'Choose Executor Toolchain';
```
hyunsik-yoon, line 38: How about `Set parameters to run inference`?

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  async function pickBackend(input: MultiStepInput, state: Partial<State>) {
+    const backends: vscode.QuickPickItem[] = Object.keys(globalBackendMap).map(label => ({ label }));
+    state.infer = await input.showQuickPick({
+      title,
+      step: 1,
+      totalSteps: 3,
+      placeholder: 'Pick Executor',
```
hyunsik-yoon, line 46: How about `Select a backend`?

src/Execute/executeQuickInput.ts
yunjayh: How about considering it later pr?

src/Execute/executeQuickInput.ts
yunjayh: I changed the titles different for each steps. :D

src/Execute/executeQuickInput.ts
YongseopKim: Umm... I was just curious about the reason why you didn't choose below options.

(1)
```ts
if (error) return rejct();
else return resolve();
```

(2)
```ts
if (error) reject()
else resolve()
```

You mixed `return reject()` and `resolve()`. I just wanted to know the reason. 😀

src/Execute/executeQuickInput.ts
YongseopKim: okay

src/Execute/executeQuickInput.ts
YongseopKim: I can't access the uri.

src/Execute/executeQuickInput.ts
YongseopKim: And it could be important because of defining the return type.

yunjayh: Oh, I'm totally ignoring the `return`.. I didn't intended but the source I referred wrote this way. I'll apply the first way 😅

src/Execute/executeQuickInput.ts
```diff
@@ -0,0 +1,166 @@
...
+  // TODO: Use quickPick window with fast grep child process
+  async function selectInputModel(input: MultiStepInput, state: Partial<State>) {
+    if (state.backend === undefined) {
+      state.error = 'Backend to infer is not chosen. Please check once again.';
+      return;
+    }
+    if (state.backend.executor() == undefined) {
+      state.error = 'Backend executor is not set yet. Please check once again.';
+      return;
+    }
```
yunjayh, line 63: I've added error handling param to the state. The lack of information to infer a model, which can be `backend`, `executor`, `modelPath`, options for executor command, will add a error message.

/cc @YongseopKim

src/Execute/executeQuickInput.ts
yunjayh: Now, it's applied :D

hyunsik-yoon (approved):

YongseopKim (approved):
LGTM

Issue #524, opened 2022-04-28
Right-clicking of cfg file will show 'Run' contextual menu


hyunsik-yoon:
for #494

This is to enable the following in #494:

- Right-clicking of cfg file will show 'Run' contextual menu
  - Clicking 'Run' will run onecc

PR #559, opened 2022-05-04
Introduce OneccRunner class to show progress

hyunsik-yoon:
This shows progress in notification when onecc is running by
clicking cfg file in OneExplorer view.

parent issue : https://github.com/Samsung/ONE-vscode/issues/524

### DIFF
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -260,7 +260,7 @@ class OneccRunner extends EventEmitter {
   private onStartRunningOnecc(runnerPromise: Promise<string>) {
     const progressOption: vscode.ProgressOptions = {
       location: vscode.ProgressLocation.Notification,
-      title: `Running onecc --config ${this.cfgUri.fsPath}`,
+      title: `Running: 'onecc --config ${this.cfgUri.fsPath}'`,
       cancellable: true
     };
 
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -273,7 +273,6 @@ class OneccRunner extends EventEmitter {
       const p = new Promise<void>(resolve => {
         runnerPromise
             .then(value => {
-              // successfully running onecc
               resolve();
               this.emit(this.finishedRunningOnecc);
             })
@@ -288,7 +287,6 @@ class OneccRunner extends EventEmitter {
   }
 
   private onFinishedRunningOnecc() {
-    vscode.window.showInformationMessage(
-        `Successfully completed`);
+    vscode.window.showInformationMessage(`Successfully completed`);
   }
 }
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -279,7 +279,7 @@ class OneccRunner extends EventEmitter {
             })
             .catch(value => {
               vscode.window.showWarningMessage(
-                  `Error: Running onecc --config ${this.cfgUri.fsPath}`);
+                  `Error occured while running: 'onecc --config ${this.cfgUri.fsPath}'`);
             });
       });
 
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -289,6 +289,6 @@ class OneccRunner extends EventEmitter {
 
   private onFinishedRunningOnecc() {
     vscode.window.showInformationMessage(
-        `Successfully finished running onecc --config ${this.cfgUri.fsPath}`);
+        `Successfully completed`);
   }
 }
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -201,7 +201,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
 }
 
 export class OneExplorer {
-  constructor(context: vscode.ExtensionContext) {
+  constructor(context: vscode.ExtensionContext, logger: Logger) {
     const rootPath =
         (vscode.workspace.workspaceFolders && (vscode.workspace.workspaceFolders.length > 0)) ?
         vscode.workspace.workspaceFolders[0].uri :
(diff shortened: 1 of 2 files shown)


hyunsik-yoon:
Screen capture: 

[progress.zip](https://github.com/Samsung/ONE-vscode/files/8617995/progress.zip)

jyoungyun (dismissed):
LGTM

YongseopKim (approved):
LGTM and the title could be changed to `Introduce OneccRunner class to show process` because I think `OneccRunner` class is very important in this PR. And the class looks great.

hyunsik-yoon:
@YongseopKim I didn't intended to introduce `OneccRunner` but I had to because some runner with event emitting should be needed. So I created this class to inherit EventEmitter. T.T

src/OneExplorer.ts
```diff
@@ -226,30 +227,68 @@ export class OneExplorer {
...
+    // Show progress UI
+    vscode.window.withProgress(progressOption, (progress, token) => {
+      token.onCancellationRequested(() => {
+        vscode.window.showWarningMessage(`Error: NYI`);
+      });
+
+      const p = new Promise<void>(resolve => {
+        runnerPromise
+            .then(value => {
+              // successfully running onecc
```
dayo09, line 276:
```suggestion
              // successfully running onecc
```

(optional)
Redundant comment

src/OneExplorer.ts
```diff
@@ -226,30 +227,68 @@ export class OneExplorer {
...
+            });
+      });
+
+      return p;
+    });
+  }
+
+  private onFinishedRunningOnecc() {
+    vscode.window.showInformationMessage(
+        `Successfully finished running onecc --config ${this.cfgUri.fsPath}`);
```
dayo09, line 292:
```suggestion
        `Successfully completed`);
```

(optional)
I think the message will be concise when it's shorter.
It will be shown what file is running until just before this message appears, so that user will know the command.

src/OneExplorer.ts
```diff
@@ -226,30 +227,68 @@ export class OneExplorer {
...
+    }
+
+    const runnerPromise = toolRunner.getRunner('onecc', oneccPath, toolArgs, cwd);
+    this.emit(this.startRunningOnecc, runnerPromise);
+  }
+
+  private onStartRunningOnecc(runnerPromise: Promise<string>) {
+    const progressOption: vscode.ProgressOptions = {
+      location: vscode.ProgressLocation.Notification,
+      title: `Running onecc --config ${this.cfgUri.fsPath}`,
```
dayo09, line 263:
```suggestion
      title: `Running: 'onecc --config ${this.cfgUri.fsPath}'`,
```
Mayby this will enhance readability

src/OneExplorer.ts
```diff
@@ -226,30 +227,68 @@ export class OneExplorer {
...
+      const p = new Promise<void>(resolve => {
+        runnerPromise
+            .then(value => {
+              // successfully running onecc
+              resolve();
+              this.emit(this.finishedRunningOnecc);
+            })
+            .catch(value => {
+              vscode.window.showWarningMessage(
+                  `Error: Running onecc --config ${this.cfgUri.fsPath}`);
```
dayo09, line 282:
```suggestion
                  `Error occured while running: 'onecc --config ${this.cfgUri.fsPath}'`);
```
Maybe this will enhance readability

src/OneExplorer.ts
```diff
@@ -216,7 +216,8 @@ export class OneExplorer {
 
     let runCfgDisposal =
         vscode.commands.registerCommand('onevscode.run-cfg', (oneNode: OneNode) => {
-          handleRunOnecc(oneNode.node.uri, new Logger);
```
dayo09, line 219: (This is an extra discussion)
(Q) Actually, I wonder why this line is critical to change. Is there any reason to share the same logger? The logger firstly looked like a log format provider. But it isn't? 

1. If we need to use the same logger across our vscode extension, why don't we provide a global context?
2. If loggers are not to be shared, why don't we just make it disposable as our inner project does?

I find it is designed to uses createOutputChannel('ONE-VSCode'), but I cannot see the difference between new Logger to shared logger.

jyoungyun (approved):
LGTM

src/OneExplorer.ts
hyunsik-yoon: That is a good point. I haven't really thought over logger and maybe we could consider after this release.
I will create an issue.

For this change,
I am not sure if having multiple `vscode.OutputChannel` is thread-safe. Maybe using one instance could be a bit safer.

src/OneExplorer.ts
```diff
@@ -226,30 +227,66 @@ export class OneExplorer {
...
+    const toolRunner = new ToolRunner(this.logger);
+
+    this.on(this.startRunningOnecc, this.onStartRunningOnecc);
+    this.on(this.finishedRunningOnecc, this.onFinishedRunningOnecc);
+
+    const toolArgs = new ToolArgs('-C', this.cfgUri.fsPath);
+    const cwd = path.dirname(this.cfgUri.fsPath);
+    let oneccPath = toolRunner.getOneccPath();
+    if (oneccPath === undefined) {
+      throw new Error('Cannot find installed onecc');
```
dayo09, line 253: (optional)
I'd be better to let user know about detail

(1) onecc is not given in cfg file
(2) The installed onecc is found in the system
(2-1) if onecc is not given in cfg file, so getOneccPath() tried to find default path ('/usr/share/one/bin/onecc') and it doens't exist.
(2-2) given onecc is not found

src/OneExplorer.ts
hyunsik-yoon: #562 was created for further discussion.

dayo09 (approved):
LGTM, I leave some comments behind(which are all optional)

YongseopKim (approved):
LGTM

src/OneExplorer.ts
dayo09: Please ignore the message above

PR #565, opened 2022-05-06
Fix duplicated item in package.json

llFreetimell:
There are two `customEditors` in `package.json`.
These should be integrated as one.


mhs4670go (approved):
LGTM

dayo09 (approved):

Issue #520, opened 2022-04-28
CircleGraph for simple Circle model graph representation and user interactions


seanshpark:
After analysis and some testing with netron source codes,
I think we can secure graph representation of circle model is possible.
This task is to discuss and share the progress of this activity.

To provide;
- show circle mode as graph preresentation with same look and feel with netron application
- select muitiple nodes with pointing device (like mouse)
- get information (node names and ID(index)) of selected nodes
- and maybe more...

seanshpark:
Previous task: #481 and items to do

TODO solve issues
- [x] fix failed to load `circle-metadata.json` file
- [ ] fix vscroll not visible
- [ ] ...

TODO first step --> make it just a simple graph view
- [x] modify ~css~ to show in color
- [x] disable show node properties
- [x] revise menu
- [ ] fix huge(?) loading time and no response for a while for iv3
- [ ] ...

TODO second step --> with selection UI
- [x] add css for selected state -> we can reuse exiting `red border`
- [x] click to change selection status
- [ ] drag selection -> conflicts with graph panning -> combine with ctrl? alt?
- ~shift-click~
- ~shift-drag~
- [ ] ...

TODO third step --> save selection information
- [ ] ...

TODO forth step
- [ ] multiple sub-graph selection ???
- [ ] ...

seanshpark:
CC @mhs4670go

seanshpark:
> fix huge(?) loading time and no response for a while for iv3

Status
- V3 mode is about 90MB.
- converting to Base64, send, recover to Uint8Array takes about 35 seconds with my Linux machine
- log shows `[Violation] 'load' handler took 631ms` and `[Violation] 'message' handler took 14113ms`

seanshpark:
In document https://code.visualstudio.com/api/references/vscode-api of `postMessage(message: any)`,

> However if your extension targets vscode 1.57+ in the engines field of its package.json, any ArrayBuffer values that appear in message will be more efficiently transferred to the webview and will also be correctly recreated inside of the webview.

```diff
diff --git a/package.json b/package.json
index fbe31a5..584c358 100644
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "0.0.1",
   "publisher": "Samsung",
   "engines": {
-    "vscode": "^1.46.0"
+    "vscode": "^1.57.0"
   },
   "categories": [
     "Other"
```
this gives about 23 seconds

seanshpark:
[...]

PR #575, opened 2022-05-09
Upgrade package engines vscode

seanshpark:
This will upgrade engines.vscode in package file to 1.57+.
- postMessage with Uint8Array works correctly and faster with this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "vsce": "^2.7.0"
       },
       "engines": {
-        "vscode": "^1.46.0"
+        "vscode": "^1.57.0"
       }
     },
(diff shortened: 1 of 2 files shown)


seanshpark:
In document https://code.visualstudio.com/api/references/vscode-api of `postMessage(message: any)`,

> However if your extension targets vscode 1.57+ in the engines field of its package.json, any ArrayBuffer values that appear in message will be more efficiently transferred to the webview and will also be correctly recreated inside of the webview.

related #520

hyunsik-yoon (approved):
LGTM

jyoungyun (approved):
LGTM

PR #579, opened 2022-05-10
[OneExplorer] Implement create-cfg

dayo09:
This commit implements create-cfg command and it's UI of 'add' icon
as inline command of baseModel(tflite,onnx) on OneExplorer view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,12 @@
         "icon": "$(refresh)"
       },
       {
+        "command": "onevscode.create-cfg",
+        "title": "Create new ONE configuration",
+        "category": "ONE",
+        "icon": "$(add)"
+      },
+      {
         "command": "onevscode.flip-cfg-to-gui",
         "title": "Open cfg in GUI",
         "category": "ONE",
@@ -184,6 +190,11 @@
           "command": "onevscode.run-cfg",
           "when": "view == OneExplorerView && viewItem == config",
           "group": "navigation"
+        },
+        {
+          "command": "onevscode.create-cfg",
+          "when": "view == OneExplorerView && viewItem == baseModel",
+          "group": "inline"
         }
       ],
       "editor/title": [
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -17,9 +17,10 @@
 import * as fs from 'fs';
 import * as ini from 'ini';
 import * as path from 'path';
+import {TextEncoder} from 'util';
 import * as vscode from 'vscode';
-import {CfgEditorPanel} from './CfgEditor/CfgEditorPanel';
 
+import {CfgEditorPanel} from './CfgEditor/CfgEditorPanel';
 import {ToolArgs} from './Project/ToolArgs';
 import {ToolRunner} from './Project/ToolRunner';
 import {Logger} from './Utils/Logger';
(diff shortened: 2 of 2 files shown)


src/OneExplorer.ts
```diff
@@ -128,6 +129,25 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
 
+  createCfg(oneNode: OneNode): void {
+    const dirname = path.parse(oneNode.node.path).dir;
+    const filename = path.parse(oneNode.node.path).name;
+    const extname = path.parse(oneNode.node.path).ext.slice(1);
+
+    const encoder = new TextEncoder;
+
+    // TODO(dayo) Auto-configure more fields
+    vscode.workspace.fs.writeFile(vscode.Uri.file(`${dirname}/${filename}.cfg`), encoder.encode(`
```
jyoungyun, line 140: How about giving users a chance to specify the name of cfg file?

dayo09: Yes, it requires some event listeners, so I have added https://github.com/Samsung/ONE-vscode/issues/584 instead.

package.json
```diff
@@ -67,6 +67,12 @@
         "category": "ONE",
         "icon": "$(refresh)"
       },
+      {
+        "command": "onevscode.create-cfg",
+        "title": "Create new ONE configuration",
```
hyunsik-yoon, line 72: There are people who have _no_ idea what ONE configuration file is. So I think this title is very important for them. :-)
They must understand why they need to create cfg file by looking at this title.

How about adding 'why' message, for example., like the following?

"Create a new ONE config file for compilation, execution, or profiling"

dayo09: Well I personally prefer the simple message.

I think we need to inform the users more properly if they don't know about ONE config - as ONE config is kinda a key notion for our tool. (It's not just a config but it's kidna master file, such as CMakeLists.txt.)

**Instead, I'd like to suggest: (1) or (2)**

#### (1) notification for the users who don't have ANY config file.
For those user who don't have idea about ONE config, we can provide vscode notification saying: You don't have any ONE configuration in your workspace. Try create one by clicking + next to your model files. [reference:notification](https://code.visualstudio.com/api/ux-guidelines/notifications)

#### (2) welcome view
Or we can provide welcome view for more details.
[welcome view](https://code.visualstudio.com/api/ux-guidelines/views#welcome-views)

dayo09: I will add those two things as TODOs at #584

hyunsik-yoon: Those two ideas are great. 😸 

I think tooltip is also important because small percentage of users may not remember what they saw in welcome view or notification.

If you prefer shorter sentence, how about the following?

- `Create a config file for compilation`

dayo09: I am not sure ... I don't think that icon tooltip should provide that much information.

jyoungyun (approved):
LGTM

package.json
hyunsik-yoon: >  then the naming 'ONE configuration' is weak

I agree. That is why I didn't use `ONE configuration file` but `config file`.
I think people like me will think this tooltip helpful.

If you don't like this approach, please ignore the above comment and continue without my approval.

dayo09: @hyunsik-yoon I appreciate your understanding. We can discuss about it later on. 

> I agree. That is why I didn't use ONE configuration file but config file. 

I personally think the both "ONE configuration" and "config file" are weak, 

I thought the naming should be somthing like
(naming) 'config file' ->  'onecc file' 
(extension) '.cfg' -> '.one' 

Such like that, to let users know that :
>  ONE config is kinda a key notion for our tool. (It's not just a config but it's kidna master file, such as CMakeLists.txt of cmake.)

hyunsik-yoon: I nowadays think that UI is too difficult.
I am a person who could represents only a small portion of various customers.
So I think that usability test in future will be very important for various customers.

hyunsik-yoon (approved):

PR #581, opened 2022-05-10
[CircleGraph] Introduce formatted view.js

seanshpark:
This will introduce index.js with format applied

- added '// eslint-disable-next-line' for variables don't follow camelCase
- changed '==' to '===' and '!=' to '!=='
- use tab size 4 to follow original source
- 'clang-format off' for better readable codes


seanshpark:
For #520

Our modification will be applied to this `view.js` and `external/view.js` will be left as-is for reference.

hyunsik-yoon (dismissed):

hyunsik-yoon:
Out of curious, can we just use `external/view.js`?

seanshpark:
> Out of curiosity, can we just use external/view.js?

we need some modification; revise menu, provide multiple-selection to the node and maybe more.

seanshpark:
@hyunsik-yoon , I added a small change to match clang-format set. PTAL again.

llFreetimell (approved):

hyunsik-yoon (approved):

PR #591, opened 2022-05-11
[CircleGraph] Enable retainContextWhenHidden

seanshpark:
This will add retainContextWhenHidden option to prevent reloading models.

### DIFF
--- a/src/CircleGraph/CircleGraph.ts
+++ b/src/CircleGraph/CircleGraph.ts
@@ -110,12 +110,16 @@ export class CircleGraphPanel {
   }
 }
 
-function getWebviewOptions(extensionUri: vscode.Uri): vscode.WebviewOptions {
+function getWebviewOptions(extensionUri: vscode.Uri): vscode.WebviewOptions&
+    vscode.WebviewPanelOptions {
   return {
     // Enable javascript in the webview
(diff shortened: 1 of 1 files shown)


seanshpark:
from https://github.com/Samsung/ONE-vscode/issues/520#issuecomment-1123083990

src/CircleGraph/CircleGraph.ts
```diff
@@ -110,12 +110,16 @@ export class CircleGraphPanel {
   }
 }
 
-function getWebviewOptions(extensionUri: vscode.Uri): vscode.WebviewOptions {
+function getWebviewOptions(extensionUri: vscode.Uri): vscode.WebviewOptions&
+    vscode.WebviewPanelOptions {
```
hyunsik-yoon, line 114: Could you explain a bit about the reason of changing the return type?

seanshpark: `createWebviewPanel` option is `options?: WebviewPanelOptions & WebviewOptions` type.
I added `WebviewOptions` type to use `WebviewPanelOptions.retainContextWhenHidden`

hyunsik-yoon (approved):

PR #482, opened 2022-04-20
DRAFT: circle node selector from netron

seanshpark:
on-going draft to provide circle node selector view from netron codes.

ONE-vscode-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #592, opened 2022-05-11
[CircleGraph] Revise view to support only circle

seanshpark:
This will revise view.js to support only circle format.

### DIFF
--- a/media/CircleGraph/view.js
+++ b/media/CircleGraph/view.js
@@ -1625,53 +1625,7 @@ view.ModelFactoryService = class {
         this._host = host;
         this._extensions = new Set(['.zip', '.tar', '.tar.gz', '.tgz', '.gz']);
         this._factories = [];
-        // clang-format off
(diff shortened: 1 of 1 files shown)


seanshpark:
for #520, from draft #482

hyunsik-yoon:
Question! Maybe in future, we may want to show tflite graph. In such case, can we use this circle graph to show tflite graph?

seanshpark:
> Question! Maybe in future, we may want to show tflite graph. In such case, can we use this circle graph to show tflite graph?

We can add `this.register('./tflite', [ '.tflite', '.lite', '.tfl', '.bin', '.pb', '.tmfile', '.h5', '.model', '.json', '.txt' ]);` item back again and related files.

What I think is like this;
- This `CircleGraph` purpose is for extended circle graph actions, like node selection as of now.
- If we are to provide such functions with `tflite` then it is OK to add.
- If we are to just provide tflite graph view then we can invoke netron somehow.

hyunsik-yoon (approved):

PR #596, opened 2022-05-11
[CircleGraph] Tidy index.html

seanshpark:
This will tidy index.html that will not be shown with circle graph.


seanshpark:
for #520, from draft #482

removed parts are original author name which is shown in welcome screen, buttons for loading and for about dialog.

hyunsik-yoon (approved):

seanshpark:
@hyunsik-yoon , I've added `1 approval` label but you've added another review request.
Was it done by some kind of habbit or do you think we need another review?

hyunsik-yoon:
@seanshpark Sorry, it was a bad habbit. 😹 I will remove ONE-vscode.

PR #605, opened 2022-05-11
[CirclrGraph] Add items and inputs/outputs from tflite

seanshpark:
This will add items and inputs/outputs attribute from tflite metadata.


seanshpark:
from #482

hyunsik-yoon (approved):

PR #606, opened 2022-05-11
[CircleGraph] Revise require

seanshpark:
This will revise require for javascript source used to process model files.
- we cannot use dynamic node changes in DOM cause of security
- circle.js and circle-schema.js are added in index.html as static
- this change will make callers to work as expected

### DIFF
--- a/media/CircleGraph/index.js
+++ b/media/CircleGraph/index.js
@@ -273,26 +273,16 @@ host.BrowserHost = class {
             return Promise.resolve(this.window.__exports__[url]);
         }
         return new Promise((resolve, reject) => {
-            this.window.module = {exports: {}};
-            const script = document.createElement('script');
-            script.setAttribute('id', id);
-            script.setAttribute('type', 'text/javascript');
-            script.setAttribute('src', url);
-            script.onload = (e) => {
-                if (this.window.module && this.window.module.exports) {
-                    const exports = this.window.module.exports;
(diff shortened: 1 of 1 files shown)


seanshpark:
from draft #482

media/CircleGraph/index.js
```diff
@@ -273,26 +273,17 @@ host.BrowserHost = class {
...
-            };
-            script.onerror = (e) => {
-                delete this.window.module;
-                reject(new Error('The script \'' + e.target.src + '\' failed to load.'));
-            };
-            this.document.head.appendChild(script);
+            // NOTE this only implements for require('circle')
+            // TODO add more for other formats
+            if (id === 'circle' || id === './circle') {
+                this.window.__modules__[id] = circle;
```
seanshpark, line 279: `circle` is global object created in `external/circle.js`

media/CircleGraph/index.js
```diff
@@ -273,26 +273,17 @@ host.BrowserHost = class {
...
+                resolve(circle);
+            }
+            if (id === 'circle-schema' || id === './circle-schema') {
+                this.window.__modules__[id] = $root.circle;
```
seanshpark, line 283: `$root.circle` is global object created in `external/circle-schema.js`

hyunsik-yoon (approved):

PR #607, opened 2022-05-11
[DRAFT] Device API update and view update

struss:
This is only for a Draft for Device API.


struss:
Oh I will make some compensate for this!

PR #610, opened 2022-05-12
[CircleGraph] Use defines for string

seanshpark:
This will use defines for constant string.


seanshpark:
from draft #482

llFreetimell (approved):

PR #613, opened 2022-05-12
[infra] Fix one-vscode checking logic

llFreetimell:
`pkgName` is like `one-vscode@0.0.1`.
To filter all versions of `one-vscode`, logic should be fixed.
This commit will fix it.


hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

PR #614, opened 2022-05-12
[CirclrGraph] Toggle selection with click

seanshpark:
This will revise to toggle selection with click instead of showing
properties window.


seanshpark:
from draft #482

hyunsik-yoon (approved):

PR #616, opened 2022-05-12
[CircleGraph] Disable show node properties window

seanshpark:
This will disable showing node properties window when clicked.
- click is changed to toggle selection of the node


seanshpark:
from draft #482
this change is to provide multiple selection feature

hyunsik-yoon (approved):

PR #619, opened 2022-05-12
[CircleGraph] loadmodel message in extension

seanshpark:
This will handle loadmodel message in extension to send model to WebView.
- this handles model size smaller than 10MB as single message.


seanshpark:
from draft #482

PR #620, opened 2022-05-12
[infra] Update format checker

llFreetimell:
This commit updates following things for format checker
- Exclude `media/CircleGraph/external` explicitly from format checker list
- Exclude json files in directorys which are included at `DIRECTORIES_NOT_TO_BE_TESTED`
- Remove wrong comment
- Remove unused variable

ONE-vscode-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #518, opened 2022-04-27
[infra] Ignore copyright checker for external

seanshpark:
This will add netron folder to ingore list.

ONE-vscode-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #621, opened 2022-05-12
[Draft] Fix copyright and reference statements

llFreetimell:
Some of copyright and reference links are incorrect.
This commit will fix it.

### DIFF
--- a/media/CircleGraph/external/base.js
+++ b/media/CircleGraph/external/base.js
@@ -1,3 +1,5 @@
+// This file is copied from
+// https://github.com/lutzroeder/netron/blob/ae449ff55642636e6a1eef092eda34ffcba1c684/source/base.js
 
 var base = base || {};
 
--- a/infra/copyright/copyright-checker.ts
+++ b/infra/copyright/copyright-checker.ts
@@ -24,9 +24,11 @@ const copyrightRegex = new RegExp('Copyright \\(c\\) [0-9]{4} Samsung Electronic
 const fileFormatsToCheck = "ts,js,css,html";
 
 const ignorePathList = [
+    "**/out/**",
     "**/node_modules/**",
     "**/media/Circletracer/external/**",
-    "**/media/CircleGraph/external/**"
+    "**/media/CircleGraph/external/**",
+    "**/src/Utils/external/**",
 ];
 
 glob("**/*.{" + fileFormatsToCheck + "}",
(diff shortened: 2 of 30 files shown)


llFreetimell:
This PR will be rebased after #620 is merged

llFreetimell:
### For reviewers

I will re-request after prerequisite PRs are all merged :)

media/CircleGraph/external/base.js
```diff
@@ -1,3 +1,5 @@
+// This file is copied from
+// https://github.com/lutzroeder/netron/blob/ae449ff55642636e6a1eef092eda34ffcba1c684/source/base.js
```
seanshpark, line 2: Adding something like this is strict requirement from our open source division?
When we have to revise to next version then we have to add all these comments for these and that I think is a waste, which I don't like to do.

llFreetimell: > Adding something like this is strict requirement from our open source division?

Unfortunately, yes it is IIUC.
Actually, I wrote about this as "Add copyright & license statement of netron, origin at the top of the file." in https://github.com/Samsung/ONE-vscode/pull/518#issuecomment-1111790784 :(

seanshpark:
> Unfortunately, yes it is IIUC.
Actually, I wrote about this as "Add copyright & license statement of netron, origin at the top of the file." in https://github.com/Samsung/ONE-vscode/pull/518#issuecomment-1111790784 :(

OK, I think I misuderstood what you wrote.

jyoungyun (approved):
LGTM

media/CircleGraph/external/base.js
seanshpark: Could you invide the person in charge for this case. Personlly I would like to know his option as you wrote `IIUC` not `I know`

CC @lemmaa

llFreetimell: I left comment about this at internal secret repo issue(168)

lemmaa:
These files all come from one place, don't they? I don't think it's necessary to specify the origin for every unmodified file. It is enough to create an ORIGIN file in a folder, specify the representative origin of all files, and move and locate the LICENSE of the project as it is. For files to be modified or newly added, we will need to add copyright boilerplate at the top of the file separately.

llFreetimell:
With https://github.com/Samsung/ONE-vscode/pull/621#issuecomment-1126601142, there is nothing to do anymore.
Close!

Issue #598, opened 2022-05-11
[OneExplorer] Support auto-refresh


dayo09:
## What?

Let's support auto-refresh on OneExplorer

dayo09:
[reference: tree-view-sample/src/jsonOutline.ts]

### API Summary

```
changeEvent: vscode.TextDocumentChangeEvent
```

dayo09:
4. Notify user about the cfg node location change 

> When user changes the cfg `input_path` data, this logic will move the cfg file path to another base model. I think in that case, it would be good to give users a notification that the cfg location will be changed. Without noti, it may be confusing that the file position suddenly changes.
> 
> _Originally posted by @jyoungyun in https://github.com/Samsung/ONE-vscode/pull/622#discussion_r871924460_

hyunsik-yoon:
@jyoungyun,

> When user changes the cfg input_path data, this logic will move the cfg file path to another base model.

I am not sure why cfg file needs to be moved. Could you explain this a bit more?

dayo09:
@hyunsik-yoon https://github.com/Samsung/ONE-vscode/pull/622 Please take a look at this PR. You can understand the case.

> llFreetimell
> onDidChange is need for auto refresh?
> I thought that onDidCreate and onDidDelete are enough @.@
> (I can be wrong...)
> 

> dayo09
> Sure, it's critical. It will reflect immediate change on .cfg file. For example you change input_path=add.tflite to input_path=sub.tflite it will show the difference immediately (move the config node to its new parent model or so one).
> 
> jyoungyun
> When user changes the cfg input_path data, this logic will move the cfg file path to another base model. I think in that case, it would be good to give users a notification that the cfg location will be changed. Without noti, it may be confusing that the file position suddenly changes.

hyunsik-yoon:
@dayo09 
Oh. there might be my misunderstanding.

> When user changes the cfg input_path data, this logic will move the cfg file path to another base model.

I thought that 'move' meant moving actual cfg file into directory (file system) where tflite file was.
But I guess what you meant is moving cfg file _node_ under tflite _node_ in OneExplorer... 

I see. Thanks.

PR #622, opened 2022-05-12
[OneExplorer] Implement AutoRefresh

dayo09:
This commit auto-refresh OneExplorer when
any file with .cfg or modelfile extensions within workspace changes.
(Create/Delete/Change)

### DIFF
--- a/src/OneExplorer.ts
+++ b/src/OneExplorer.ts
@@ -123,7 +123,18 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   readonly onDidChangeTreeData: vscode.Event<OneNode|undefined|void> =
       this._onDidChangeTreeData.event;
 
-  constructor(private workspaceRoot: vscode.Uri|undefined) {}
+  // TODO(dayo) Get the ext list(cfg,tflite..) from backend
+  private fileWatcher =
+      vscode.workspace.createFileSystemWatcher(`**/*.{cfg,tflite,onnx,circle,tvn}`);
+
+  constructor(private workspaceRoot: vscode.Uri|undefined) {
+    const fileWatchersEvents =
+        [this.fileWatcher.onDidCreate, this.fileWatcher.onDidChange, this.fileWatcher.onDidDelete];
+
+    for (let event of fileWatchersEvents) {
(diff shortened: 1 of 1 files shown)


src/OneExplorer.ts
```diff
@@ -123,7 +123,18 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
       this._onDidChangeTreeData.event;
 
-  constructor(private workspaceRoot: vscode.Uri|undefined) {}
+  // TODO(dayo) Get the ext list(cfg,tflite..) from backend
+  private fileWatcher =
+      vscode.workspace.createFileSystemWatcher(`**/*.{cfg,tflite,onnx,circle,tvn}`);
+
+  constructor(private workspaceRoot: vscode.Uri|undefined) {
+    const fileWatchersEvents =
+        [this.fileWatcher.onDidCreate, this.fileWatcher.onDidChange, this.fileWatcher.onDidDelete];
```
llFreetimell, line 132: `onDidChange` is need for auto refresh?
I thought that `onDidCreate` and `onDidDelete` are enough @.@
(I can be wrong...)

dayo09: Sure, it's critical. It will reflect immediate change on .cfg file. For example you change `input_path=add.tflite` to `input_path=sub.tflite` it will show the difference immediately.

llFreetimell (approved):
LGTM

src/OneExplorer.ts
jyoungyun: When user changes the cfg `input_path` data, this logic will move the cfg file path to another base model. I think in that case, it would be good to give users a notification that the cfg location will be changed. Without noti, it may be confusing that the file position suddenly changes.

dayo09: @jyoungyun Thanks for the opinion. 😄

jyoungyun (approved):
LGTM

src/OneExplorer.ts
dayo09: https://github.com/Samsung/ONE-vscode/issues/598#issuecomment-1125622234

PR #625, opened 2022-05-12
[CircleGraph] Handle loadmodel for multi packet and error

seanshpark:
This will implement loadmodel for multi packet and error handling.

### DIFF
--- a/src/CircleGraph/CircleGraph.ts
+++ b/src/CircleGraph/CircleGraph.ts
@@ -245,13 +245,18 @@ export class CircleGraphPanel {
         if (this._modelLength <= sendPacketSize) {
           this.sendModelSingle();
         } else {
-          // TODO impelment multi packet
+          this.sendModelMulti(sendPacketSize, offset);
         }
       } catch (err: unknown) {
-        // TODO handle error
+        this.handleLoadError(err);
       }
     } else {
-      // TODO impelment multi packet
+      const nextPacketSize = Math.min(this._modelLength - offset, sendPacketSize);
+      try {
+        this.sendModelMulti(nextPacketSize, offset);
+      } catch (err: unknown) {
+        this.handleLoadError(err);
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


seanshpark:
from draft #482

src/CircleGraph/CircleGraph.ts
```diff
@@ -245,13 +245,21 @@ export class CircleGraphPanel {
...
+        this.handleLoadError(err);
       }
     } else {
-      // TODO impelment multi packet
+      var nextPacketSize = this._modelLength - offset;
+      if (nextPacketSize > sendPacketSize) {
+        nextPacketSize = sendPacketSize;
+      }
+      try {
+        this.sendModelMulti(nextPacketSize, offset);
```
llFreetimell, line 259: FYI, https://google.github.io/styleguide/jsguide.html#features-use-const-and-let says that "The var keyword must not be used." :)
(I learned this from https://github.com/Samsung/ONE-vscode/pull/333#discussion_r824482027)


```suggestion
      try {
        const nextPacketSize = Math.min(this._modelLength - offset, sendPacketSize);
        this.sendModelMulti(nextPacketSize, offset);
```

seanshpark: Your suggestion doesn cover when `this._modelLength - offset > sendPacketSize` so cannot apply this. I'll use `let`

llFreetimell: I think it covers..?

- `Math.min(this._modelLength - offset, sendPacketSize)`
  - returns `sendPacketSize`, when `this._modelLength - offset > sendPacketSize`
  - returns `this._modelLength - offset`, when `this._modelLength - offset <= sendPacketSize`

seanshpark: Ah, I didn't notice `Math.min();` ! thanks!

llFreetimell (approved):

PR #626, opened 2022-05-12
[CircleGraph] Request load model from WebView

seanshpark:
This will enable to request load model from WebView to extension.


seanshpark:
from draft #482

hyunsik-yoon (approved):

PR #629, opened 2022-05-12
[CircleGraph] Do not use var

seanshpark:
This will revise not use var and use const or let.


seanshpark:
from https://github.com/Samsung/ONE-vscode/pull/625#discussion_r871163943

hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

PR #630, opened 2022-05-12
Revise Command.str() considering root

hyunsik-yoon:
This revises `Command.str()` in a way that when `root === true`, `str()` starts with `"sudo"`.

### DIFF
--- a/src/Backend/Command.ts
+++ b/src/Backend/Command.ts
@@ -37,7 +37,7 @@ class Command extends Array<string> {
   }
 
   str(): string {
-    return this.strs().join(' ');
+    return (this.root ? 'sudo ' : '') + this.strs().join(' ');
   }
 };
 


jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

src/Backend/Command.ts
```diff
@@ -37,7 +37,7 @@ class Command extends Array<string> {
   }
 
   str(): string {
-    return this.strs().join(' ');
+    return (this.root ? 'sudo ' : '') + this.strs().join(' ');
```
seanshpark, line 40: (optional) Just another thought. It's not a normal case but what if the user is `root`...
maybe add a `TODO: support or deny root account`

jyoungyun: At this point, it is impossible to handle it because it is not known which environment to call. However, when we actually run this command, we can consider how to ignore or deny root command.

PR #632, opened 2022-05-12
[CircleGraph] Revise selection for find dialog

seanshpark:
This will revise selection characterstics of find dialog
- will not clear when dialog opens
- scroll to node when clicked in find dialog


seanshpark:
from draft #482

hyunsik-yoon (approved):

PR #634, opened 2022-05-13
Support to run root permission job

jyoungyun:
This commit supports to run a job that requires root permission,
such as setting up the toolchain envrionment.

Draft pr: #599

### DIFF
--- a/src/Project/ToolRunner.ts
+++ b/src/Project/ToolRunner.ts
@@ -83,10 +83,20 @@ export class ToolRunner {
     return oneccRealPath;
   }
 
-  public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string) {
+  public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string, root?: boolean) {
     return new Promise<string>((resolve, reject) => {
       this.logger.outputWithTime('Running: ' + name);
-      let cmd = cp.spawn(tool, toolargs, {cwd: path});
+      if (root) {
+        // NOTE
+        // To run the root command job, it must requires a password in `process.env.userp`
+        // environment.
+        // TODO(jyoung): Need password encryption
+        tool = `echo ${process.env.userp} | sudo -S ` + tool;
+      }
+      let cmd = cp.spawn(tool, toolargs, {cwd: path, shell: true});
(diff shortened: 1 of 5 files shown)


hyunsik-yoon (approved):

src/Project/ToolRunner.ts
```diff
@@ -83,10 +83,20 @@ export class ToolRunner {
...
+      if (root) {
+        // NOTE
+        // To run the root command job, it must requires a password in `process.env.userp`
+        // environment.
+        // TODO(jyoung): Need password encryption
+        tool = `echo ${process.env.userp} | sudo -S ` + tool;
+      }
+      let cmd = cp.spawn(tool, toolargs, {cwd: path, shell: true});
+      if (root) {
+        process.env.userp = '';
```
llFreetimell, line 98: Can I ask you why this is reset?
And isn't it possible to integrate two if-statement? :D

src/Project/ToolRunner.ts
```diff
@@ -83,10 +83,20 @@ export class ToolRunner {
...
+      }
```
jyoungyun, line 99:
```suggestion
      if (root) {
        // NOTE
        // To run the root command job, it must requires a password in `process.env.userp`
        // environment.
        // TODO(jyoung): Need password encryption
        tool = `echo ${process.env.userp} | sudo -S ` + tool;
        let cmd = cp.spawn(tool, toolargs, {cwd: path, shell: true});
        process.env.userp = '';
      } else {
        let cmd = cp.spawn(tool, toolargs, {cwd: path});
      }
```
> And isn't it possible to integrate two if-statement? :D

Do you mean this code?

llFreetimell: Sort of :D
Of course, it is not mandatory!

src/Project/ToolRunner.ts
jyoungyun: Becasue this data is the password of the user machine. So I think it should be removed for safety. Actually I don't want to store this information, but there is no way to pass this value, so I stored it temporarily.

llFreetimell (approved):
LGTM

PR #639, opened 2022-05-16
Remove PipedCommand and CommandList

hyunsik-yoon:
Previously we wanted to send a list of commands as `CommandList`
and command with pipe as `PipedCommand`.
After more discussion, we will create one shell script that contains
list of command and return the path of shell script.

Due to this decision, let's remove `PipedCommand` and `CommandList`.

### DIFF
--- a/src/Backend/Command.ts
+++ b/src/Backend/Command.ts
@@ -41,26 +41,4 @@ class Command extends Array<string> {
   }
 };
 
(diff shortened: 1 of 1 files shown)


jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

Issue #643, opened 2022-05-16
Let's subscribe commands


dayo09:
## What?

Let's subscribe all the commands by ONE-vscode.
```js
context.subscriptions.push(COMMAND);
```

## Why?

This makes all the command disposed when the extension expires.
Otherwise, the command will be registered at the beginning of the extension, and remains even after the extension is disabled.

### Lists
- [x] #646
`onevscode.toggleAttrCodelens`
`onevscode.hideAttrCodelens`

- [x] #645 
`onevscode.open-cfg`
`onevscode.refresh-one-explorer`
`onevscode.create-cfg`

dayo09:
Done

PR #658, opened 2022-05-17
[Toolchain] Add toolchain tree provider

jyoungyun:
This commit adds toolchain tree provider.

### DIFF
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as vscode from 'vscode';
+import {gToolchainEnvMap} from './ToolchainEnv';
+
+enum NodeType {
+  backend,
+  toolchain,
+}
+
+export class ToolchainNode extends vscode.TreeItem {
+  constructor(
+      public readonly label: string,
+      public readonly collapsibleState: vscode.TreeItemCollapsibleState,
+      public readonly type: NodeType,
+      public readonly version?: string,
+  ) {
+    super(label, collapsibleState);
+
+    if (type === NodeType.backend) {
+      this.iconPath = new vscode.ThemeIcon('bracket');
+    } else if (type === NodeType.toolchain) {
+      this.iconPath = new vscode.ThemeIcon('circle-filled');
+      this.label = `${this.label} ${this.version}`;
+    }
+  }
+}
+
+export class ToolchainProvider implements vscode.TreeDataProvider<ToolchainNode> {
+  private _onDidChangeTreeData: vscode.EventEmitter<ToolchainNode|undefined|void> =
+      new vscode.EventEmitter<ToolchainNode|undefined|void>();
+  readonly onDidChangeTreeData?: vscode.Event<ToolchainNode|undefined|void> =
+      this._onDidChangeTreeData.event;
+
+  constructor() {}
+
+  getTreeItem(element: ToolchainNode): vscode.TreeItem {
+    return element;
+  }
+
+  getChildren(element?: ToolchainNode): Thenable<ToolchainNode[]> {
+    if (element) {
+      return Promise.resolve(this.getNode(element));
+    } else {
+      const nodes = Object.keys(gToolchainEnvMap)
+                        .map((backend) => this.toToolchainNode(NodeType.backend, backend));
+      return Promise.resolve(nodes);
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -64,9 +65,10 @@ export function activate(context: vscode.ExtensionContext) {
   new OneExplorer(context, logger);
 
   // ONE view
-  let refreshCompiler = vscode.commands.registerCommand('onevscode.refresh-toolchain', () => {
-    console.log('refresh-toolchain: NYI');
-  });
+  const toolchainProvier = new ToolchainProvider();
+  vscode.window.registerTreeDataProvider('ToolchainView', toolchainProvier);
```
dayo09, line 69: The registered treeDataProvider must be subscribed to be disposed after the extension's expiration.
```suggestion
  context.subscriptions.push(
    vscode.window.registerTreeDataProvider('ToolchainView', toolchainProvier));
```

FYI, https://github.com/Samsung/ONE-vscode/issues/643

yunjayh:
Format error T^T

src/Toolchain/ToolchainProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+      public readonly label: string,
+      public readonly collapsibleState: vscode.TreeItemCollapsibleState,
+      public readonly type: NodeType,
+      public readonly version?: string,
+  ) {
+    super(label, collapsibleState);
+
+    if (type === NodeType.backend) {
+      this.iconPath = new vscode.ThemeIcon('bracket');
+    } else if (type === NodeType.toolchain) {
```
hyunsik-yoon, line 36: (optional) "ToolchainNode type can be of toolchain or _backend_" could be confusing. 
How about renaming `ToolchainNode` to `ToolchainViewNode` ?

src/Toolchain/ToolchainProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      return new ToolchainNode(name, vscode.TreeItemCollapsibleState.Expanded, type);
+    }
+  }
+
+  private getNode(node: ToolchainNode): ToolchainNode[] {
+    if (node.type === NodeType.backend) {
+      if (node.label in gToolchainEnvMap) {
+        const toolchain = gToolchainEnvMap[node.label].listInstalled();
+        if (toolchain && toolchain.info.version) {
```
hyunsik-yoon, line 77: Can I ask a question?
Current code of `listInstalled()` seems to return `undefined` even though toolchain was installed by user.

https://github.com/Samsung/ONE-vscode/blob/911636d21dadca657edd8175913c5b7424d6bdcc/src/Toolchain/ToolchainEnv.ts#L116-L118


In such case shouldn't we check if there's any toolchain already installed?

src/Toolchain/ToolchainProvider.ts
jyoungyun: This variable always uses with `NodeType`. So I think it seems unnecessary that the node is attached to the name again.

src/Toolchain/ToolchainProvider.ts
jyoungyun: I already checked the undefine variable.

> In such case shouldn't we check if there's any toolchain already installed?

I don't understand what you mean..

src/Toolchain/ToolchainProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+      public readonly type: NodeType,
+      public readonly version?: string,
+  ) {
+    super(label, collapsibleState);
+
+    if (type === NodeType.backend) {
+      this.iconPath = new vscode.ThemeIcon('bracket');
+    } else if (type === NodeType.toolchain) {
+      this.iconPath = new vscode.ThemeIcon('circle-filled');
+      this.label = `${this.label} ${this.version}`;
```
dayo09, line 38: This line may removed by 
```js
super(`${label} ${version}`, collapsibleState);
```

src/Toolchain/ToolchainProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (element) {
+      return Promise.resolve(this.getNode(element));
+    } else {
+      const nodes = Object.keys(gToolchainEnvMap)
+                        .map((backend) => this.toToolchainNode(NodeType.backend, backend));
+      return Promise.resolve(nodes);
+    }
+  }
+
+  private toToolchainNode(type: NodeType, name: string, version?: string): ToolchainNode {
```
dayo09, line 65: How about merging this into ToolchainNode constructor?

src/Toolchain/ToolchainProvider.ts
jyoungyun: No, it can't.
The label value is different by type.

src/Toolchain/ToolchainProvider.ts
jyoungyun: I will do it later.

src/Toolchain/ToolchainProvider.ts
dayo09: Sure,.

src/Toolchain/ToolchainProvider.ts
dayo09: `toToolchainNode` sounds like a mapper. 
It's optional anyways.

dayo09 (approved):

src/Toolchain/ToolchainProvider.ts
jyoungyun: Yes, you're right. This function will be put into another function with other functions implemented.

src/Toolchain/ToolchainProvider.ts
hyunsik-yoon: Ah. I realized that `ToolchainEnv.installed` was _already_ set when ONE-vscode starts to run. Sorry for confusion.

hyunsik-yoon (approved):

Issue #659, opened 2022-05-17
Introduce method that returns installed toolchains


hyunsik-yoon:
Let's provide a method that returns installed toolchains in backend.

```ts
public getInstalledToolchains(toolchainType: string): Toolchains;
```

hyunsik-yoon:
all merged.

PR #666, opened 2022-05-18
Add Version.equals() method

hyunsik-yoon:
This commit adds` Version.equals()` method.

This was required while working on a draft for #659.

### DIFF
--- a/src/Backend/Version.ts
+++ b/src/Backend/Version.ts
@@ -32,7 +32,7 @@ class Version {
     return ret;
   }
 
-  equals(ver: Version): Boolean {
+  equals(ver: Version): boolean {
     return (
         this.major === ver.major && this.minor === ver.minor && this.patch === ver.patch &&
         this.option === ver.option);
--- a/src/Backend/Version.ts
+++ b/src/Backend/Version.ts
@@ -32,7 +32,13 @@ class Version {
     return ret;
   }
 
-  // TODO: Add operator <, <=, ==, >, >=
+  equals(ver: Version): Boolean {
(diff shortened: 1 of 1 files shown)


src/Backend/Version.ts
```diff
@@ -32,7 +32,13 @@ class Version {
     return ret;
   }
 
-  // TODO: Add operator <, <=, ==, >, >=
+  equals(ver: Version): Boolean {
```
seanshpark, line 35: (optional) I'm not sure about this but there is `Boolean` Object and `boolean` primitive type. I think return type would be `boolean`

yunjayh: FYI) https://stackoverflow.com/questions/64443288/what-is-the-difference-between-boolean-and-boolean-in-typescript

hyunsik-yoon: Thanks!

seanshpark (approved):
LGTM

src/Backend/Version.ts
hyunsik-yoon: Interesting.
```ts
if (new Boolean(false)) {
    console.log('run only when if true');
}
```

seanshpark: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean
```
var x = Boolean(expression);     // use this...
var x = !!(expression);          // ...or this
var x = new Boolean(expression); // don't use this!
```

yunjayh (approved):

PR #668, opened 2022-05-18
[Draft/Device] Add DeviceType api for device register

struss:
This commit will add DeviceType for use on device add.


src/Device/DeviceType.ts
```diff
@@ -0,0 +1,36 @@
...
+  name(): string;
+
+  // device Executable ext
+  getExecutableExt(): string[];
+
+  // list specified name for connectable devices
+  getConnectableDevices(): Command;
+
+  // check certain device connected
+  checkConnected(name: string): Command;
```
struss, line 30: Little bit confused as this will return real result(like `string[]` or `boolean` on this.

PR #674, opened 2022-05-18
[OneExplorer] Add rename command

dayo09:
This commit adds rename command by right-clicking on the nodes of all
types (model files and config).
It doesn't change path, only the filename.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -176,24 +176,20 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       // TODO automatically change the corresponding files
       warningMessage = 'WARNING: You may need to change input paths in the following cfg files:\n';
       for (const conf of oneNode.node.childNodes) {
-        warningMessage += `${conf.name} \n`;
+        // NOTE A newline character is not supported in input box.
+        // vscode-extension doesn't plan to support multiple lines inside in input box or in
+        // notification box.
+        warningMessage += `${conf.name} `;
       }
     } else {
       warningMessage = 'WARNING: Renaming may result in some unexpected changes on the tree view.';
     }
 
     vscode.window
-        .showInputBox({title: warningMessage, placeHolder: 'Enter(proceed) / Escape(cancel)'})
-        .then((answer) => {
-          if (answer === undefined) {
-            // User has pressed esc to cancel.
-            return;
-          } else {
-            return vscode.window.showInputBox({
-              title: 'Enter a file name:',
-              placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`
-            });
-          }
+        .showInputBox({
+          title: 'Enter a file name:',
+          placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`,
+          prompt: warningMessage
         })
         .then(newname => {
           if (newname) {
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -163,6 +163,50 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
     this._onDidChangeTreeData.fire();
   }
 
+  // TODO: Add move()
+
+  /**
+   * Rename a file of all types of nodes (baseModel, derivedModel, config) excepts for directory.
+   * It only alters the file name, not the path.
+   */
+  rename(oneNode: OneNode): void {
+    // TODO: prohibit special characters for security ('..', '*', etc)
+    let warningMessage;
+    if (oneNode.node.type === NodeType.baseModel) {
+      // TODO automatically change the corresponding files
+      warningMessage = 'WARNING: You may need to change input paths in the following cfg files:\n';
+      for (const conf of oneNode.node.childNodes) {
+        warningMessage += `${conf.name} \n`;
+      }
+    } else {
+      warningMessage = 'WARNING: Renaming may result in some unexpected changes on the tree view.';
+    }
+
+    vscode.window
+        .showInputBox({title: warningMessage, placeHolder: 'Enter(proceed) / Escape(cancel)'})
+        .then((answer) => {
+          if (answer === undefined) {
+            // User has pressed esc to cancel.
+            return;
+          } else {
+            return vscode.window.showInputBox({
+              title: 'Enter a file name:',
+              placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`
+            });
+          }
+        })
+        .then(newname => {
+          if (newname) {
+            const dirpath = path.dirname(oneNode.node.uri.fsPath);
+            const newpath = `${dirpath}/${newname}`;
+            vscode.workspace.fs.rename(oneNode.node.uri, vscode.Uri.file(newpath));
+
+            this.refresh();
+            return;
+          }
+        });
+  }
+
   /**
    * Create ONE configuration file for a base model
    * Input box is prefilled as <base model's name>.cfg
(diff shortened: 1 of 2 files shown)


dayo09:
Take a look at the gif file : https://github.sec.samsung.net/RS7-RuntimeNTools/secret/issues/187#issuecomment-1522053

src/OneExplorer/OneExplorer.ts
```diff
@@ -163,6 +163,50 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+   * It only alters the file name, not the path.
+   */
+  rename(oneNode: OneNode): void {
+    // TODO: prohibit special characters for security ('..', '*', etc)
+    let warningMessage;
+    if (oneNode.node.type === NodeType.baseModel) {
+      // TODO automatically change the corresponding files
+      warningMessage = 'WARNING: You may need to change input paths in the following cfg files:\n';
+      for (const conf of oneNode.node.childNodes) {
+        warningMessage += `${conf.name} \n`;
```
jyoungyun, line 179: (Q) Does quick input title support new line character?

jyoungyun, line 179: (Q) Does quick input title support new line character?

src/OneExplorer/OneExplorer.ts
```diff
@@ -163,6 +163,50 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+      }
+    } else {
+      warningMessage = 'WARNING: Renaming may result in some unexpected changes on the tree view.';
+    }
+
+    vscode.window
+        .showInputBox({title: warningMessage, placeHolder: 'Enter(proceed) / Escape(cancel)'})
```
jyoungyun, line 186: You don't need to put `Enter(proceed) / Escape(cancel)` string. Because ShowInputBox already provides the help message on bottom.
rename_help_message

jyoungyun, line 186: You don't need to put `Enter(proceed) / Escape(cancel)` string. Because ShowInputBox already provides the help message on bottom.
rename_help_message

src/OneExplorer/OneExplorer.ts
```diff
@@ -163,6 +163,50 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+          if (answer === undefined) {
+            // User has pressed esc to cancel.
+            return;
+          } else {
+            return vscode.window.showInputBox({
+              title: 'Enter a file name:',
+              placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`
+            });
+          }
+        })
```
yunjayh, line 197: I pulled this pr and tried to use the function. For me, I felt the warning step is a little bit annoying. 
image
I read the description and put a new name to the empty box and pressed enter. However, nothing happened but a new input box appeared and requests new file name again. How about notifying the warning message to user while getting new file name? (Then, the message should be shorter than now..)

```ts
    vscode.window
        .showInputBox({title: 'Enter a file name. ([WARNING] Relative paths between model and cfg could be broken)',
              placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`})
        .then(newname => {
          if (newname) {
            const dirpath = path.dirname(oneNode.node.uri.fsPath);
            const newpath = `${dirpath}/${newname}`;
            vscode.workspace.fs.rename(oneNode.node.uri, vscode.Uri.file(newpath));

            this.refresh();
            return;
          }
        });
```

dayo09: No... It's just a ask box to notify the users. You push enter in the box and then the next box which takes the name will appear.

src/OneExplorer/OneExplorer.ts
dayo09: Sure. Let me remove this then.

src/OneExplorer/OneExplorer.ts
dayo09: Nope.

dayo09:
@jyoungyun @yunjayh I think the first show box only makes people confused. I will remove the first input box which asks people if they will proceed.

src/OneExplorer/OneExplorer.ts
dayo09: I think the UI is making people confused. I have revised the behavior: please visit the gif link again and find the updated gif.

src/OneExplorer/OneExplorer.ts
dayo09: https://stackoverflow.com/a/57735348/7949783

They say vscode team are explicitly removing \n and they will not do support them. I may remove this.

dayo09: https://github.com/microsoft/vscode/issues/48900#issuecomment-385332787

src/OneExplorer/OneExplorer.ts
jyoungyun: I like updated gif procedure. It's much clearer than before. :)

src/OneExplorer/OneExplorer.ts
```diff
@@ -163,6 +163,46 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+          prompt: warningMessage
+        })
+        .then(newname => {
+          if (newname) {
+            const dirpath = path.dirname(oneNode.node.uri.fsPath);
+            const newpath = `${dirpath}/${newname}`;
+            vscode.workspace.fs.rename(oneNode.node.uri, vscode.Uri.file(newpath));
+
+            this.refresh();
+            return;
```
jyoungyun, line 201:
```suggestion
```

jyoungyun: The others looks good to me. :)

src/OneExplorer/OneExplorer.ts
```diff
@@ -163,6 +163,45 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+        .showInputBox({
+          title: 'Enter a file name:',
+          placeHolder: `${path.basename(oneNode.node.uri.fsPath)}`,
+          prompt: warningMessage
+        })
+        .then(newname => {
+          if (newname) {
+            const dirpath = path.dirname(oneNode.node.uri.fsPath);
+            const newpath = `${dirpath}/${newname}`;
+            vscode.workspace.fs.rename(oneNode.node.uri, vscode.Uri.file(newpath));
```
yunjayh, line 198: How about giving proper extension name if it's not given by user? I cannot imagine the danger if we apply following change. 

```suggestion
          if (newname) {
            const dirpath = path.dirname(oneNode.node.uri.fsPath);
            const newpath = `${dirpath}/${newname}`;
            const ext = path.extname(oneNode.node.uri.fsPath);
            if (path.extname(newname) != ext) {
              newname = `${newname}.${ext}`;
            }
            vscode.workspace.fs.rename(oneNode.node.uri, vscode.Uri.file(newpath));
```

dayo09: IMO, it's a little bit too much... Maybe it will bring about much confusion to user. Well, how about just putting an error message instead?

yunjayh: > IMO, it's a little bit too much... Maybe it will bring about much confusion to user.

I agree with you. Since there is a warning message at the bottom of input box, user would pay attention with his/her file naming. IMHO Defining the situation when the error message is popping up will be pretty difficult. 
Then, never mind about the above suggestion!

yunjayh (approved):
LGTM 🔥

src/OneExplorer/OneExplorer.ts
dayo09: @yunjayh I think we can use 'validateInput' API to check if user has given a right ext. I will apply that on the next pr. thanks for your opinion.

jyoungyun (approved):
LGTM Thank you :)

Issue #585, opened 2022-05-10
Define necessary device API


struss:
Need to add `Device` interface to use.

so, we need to define those api below

- interface `DeviceType` to define `user-defined` device action, like `listConnectableDevices` or `runInference`
- interface `ConnectionType` to define local and remote command execution.

struss:
First of all, we need to think about how to register(or connect) device on ONE-vscode extension. 

steps in my thought,

1-1. show `DeviceType` which will define some command about device. this will be register on user-defined `.visx` like `backend`

1-2. select type of host(local or remote)

2. show Connectable device on certain host(local or remote)

3. add selected device on `Target Device View` as a single item.

struss:
Need to implement 3 API

1. DeviceType : A type like backend to register on. user-defined class
2. ConnectType : connection type like local or remote
3. Device : class for `Device View` list display

YongseopKim:
I and @struss discussed offline
- Two concepts
- Device: it is an execution env.
    - HW: x86, arm, npu, ...
    - SW: android, tizen, triv, tflite, ...
- Executor: one-infer, npu_binary_file, ...
- Example) a toolchain is installed on local pc and it has a simulator on pc and npu runner for tv.
    - On `Target Devices`, there could be two targets
    - 1. x86/triv/simulator
    - 2. arm(or tv)/triv/one-infer

YongseopKim:
To clear more, IMO,
- Device(ExecutionEnv): it has both perspectives of HW and SW
- Executor: command set for one-infer, npu_binary_file, ...
- Example0) simulator
    - Device(ExecutionEnv, HW:x86_64;SW:triv)
        - runInfer(): runs the runner with the below command
        - connect(): connects with the below command
        - ...
    - Executor: command set for declaring the runner and peripherals
        - runner: [`one-infer`, 'arg0', ...]
        - peripherals: { 'connect': '', ... }
- Example1) dtv
    - Device(ExecutionEnv, HW:tv;SW:tizen_or_triv)
        - runInfer(): runs the runner with the below command
        - connect(): connects with the below command
        - ...
    - Executor: command set for declaring the runner and peripherals
        - runner: [`npu_binary.file`, 'arg0', 'arg1', ...]
        - peripherals: { 'connect': 'sdb shell', ... }
        - ...

After writing this, I think `ExecutionEnv` can be just inferface enough.

YongseopKim:
[...]

PR #675, opened 2022-05-18
Draft: Device API

struss:
For https://github.com/Samsung/ONE-vscode/issues/585

This commit will draft for device view.

This is just in my thought. any question or suggestion welcome!


struss:
https://github.com/Samsung/ONE-vscode/issues/585#issuecomment-1132476919

follow from this suggestion, I need to fix this with `Executor` API. I will update about this.

struss:
This only handle about Device, not include `Simulator`

Issue #644, opened 2022-05-16
[OneExplorer] Find config file from any directory


dayo09:
## What?

Let's find configuration files from any directory.

Currently, OneExplorer only searches the configuration for the model within the same directory.
Let's find the configuration files within the workspace. 

### Details 
- find the configuration files within the workspace. 
- show the config file's name as it's relative path. -> show only the name

### Related PRs
- #669 
- #672
- #681

hyunsik-yoon:
I realized that when `input_path` is relative path, it is relative to `onecc`'s working directory.
Only by looking into `input_path` in cfg file, it seems very difficult to find the base model, doesn't it? 😿 

```shell
$ pwd
/home/eric/m/src/model/onecc-test 

$ tree
.
├── cfg
│   └── inception_v3_conv2d.0to128.cfg
├── inception_v3
│   └── 2020-01-31
│       └── inception_v3_conv2d.0to128.tflite
└── trash

4 directories, 2 files

$ cat cfg/inception_v3_conv2d.0to128.cfg
[onecc]
one-import-tf=False
one-import-tflite=True
one-import-bcq=False
one-import-onnx=False
one-optimize=True
one-quantize=True
one-pack=False
one-codegen=True

[one-import-tflite]
input_path=./inception_v3/2020-01-31/inception_v3_conv2d.0to128.tflite
output_path=./trash/model.circle

[one-optimize]
input_path=./trash/model.circle
output_path=./trash/model.opt.circle
all=True
remove_redundant_transpose=True

[one-quantize]
input_path=./trash/model.opt.circle
output_path=./trash/model.q8.circle
granularity=channel

[one-codegen]
backend=tv2
command=-o inception_v3/2020-01-31/model.tvn --DSP-quota 64K ./trash/model.q8.circle

$ onecc -C ./cfg/inception_v3_conv2d.0to128.cfg
Recording 0'th data
Recording 1'th data
Recording 2'th data
Recording finished. Number of recorded data: 3
<< compiling message was omitted >>

$ tree
.
├── cfg
│   └── inception_v3_conv2d.0to128.cfg
├── inception_v3
│   └── 2020-01-31
│       ├── inception_v3_conv2d.0to128.tflite
│       └── model.tvn
└── trash
    ├── model.circle
    ├── model.circle.log
    ├── model.opt.circle
    ├── model.opt.circle.log
    ├── model.q8.circle
    └── model.q8.circle.log

4 directories, 9 files
```

[...]

PR #681, opened 2022-05-19
[OneExplorer] Introduce ConfigObject

dayo09:
This commit introduces ConfigObject class which is composed of
parsed dictionary object of given ini file.

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -132,6 +132,8 @@ export class ConfigObj {
    */
   private static parseBaseModels = (uri: vscode.Uri, iniObj: object): vscode.Uri[] => {
     let baseModels: string[] = [];
+    
+    // TODO Get ext list from backend
     for (const ext of ['tflite', 'pb', 'onnx']) {
       let confSection = iniObj[`one-import-${ext}` as keyof typeof iniObj];
       let confKey = confSection ?.['input_path' as keyof typeof iniObj] as string;
@@ -182,6 +184,7 @@ export class ConfigObj {
         section: 'one-codegen',
         key: 'command',
         filt: (str: string): string[] => {
+          // TODO Get ext list from backend
           return str.split(' ').filter(
               e => path.extname(e) === '.tvn' || path.extname(e) === '.circle');
         }
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -1,0 +1,194 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from 'fs';
+import * as ini from 'ini';
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+/**
+ * @brief Parsed .cfg file into an 'obj' which contains fields such as
+ *        'baseModels' or 'derivedModels'
+ *
+ * @usage Direct Parse
+ * ```
+ * const {baseModels, derivedModels} = configObj.parse(uri);
+ * ```
+ *
+ * @usage Create Parsed Config Object
+ *        (use a factory function `createConfigObj`)
+ * ```
+ * const configObj = createConfigObj(uri);
+ * ```
+ */
+export class ConfigObj {
+  /**
+   * source uri of config file
+   */
+  uri: vscode.Uri;
+
+  /**
+   * a raw ini object read from config file
+   */
+  rawObj: object;
+
+  /**
+   * a parsed config object
+   */
+  obj: {
+    baseModels: vscode.Uri[],
+    derivedModels: vscode.Uri[],
+    // TODO: Add more fields
+  };
+
+  private constructor(uri: vscode.Uri, rawObj: object) {
+    this.uri = uri;
+    this.rawObj = rawObj;
+    this.obj = {
+      baseModels: ConfigObj.parseBaseModels(uri, rawObj),
+      derivedModels: ConfigObj.parseDerivedModels(uri, rawObj)
+    };
+  }
+
+  /**
+   * A simple parse function with directly `ConfigObj.obj` fields after parsing
+   * without keeping an ConfigObj instance.
+   *
+   * @param uri
+   * @returns `{baseModels, derivedModels}` or
+   *          `null` if failed to open/parse into ini object
+  }
+   */
+  public static parse(uri: vscode.Uri): {baseModels: vscode.Uri[],
+                                         derivedModels: vscode.Uri[]}|null {
+    const cfgObj = this.createConfigObj(uri);
+
+    if (!cfgObj) {
+      return null;
+    }
+
+    return cfgObj.obj;
+  }
+
+  /**
+   * A factory function to create ConfigObj for the given uri.
+   *
+   * @param uri
+   * @returns `ConfObj` or
+   *          `null`    if failed to open/parse into ini object
+   */
+  public static createConfigObj(uri: vscode.Uri) {
+    const obj = this.importIni(uri.fsPath);
+    if (!obj) {
+      console.error(`Cannot open ${uri.fsPath}`);
+
+      return null;
+    }
+
+    return new ConfigObj(uri, obj);
+  }
+
+  /**
+   * Import an ini file
+   * @param filePath
+   * @returns `object` if file read is successful, or `null` if file open has failed
+   *
+   */
+  private static importIni(filePath: string): object|null {
+    let configRaw: string;
+    try {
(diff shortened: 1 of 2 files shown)


src/OneExplorer/ConfigObject.ts
```diff
@@ -0,0 +1,194 @@
...
+ * ```
+ * const {baseModels, derivedModels} = configObj.parse(uri);
+ * ```
+ *
+ * @usage Create Parsed Config Object
+ *        (use a factory function `createConfigObj`)
+ * ```
+ * const configObj = createConfigObj(uri);
+ * ```
+ */
```
dayo09, line 36: Please read this part to get the idea of this class

src/OneExplorer/ConfigObject.ts
```diff
@@ -0,0 +1,194 @@
...
+   * ABOUT MULTIPLE BASE MODELS
+   * onecc doesn't support multiple base models.
+   * However, OneExplorer will show the config node below multiple base models
+   * to prevent a case that users cannot find their faulty config files on ONE explorer.
+   */
+  private static parseBaseModels = (uri: vscode.Uri, iniObj: object): vscode.Uri[] => {
+     const baseModels = ['tflite','pb','onnx']
+     .filter(ext=>iniObj[`one-import-${ext}` as keyof typeof iniObj] ?.['input_path'] !== undefined)
+     .map(ext => iniObj[`one-import-${ext}` as keyof typeof iniObj] ?.['input_path'])
+     .map(relpath=>path.join(path.dirname(uri.fsPath),relpath))
```
hyunsik-yoon, line 135: From the example in https://github.com/Samsung/ONE-vscode/issues/644#issuecomment-1131374528,
- cfg uri.fspath = /home/eric/m/src/model/onecc-test/cfg/inception_v3_conv2d.0to128.cfg
- relpath = ./inception_v3/2020-01-31/inception_v3_conv2d.0to128.tflite

It seems that joining cfg  uri.fspath and relpath in L135 produces non-existing path. 😿 

If we need to assume some constraint regarding path of cfg and base model, could you write that as comment?

dayo09: Oh sure. I can use RealPath to resolve the problem. Thanks for finding that out.

dayo09: @hyunsik-yoon I examined the point and found out that your `OneccRunner` runs onecc on `config file's directory`.
I think that's why I didn't confront an error while testing the current PR.

https://github.com/Samsung/ONE-vscode/blob/5c3c1f636c6ac115b2c8fb46d49bf85fd66c660d/src/OneExplorer/OneExplorer.ts#L477

So it resolves the problem about cwd problem, and it seems okay.

hyunsik-yoon: Directory issue is currently under discussion in other issue, so I guess we'd better pass the issue this time. 😸

src/Tests/OneExplorer/ConfigObject.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  destructor() {
+    fs.rmdirSync(this.testDir, {recursive: true});
+  }
+};
+
+suite('OneExplorer', function() {
+  suite('ConfigObject', function() {
+    suite('#createConfigObj()', function() {
+      test('Returns null when file read failed', function() {
```
hyunsik-yoon, line 66: I think this is negative test. It would be better to use different naming to make SE process easier later.

In backend repo, we use `NEG: ` prefix for negative test like the following:

```suggestion
      test('NEG: Returns null when file read failed', function() {
```

src/OneExplorer/ConfigObject.ts
```diff
@@ -0,0 +1,194 @@
...
+    return ini.parse(configRaw.toString());
+  }
+
+  /**
+   * Parse base models written in the ini object and return the absolute path.
+   *
+   * ABOUT MULTIPLE BASE MODELS
+   * onecc doesn't support multiple base models.
+   * However, OneExplorer will show the config node below multiple base models
+   * to prevent a case that users cannot find their faulty config files on ONE explorer.
```
hyunsik-yoon, line 129: As far as I understand, you mean that we should show such falty one cfg file in the view even though that has error with base model, because that will give user chance to fix the cfg file. I wonder I understand your intention correctly.

src/Tests/OneExplorer/ConfigObject.test.ts
hyunsik-yoon: Did you maybe forgot to change this? 😸

hyunsik-yoon:
Format error... T.T

src/Tests/OneExplorer/ConfigObject.test.ts
dayo09: Oh sure. I will do this now 😄

src/OneExplorer/ConfigObject.ts
```diff
@@ -0,0 +1,215 @@
...
+      {section: 'one-optimize', key: 'output_path'},
+      {section: 'one-quantize', key: 'input_path'},
+      {section: 'one-quantize', key: 'output_path'},
+      {
+        section: 'one-codegen',
+        key: 'command',
+        filt: (str: string): string[] => {
+          return str.split(' ').filter(
+              e => path.extname(e) === '.tvn' || path.extname(e) === '.circle');
+        }
```
yunjayh, line 187: Can you please add todo to apply  `Executor.getExecutableExt()` method here? If you don't mind, I'll update after some days.

dayo09: Thanks for the suggestion. I will apply them. (I think they are many)

src/Tests/OneExplorer/ConfigObject.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        const content = `
+[onecc]
+one-import-tflite=True
+one-quantize=True
+[one-import-tflite]
+input_path=/dummy/dummy/../../${baseModelName}
+output_path=/dummy/dummy/../../${derivedModelName1}
+[one-quantize]
+input_path=////${derivedModelName1}
+output_path=/dummy/dummy/../..//${derivedModelName2}
```
hyunsik-yoon, line 205: These example seems confusing to me. 🐱 
Since path starts from `/`, I think these should be interpreted as absolute paths.

For example,
```
$ pwd
/home/eric/m/work/3/ONE-vscode
$ realpath -m ../dummy/dummy/..//model.tflite   # relative path
/home/eric/m/work/3/dummy/model.tflite
$ realpath -m ///dummy/dummy/..//model.tflite  # absolute path
/dummy/model.tflite
```

When I try to print `configObj!.obj.baseModels` and `configObj!.obj.derivedModels` at L226, they are all in `/tmp/one-vscode.test/ConfigObject/4` and I am not sure they are correct. (e.g., `'/tmp/one-vscode.test/ConfigObject/ConfigObject/4/model.tflite'`)

When a path is in absolute path, I think we should not join like L145.

src/Tests/OneExplorer/ConfigObject.test.ts
```diff
@@ -0,0 +1,235 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {assert} from 'chai';
+import * as fs from 'fs';
+import * as vscode from 'vscode';
+import {FillOptions} from '../../Circlereader/circle-analysis/circle/fill-options';
```
hyunsik-yoon, line 20: This seems not used.

src/OneExplorer/ConfigObject.ts
dayo09: Well I checked the code but it will not only contain executable exts, I will also involve those all derived models. I will add `// Get ext lists from backend` instead. :-D

yunjayh: Oh, Thank you for your check. 🙇 That seems a better comment.

src/Tests/OneExplorer/ConfigObject.test.ts
dayo09: Sure. Actually that went through my mind which I just ignored... Let me work on that.

dayo09: > When I try to print configObj!.obj.baseModels and configObj!.obj.derivedModels at L226, they are all in /tmp/one-vscode.test/ConfigObject/4 and I am not sure they are correct. (e.g., '/tmp/one-vscode.test/ConfigObject/ConfigObject/4/model.tflite')

It's right. I created a temporal directory with numbers for the test debug.

dayo09: I have updated the code to work by [Taking as an absolute path when it starts with /](https://github.com/Samsung/ONE-vscode/pull/681/commits/2899ea77e812533a91e90e905e464fb630726a9f)

src/Tests/OneExplorer/ConfigObject.test.ts
```diff
@@ -0,0 +1,328 @@
...
+          assert.isFalse(configObj!.obj.derivedModels.map(derivedModel => derivedModel.fsPath)
+                             .includes(derivedModelPath1));
+          assert.isFalse(configObj!.obj.derivedModels.map(derivedModel => derivedModel.fsPath)
+                             .includes(derivedModelPath2));
+          ;
+        }
+      });
+
+
+      test('Parse config with absolute paths', function() {
```
dayo09, line 280: It will regard /dummy/... as an absolute path now. It's a positive test :-D @hyunsik-yoon

hyunsik-yoon (approved):
Thanks. LGTM

yunjayh (approved):

PR #663, opened 2022-05-17
Use singleton logger

hyunsik-yoon:
This makes logger as a singleton.
This removes problem mentioned at https://github.com/Samsung/ONE-vscode/issues/662#issuecomment-1128638648.

For #662

ONE-vscode-DCO-1.0-Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

Issue #662, opened 2022-05-17
Investigate and apply logging


hyunsik-yoon:
We are using `Logger` and `console.log`.
Let's find the answer for the following:
- How are they different?
- What is good way?

hyunsik-yoon:
After installing vsix files, I couldn't find where I see output of `console.log()`.

From https://stackoverflow.com/a/69216474,

> Just open vscode and go to menu "Help"->"Toggle Developer Tools" and the console is displayed on rigth window.

I tried and found the log there. The messages are mixed with messages from other extensions.

Screen shot ===> [consoleLog.zip](https://github.com/Samsung/ONE-vscode/files/8706963/consoleLog.zip)

So `console.log()` is for **debugging-purpose**, not good for end-user.

Anyways, for our development purpose, it would be better to use some prefix, e.g., `one-vscode` so that we can filter messages in "Toggle Developer Tools".

hyunsik-yoon:
Also after installing vsix files, I found weird things.

There are two `ONE-vscode`  output channel.     =========> [Screenshot : onevscode.logging.zip](https://github.com/Samsung/ONE-vscode/files/8707029/onevscode.logging.zip)

When we call `new util.Logger()`, new output channel is created. So we should consider, e.g., singleton for Logger.

hyunsik-yoon:
https://nodejs.org/en/docs/guides/dont-block-the-event-loop/ mentions that
node js has only a small number of threads:
1. one for event loop
  - this is where most code runs
2. others for worker pool
  - for heavy stuffs such as DNS  lookup, file read/write

I think logging is done by event loop thread. Therefore there would not be a typical multi-thread issue.

hyunsik-yoon:
#### Let's not use `console.log()`.
#### Let's use `Logger.outputWithTime()`

I will make PR for this.

hyunsik-yoon:
More opinions: 
- tag: https://github.com/Samsung/ONE-vscode/pull/682#pullrequestreview-979347813
- rename: https://github.com/Samsung/ONE-vscode/pull/682#discussion_r8769117314
- Android example: https://github.com/Samsung/ONE-vscode/pull/682#issuecomment-1132398728

hyunsik-yoon:
### Typical logging foirmat of vscode and vscode extensions
- `time module tag msg` where tag is, e.g., `info`, `warning`, etc.

hyunsik-yoon:
Can we get function name or class name automatically?

[...]

PR #682, opened 2022-05-19
Use Logger class and static methods

hyunsik-yoon:
Parent issue: #662 
Previously, we tried to use singleton Logger. (#663)
However, code needs `Logger.getInstance().outputWithTime(msg)`, which seems relatively long.

Instead let's use `Logger.outputWithTime(msg)`.

This commit introduces three changes.

1. Make methods in `Logger` as `static` to shorten its usage: e.g., `Logger.outputWithTime(msg)`
2. From 1, we don't have any `logger` object. 
    - Do not pass `logger` as param any more.
    - Do not keep `logger` as object variable
3. Change `console.log()` to `Logger.outputWithTime()`.

### DIFF
--- a/src/Utils/Logger.ts
+++ b/src/Utils/Logger.ts
@@ -17,9 +17,11 @@
 import * as vscode from 'vscode';
 
 export class Logger {
-  outputChannel: vscode.OutputChannel;
-  firstFocus: boolean;
+  static outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('ONE-VSCode');
+  static firstFocus: boolean;
 
+  // TODO Remove code in comment. These lines were commented out for easy review.
+  /*
   private static logger: Logger|null = null;
 
   private constructor() {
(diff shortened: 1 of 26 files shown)


src/Utils/Logger.ts
```diff
@@ -34,27 +36,27 @@ export class Logger {
...
-    this.checkShow();
-    this.outputChannel.appendLine('[' + dateTime.toLocaleString() + '] ' + msg);
+    Logger.checkShow();
+    Logger.outputChannel.appendLine('[' + dateTime.toLocaleString() + '] ' + msg);
   }
 
-  public output(msg: string) {
-    this.checkShow();
-    this.outputChannel.append(msg);
+  public static output(msg: string) {
```
dayo09, line 53: (Q) What about `Logger.log()` or `Logger.logLine` `Logger.logWithTime`? For (1) It sounds better with verb at the front and (2) `log` sounds familiar than `output` (ex. console."log"). BTW I understand that Logger logs just like console.log but into its

jyoungyun (approved):
LGTM

(suggestion)
How about adding `TAG` about logs?
Or it is another way to create a separate logger for each TAG.

hyunsik-yoon:
@dayo09 @jyoungyun Thanks for good opinion. I think I have to do some research about logger. 😸

YongseopKim:
> How about adding TAG about logs?
Or it is another way to create a separate logger for each TAG.

It's very good idea. Android does like [this](https://developer.android.com/studio/debug/am-logcat?hl=ko#WriteLogs).

hyunsik-yoon:
Let me close this. Please refer to #700.

PR #686, opened 2022-05-20
Revise to use common getNonce

seanshpark:
This will revise to use getNonce in Utils/external/Nonce.


hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

PR #692, opened 2022-05-20
[Device] Add ExecutionEnv to specify Device

struss:
Follow From https://github.com/Samsung/ONE-vscode/issues/585#issuecomment-1132476919

This commit will Add ExecutionEnv API to use in DeviceView.


struss:
TODO on this
 - [ ] need to explain about `ExecutionEnv`.
 - [ ] Add Test about this API Use.

YongseopKim:
For reviewers, it would be great to read https://github.com/Samsung/ONE-vscode/issues/585#issuecomment-1132476919 before reviewing.

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,92 @@
...
+  host(): string;
+  listExecutableExt(): string[];
+  isExecutionEnvAvail(name: string): boolean;
+  getEnableEnvList(): string[];
+  getInferenceCmd(model_path: string, _options?: string[]): Command;
+}
+
+class ToolchainExecutorEnv implements ExecutionEnv {
+  envName: string;  // Specified name for Each Env
+  simulator: Executor;
```
dayo09, line 33: (Q) Will this property be valid only if it's a simulator?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,92 @@
...
+  constructor(backendName: string, sim: Executor) {
+    this.envName = backendName;
+    this.simulator = sim;
+  }
+  name(): string {
+    return this.envName;
+  }
+  host(): string {
+    return 'x86';
```
dayo09, line 42: Maybe `os.platform() / process.platform()` can be used here

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+  getEnableEnvList(): string[] {
+    // TODO: if multiple Toolchain installed on
+    // we need to fix this to show all installed Executors
+    if (gToolchainEnvMap[this.envName].listInstalled()) {
+      return ['Simulator'];
+    }
+    return [];
+  }
+  listExecutableExt(): string[] {
```
dayo09, line 58: How about `getListExecutableExt`?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,92 @@
...
+import {Command} from '../Backend/Command';
+import {Executor} from '../Backend/Executor';
+import {gToolchainEnvMap} from '../Toolchain/ToolchainEnv';
+
+// Specify Device with 'name' on `getEnableEnvList()` and `ExecutionEnv`.
+interface ExecutionEnv {
+  name(): string;
+  host(): string;
+  listExecutableExt(): string[];
+  isExecutionEnvAvail(name: string): boolean;
```
dayo09, line 26: How about `isAvailable`?
I think it's easy to read if not repeating the class name.

src/Execute/ExecutionEnv.ts
dayo09: If it's only about a simulator, how about 'SimulatorEnv' or something instead of 'ToolchainExecutorEnv'?

struss: no, I just name this as a `simulator`, but this property comes from `toolchain` executor. if any name for this, please suggest!

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+  host(): string {
+    return platform();
+  }
+  isAvailable(name: string): boolean {
+    if (gToolchainEnvMap[this.envName].listInstalled()) {
+      return this.getEnableEnvList().includes(name);
+    }
+    return false;
+  }
+  getEnableEnvList(): string[] {
```
hyunsik-yoon, line 51: I am not sure what this method is for and meaning out return value. Could you write some comment about this method?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.envName = backendName;
+    this.simulator = sim;
+  }
+  name(): string {
+    return this.envName;
+  }
+  host(): string {
+    return platform();
+  }
+  isAvailable(name: string): boolean {
```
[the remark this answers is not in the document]
hyunsik-yoon: Also, looking into L51, which returns `['Simulator']`, should we call `isAvailable('Simulator')`? 
I am not sure about the usage of this method. maybe some comment for this method would be helpful.

src/Execute/ExecutionEnv.ts
struss: > If it's only about a simulator,

um... I just name this as simulator, but I thought this comes from all executor comes from `toolchain` that is installed on.

> I think that TizenEnv kinda executes the files from compiled the toolchain, 

I think this is right, as I know.

>so ToolchainExecutorEnv sounds to be a superset of TizenEnv.

um.... I thought `ToolchainExecutorEnv` based on toolchain, and `TizenEnv` based on tizen.
so, follow from this, what about `TizenExectuionEnv`?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return [];
+  }
+  getListExecutableExt(): string[] {
+    return this.simulator.getExecutableExt();
+  }
+  getInferenceCmd(modelPath: string, _options?: string[]): Command {
+    return this.simulator.runInference(modelPath);
+  }
+}
```
hyunsik-yoon, line 65: If you enter a newline between methods, it would be easier to read. :-)

src/Execute/ExecutionEnv.ts
struss: Oh, I make a name so weird . I will fix this and comment about it.

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+  simulator: Executor;
+  constructor(backendName: string, sim: Executor) {
+    this.envName = backendName;
+    this.simulator = sim;
+  }
+  name(): string {
+    return this.envName;
+  }
+  host(): string {
+    return platform();
```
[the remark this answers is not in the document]
struss: I think this will be used on next step, described in https://github.com/Samsung/ONE-vscode/issues/585#issuecomment-1132476919

> Example) a toolchain is installed on local pc and it has a simulator on pc and npu runner for tv.
> * On `Target Devices`, there could be two targets
> * 1. x86/triv/simulator
> * 2. arm(or tv)/triv/one-infer

src/Execute/ExecutionEnv.ts
hyunsik-yoon: > um... I just name this as simulator, but I thought this comes from all executor comes from toolchain that is installed on.

Toolchain could contain many things (simulator, emulator, ... runtume, ...) depending on backend vendors.
Another approach could be renaming this to `SimulatorEnv`. 😸 
So, I guess specific name like `SimulatorEnv` might be better.... we might ask backend vendors to implement the details of its method.

hyunsik-yoon: While looking at `TizenEnv`... one question came up to my mind. 

If a company named FOO wants to create a backend extension for their special emulator `FooEmulator` which I guess implements `interface Executor`, 
should we have `FooExecutionEnv` in ONE-vscode extension to support `FooEmulator`?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+interface ExecutionEnv {
+  name(): string;
+  host(): string;
+  getEnableEnvList(): string[];
+  isAvailable(name: string): boolean;
+  getListExecutableExt(): string[];
+  getInferenceCmd(model_path: string, _options?: string[]): Command;
+}
+
+class ToolchainExecutorEnv implements ExecutionEnv {
```
jyoungyun, line 32: ToolchainExecutorEnv means simulator, doesn't it? Then how about using `ToolchainSimulatorEnv`?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+  getListExecutableExt(): string[] {
+    return this.simulator.getExecutableExt();
+  }
+  getInferenceCmd(modelPath: string, _options?: string[]): Command {
+    return this.simulator.runInference(modelPath);
+  }
+}
+
+class TizenEnv implements ExecutionEnv {
```
jyoungyun, line 67: (optional)
How about separating this class into separate files?
If devices are managed separately from the `ExecutionEnv` interface, it will be easier to identify modifications to each device.

e.g.,
`src/Executor or Device/[ToolchainSimulatorEnv, TizenEnv]`

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,93 @@
...
+import {gToolchainEnvMap} from '../Toolchain/ToolchainEnv';
+
+// Specify Device with 'name' on `getEnableEnvList()` and `ExecutionEnv`.
+interface ExecutionEnv {
+  name(): string;
+  host(): string;
+  getEnableEnvList(): string[];
+  isAvailable(name: string): boolean;
+  getListExecutableExt(): string[];
+  getInferenceCmd(model_path: string, _options?: string[]): Command;
```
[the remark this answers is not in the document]
struss: I will add a comment about this.

src/Execute/ExecutionEnv.ts
struss: umm.... I think that `tizenEnv` makes some confusing point. I thought this could be a base class like `compilerBase` or 'executorEnv' to use in backend. I think I need to change this as a `ExecutionEnvBase`.

> should we have `FooExecutionEnv` in ONE-vscode extension to support `FooEmulator`?

then if someone wants to add a new emulator on this, just create new class extends `ExecutionEnvBase` on their backend.

src/Execute/ExecutionEnv.ts
YongseopKim: Good point. This part could be replaced by https://nodejs.org/api/os.html .

src/Backend/Backend.ts
```diff
@@ -27,9 +28,18 @@ interface BackendMap {
...
 }
 
+interface ExecutionEnvMap {
+  [key: string]: ExecutionEnv;
+}
+
 // List of backend extensions registered
 let globalBackendMap: BackendMap = {};
 
+// As Execution Env will Contains components below
```
YongseopKim, line 38: I'd like you to explain more.

src/Backend/Backend.ts
```diff
@@ -40,11 +50,23 @@ function backendRegistrationApi() {
...
+      // This will register case 1, ExecutionEnv with Backend Executor
+      // Executor will handled as a Simulator.
+      const executor = backend.executor();
+      if (executor) {
+        globalExecutionEnvMap[backend.name()] = new ToolchainExecutorEnv(backend.name(), executor);
+      }
       console.log(`Backend ${backendName} was registered into ONE-vscode.`);
+    },
+    // This will register case 2, ExecutionEnv with specific HW and specific SW(Device)
+    registerExecutionEnv(execEnv: ExecutionEnv) {
```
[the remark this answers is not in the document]
YongseopKim: But, is it necessary for `ExecutionEnv` as `interface` that is a kind of abstract/parent class.

src/Execute/ExecutionEnv.ts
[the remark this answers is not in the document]
YongseopKim: Why is it necessary for `static`?

src/Execute/ExecutionEnv.ts
YongseopKim: The `connect` is meaning (`ssh connect` or `sdb connect`)? or (`ssh connect` and `sdb connect`)

src/Execute/ExecutionEnv.ts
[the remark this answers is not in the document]
YongseopKim: @struss you can mix arch(), platform() and release()

struss:
after conversation with @hyunsik-yoon , It looks more natural to define `ExecutionEnv` as a two type.

1. SWExecutionEnv
    - Defined with `Executor` and `toolchian`.

2. HWExecutionEnv
    - Defined by Backend provider. this will use specific HW and SW to use real device.

so, I will revise this PR with this and make a comment for each functions.

struss:
After Discussion, I will revise this. Sorry for late.

struss:
This will be changed to other PRs using #744

PR #697, opened 2022-05-23
[DRAFT] Whole new cfg editor UI

llFreetimell:
This draft is for whole new cfg editor UI.

---

#### Related Issues

- #698

---

#### Before & After

| UI | Progress bar | image | image |
| UI | Summary | image | :factory_worker: WIP |
| UI | Import Step | image |  import |
| UI | Optimize Step | :x: NYI | optimize |
| UI | Quantize Step | :x: NYI | :factory_worker: WIP |
| UI | Codegen Step | image | image |
| UI | Profile Step | image | image |
| New Feature | Sync Changes (Editor -> .cfg) | :x: NYI | :heavy_plus_sign: Changes from editor are applied to cfg file |
| Bugfix | Sync Changes (.cfg -> Editor) | :lady_beetle: `onecc` section was not supported<br> | :hammer_and_wrench: `onecc` section is now supported |


src/extension.ts
```diff
@@ -96,22 +96,22 @@ export function activate(context: vscode.ExtensionContext) {
...
-  context.subscriptions.push(disposableToggleCodelens);
-
-  let codelens = new CodelensProvider(context);
-  let disposableCodelens = vscode.languages.registerCodeLensProvider('ini', codelens);
-  context.subscriptions.push(disposableCodelens);
-
-  let hover = new HoverProvider();
-  let disposableHover = vscode.languages.registerHoverProvider('ini', hover);
-  context.subscriptions.push(disposableHover);
+  // let disposableToggleCodelens =
```
YongseopKim, line 99: Is it what we talked about `deprecation`?

package.json
```diff
@@ -248,20 +248,6 @@
           ".cfg"
         ]
       }
-    ],
```
YongseopKim, line 251: `deprecated`?

media/CfgEditor/cfgeditor.js
```diff
@@ -13,32 +13,8 @@
...
- * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
- * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- */
-/*
-This code refers to
-https://github.com/microsoft/vscode-webview-ui-toolkit-samples/blob/b807107df40271e83ea6d36828357fdb10d71f12/default/hello-world/webview-ui/main.js
-*/
+
```
YongseopKim, line 16: Q. How about transforming to `ts`?

YongseopKim: Is it possible? or is it making you feel difficult? (if so, I would do it after landing this code.)

src/extension.ts
llFreetimell: Yes, but i think this is out of scope for this draft :)
Please ignore these removals!

media/CfgEditor/cfgeditor.js
llFreetimell: At first, I am saying that I can be wrong because I am not good at HTML and web.
So if there are something wrong, please let me know :)

---

AKAIK, HTML only understands javascript, not typescript.
Therefore, if we want to use typescript in HTML, we should compile `.ts` file first and then include the compiled `.js` file.
However, current structure for our repository seems not allow to do so :(
Briefly speaking, current structure is as following.

- `src` directory : includes **dynamic** `ts` files for using VSCode extension APIs
- `media` directory : includes **static** `html`, `css`, `js` files which are loaded on webview

With this status, let's say that I created `cfgeditor.ts` instead of `cfgeditor.js`.
Then how should I handle this?

- If I put `cfgeditor.ts` to `src` directory, it is not dynamic file.
- If I put `cfgeditor.ts` to `static` directory, the file is not compiled so we cannot use it directly.

In conclusion, I think that's why MS introduced sample VSCode extension as current structure.
(https://github.com/microsoft/vscode-extension-samples/tree/main/webview-view-sample)

If you have a good idea to transforming these `.js` files to `.ts` files, let's discuss about it together :D

media/CfgEditor/index.html
```diff
@@ -0,0 +1,302 @@
...
+        </div>
+        <div class="options" id="optionImport">
+            <div class="prerequisite">
+                <div class="title">
+                    Basic Options
+                </div>
+                <div class="option">
+                    <vscode-radio-group id="importInputModelType">
+                        <label slot="label">Input Model Type</label>
+                        <vscode-radio value="pb" checked="true">pb</vscode-radio>
```
llFreetimell, line 91: TODO / Main branch bugfix

```suggestion
                        <vscode-radio value="pb">pb</vscode-radio>
```

media/CfgEditor/cfgeditor.js
YongseopKim: Now I read your kind comment. Thanks!

llFreetimell:
This draft is now tool old to apply current main branch.
So let's close this PR and then create new one if needed! :D

PR #700, opened 2022-05-24
Apply static logger

hyunsik-yoon:
This applies static logger.

1. Make all class var and method as static.
2. Remove `logger` param being passed around and `import` 
3. Remove deprecated methods
4. Classes now have logging tag

While doing the above, one method `appendLine` was added into Logger
because ToolRunner needs both `append` and `appendLine`.

For #662 

### DIFF
--- a/src/Project/BuilderCfgFile.ts
+++ b/src/Project/BuilderCfgFile.ts
@@ -152,7 +152,7 @@ export class BuilderCfgFile extends EventEmitter implements helpers.FileSelector
 
     console.log('importTFlite = ', importTFlite);
     this.jobOwner.addJob(importTFlite);
-    Logger.append('Add Import: ' + inputModel);
+    Logger.info(this.tag, 'Add Import: ' + inputModel);
   }
 
   private cfgImportOnnx(prop: any) {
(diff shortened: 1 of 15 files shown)


hyunsik-yoon:
image

YongseopKim (approved):
LGTM 👍

src/Project/BuilderCfgFile.ts
```diff
@@ -153,7 +152,7 @@ export class BuilderCfgFile extends EventEmitter implements helpers.FileSelector
 
     console.log('importTFlite = ', importTFlite);
     this.jobOwner.addJob(importTFlite);
-    this.logger.outputLine('Add Import: ' + inputModel);
```
yunjayh, line 156: IMHO, every `Logger.append(..)` in each function should replaced to `Logger.info(this.tag, ...)` since it cannot guarantee its information has been printed before. 

For example, this line cannot sure that `cfgImportTf` already executed before `cfgImportTflite`.. If I'm misunderstood something, please let me know.. 😺

hyunsik-yoon: I changed `Logger.append` in `BuildCfgFile.ts` to `Logger.info`.

Log screen seems OK.

image

jyoungyun (approved):
LGTM

yunjayh (approved):

YongseopKim (approved):

PR #704, opened 2022-05-24
[Toolchain] Add success and failure callback to job request

jyoungyun:
This commit adds success and failure callback to job request.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #707, opened 2022-05-24
[Toolchain] Uninstall toolchain

jyoungyun:
This commit uninstalls the installed toolchain.
After uninstalled the toolchain, the toolchain view will be updated
automatically.

### DIFF
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -28,6 +28,7 @@ export class ToolchainNode extends vscode.TreeItem {
       public readonly label: string,
       public readonly collapsibleState: vscode.TreeItemCollapsibleState,
       public readonly type: NodeType,
+      public readonly backend?: string,
       public readonly toolchain?: Toolchain,
   ) {
     super(label, collapsibleState);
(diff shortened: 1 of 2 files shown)


jyoungyun:
Dependencies on #704

src/Toolchain/ToolchainProvider.ts
```diff
@@ -28,14 +28,15 @@ export class ToolchainNode extends vscode.TreeItem {
       public readonly label: string,
       public readonly collapsibleState: vscode.TreeItemCollapsibleState,
       public readonly type: NodeType,
+      public readonly backend?: string,
```
YongseopKim, line 31: Q. Any reason to assign `backend` as `string | undefined`? If you assign it as just the only `string`, we don't have to care about `backend` is undefined in ts.

jyoungyun: The `backend` is optional parameter so it can either have a value based on the type defined or its value can be `undefined`. However, when I use `backend: string | undefined` code, this value must be inputted. But `NodeType.backend` does not use `backend` string. So I use it as optional parameter.

jyoungyun:
PTAL /cc @Samsung/one-vscode

src/Toolchain/ToolchainProvider.ts
YongseopKim: I understand your point.

YongseopKim (approved):
LGTM!

IHMO, items in the future
- ToolchainNode seems to be divided to something for backend and something for toolchain

hyunsik-yoon (approved):

jyoungyun:
Um, ToolchainNode can't be separated for backend and toolchain. Because these item should be shared in this tree view. So it has the same root interface to share these one. I can separate the 3 class for parent and 2 children for backend and toolchain that inherit parent class. However, in most tree view example codes, it is designed to implement both parent and children nodes as one node. So I'm not sure if it really needs to be modified.... Anyway, I will consider about your suggestion more. :)

YongseopKim:
Trial: https://github.com/Samsung/ONE-vscode/pull/717

PR #718, opened 2022-05-26
[OneExplorer] Validate input for rename

dayo09:
This commit validates an input for renaming.
It checks (1) ext, (2) file already exists.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -188,7 +188,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
     const validateInputPath = (newname: string): string|undefined => {
       const oldpath = oneNode.node.path;
       const dirpath = path.dirname(oneNode.node.uri.fsPath);
-      const newpath: string = path.join(${dirpath}, ${newname});
+      const newpath: string = path.join(dirpath, newname);
 
       if (path.extname(newpath) !== path.extname(oldpath)) {
         return `File ext must be (${path.extname(oldpath)})`;
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -195,7 +195,8 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       }
 
       if (fs.existsSync(newpath)) {
-        return `A file or folder ${newpath} already exists at this location. Please choose a different name.`;
+        return `A file or folder ${
+            newpath} already exists at this location. Please choose a different name.`;
       }
     };
 
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -185,11 +185,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       warningMessage = 'WARNING: Renaming may result in some unexpected changes on the tree view.';
     }
 
+    const validateInputPath = (newname: string): string|undefined => {
+      const oldpath = oneNode.node.path;
+      const dirpath = path.dirname(oneNode.node.uri.fsPath);
+      const newpath: string = `${dirpath}/${newname}`;
+
+      if (path.extname(newpath) !== path.extname(oldpath)) {
+        return `File ext must be (${path.extname(oldpath)})`;
+      }
+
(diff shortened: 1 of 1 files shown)


dayo09:
@yunjayh 
The feature from [this conversation](https://github.com/Samsung/ONE-vscode/pull/674#discussion_r882313698) :-D

> 
> How about giving proper extension name if it's not given by user? I cannot imagine the danger if we apply following change.
> 
> @[dayo09](https://github.com/dayo09) dayo09 [4 hours ago](https://github.com/Samsung/ONE-vscode/pull/674#discussion_r882321163) • 
> IMO, it's a little bit too much... Maybe it will bring about much confusion to user. Well, how about just putting an error message instead in case of user puts incorrect ext?
> 
> @[yunjayh](https://github.com/yunjayh) yunjayh [4 hours ago](https://github.com/Samsung/ONE-vscode/pull/674#discussion_r882324952)
> IMO, it's a little bit too much... Maybe it will bring about much confusion to user.
> 
> I agree with you. Since there is a warning message at the bottom of input box, user would pay attention with his/her file naming. IMHO Defining the situation when the error message is popping up will be pretty difficult.
> Then, never mind about the above suggestion!
> 
> @[dayo09](https://github.com/dayo09) dayo09 [3 hours ago](https://github.com/Samsung/ONE-vscode/pull/674#discussion_r882327319)
> @yunjayh I think we can use 'validateInput' API to check if user has given a right ext. I will apply that on the next pr. thanks for your opinion.

dayo09:
[Here](https://github.sec.samsung.net/RS7-RuntimeNTools/secret/issues/187#issuecomment-1523764) is a gif link. PTAL.

src/OneExplorer/OneExplorer.ts
```diff
@@ -185,11 +185,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+      const oldpath = oneNode.node.path;
+      const dirpath = path.dirname(oneNode.node.uri.fsPath);
+      const newpath: string = `${dirpath}/${newname}`;
+
+      if (path.extname(newpath) !== path.extname(oldpath)) {
+        return `File ext must be (${path.extname(oldpath)})`;
+      }
+
+      if (fs.existsSync(newpath)) {
+        return `Invalid: Path already exists!`;
```
yunjayh, line 198:
```suggestion
        return `Invalid: File already exists!`;
```

It looks not to renaming directory. So, File already exists would be more clear for user IMO

src/OneExplorer/OneExplorer.ts
```diff
@@ -185,11 +185,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
       warningMessage = 'WARNING: Renaming may result in some unexpected changes on the tree view.';
     }
 
+    const validateInputPath = (newname: string): string|undefined => {
+      const oldpath = oneNode.node.path;
+      const dirpath = path.dirname(oneNode.node.uri.fsPath);
+      const newpath: string = `${dirpath}/${newname}`;
```
hyunsik-yoon, line 191: (optional) how about using `path.join` for future when we support Windows? 😸 

```suggestion
      const newpath: string = path.join(${dirpath}, ${newname});
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -185,11 +185,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+      const oldpath = oneNode.node.path;
+      const dirpath = path.dirname(oneNode.node.uri.fsPath);
+      const newpath: string = path.join(dirpath, newname);
+
+      if (path.extname(newpath) !== path.extname(oldpath)) {
+        return `File ext must be (${path.extname(oldpath)})`;
+      }
+
+      if (fs.existsSync(newpath)) {
+        return `Invalid: File already exists!`;
```
dayo09, line 198:
```suggestion
        return `A file or folder ${newpath} already exists at this location. Please choose a different name.`;
```

dayo09: It's what the vscode default explorer says. Let's sync with theirs.

hyunsik-yoon (approved):

yunjayh (approved):

hyunsik-yoon (approved):

PR #716, opened 2022-05-26
DRAFT: Cancel running onecc

hyunsik-yoon:
This draft is to terminate onecc process by pressing 'Cancel' button.

For #715

ONE-vscode-DCO-1.0-Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #720, opened 2022-05-26
Introduce ToolRunner.kill() method

hyunsik-yoon:
This commit adds `kill()` method that sends signals to kill spawned process.

For #716


src/Project/ToolRunner.ts
```diff
@@ -57,6 +87,35 @@ export class ToolRunner {
...
+   * Function to kill child process
+   */
+  public kill(): boolean {
+    this.killedByMe = false;
+
+    if (this.child === undefined || this.child.killed) {
+      throw Error('No process to kill');
+    }
+
+    const signals: NodeJS.Signals[] = ['SIGINT', 'SIGTERM', 'SIGKILL'];
```
llFreetimell, line 100: AFAIK, `SIGTERM` is for exiting program as normal as possible.
But pressing `Cancel` button do not seem to be normal I think.
Could you let me know why `SIGTERM` is included here? :)

src/Project/ToolRunner.ts
```diff
@@ -25,29 +25,59 @@ const which = require('which');
...
+// successfully killed by calling kill() method
+export type SelfKilled = 'SelfKilled';
 
 export class ToolRunner {
   tag = this.constructor.name;  // logging tag
 
+  private child: cp.ChildProcessWithoutNullStreams|undefined = undefined;
+
+  // When the spawned process was killed by kill() method, set this true
+  private killedByMe = false;
```
llFreetimell, line 37: (just for fun) Suicide? :smile:

hyunsik-yoon: kinda of. but more specially, parent(`ToolRunner`) kills its child. 😢

src/Project/ToolRunner.ts
hyunsik-yoon: I referred to this article: https://www.baeldung.com/linux/sigint-and-other-termination-signals#how-sigint-relates-to-sigterm-sigquit-and-sigkill

I didn't try SIGQUIT since it generates core dump.

src/Project/ToolRunner.ts
```diff
@@ -25,29 +25,59 @@ const which = require('which');
...
+      }
+
+      // when child exited
       let codestr = code.toString();
       console.log('child process exited with code ' + codestr);
       if (codestr === '0') {
         Logger.info(this.tag, 'Build Success.');
         Logger.appendLine('');
-        resolve(codestr);
+        resolve('0');
```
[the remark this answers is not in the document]
hyunsik-yoon: It's from personal taste.. 
when I first see `resolve(codestr)` I need to find out what `codestr` is.... so I have to search the code to find out `codestr` is actually `0`. Instead, we could simply write `resolve('0')`. 
However, it is not a big matter and this change is not related to this PR. I will revert this change.

jyoungyun:
I found `CancellationToken` in vscode. This token is passed to an asynchronous or long running operation to request cancellation.

[CancellationToken](https://vshaxe.github.io/vscode-extern/vscode/CancellationToken.html#:~:text=package%20vscode,use%20a%20%7B%40link%20CancellationTokenSource%7D.)

And here are several examples of this.
https://github.com/Microsoft/vscode/issues/22318#issuecomment-316459749
https://github.com/microsoft/vscode-docker/blob/main/src/utils/spawnAsync.ts

I'm not sure how much code modification is needed to use this. If it costs too much, your code is also good.

In this PR, it doesn't consider multi requests at the same time. `this.child` is only one variable and is shared between different processes.

hyunsik-yoon:
@jyoungyun 
I will check. Thanks.

src/Project/ToolRunner.ts
```diff
@@ -25,29 +25,59 @@ const which = require('which');
...
+      // when child was terminated due to a signal
+      if (code === null) {
+        Logger.debug(this.tag, `Child process was killed (signal: ${signal!})`);
+
+        if (this.killedByMe) {
+          resolve('SelfKilled');
+        } else {
+          reject(signal!);
+        }
+        return;
```
YongseopKim, line 71: This `return` is unnecessary.

src/Project/ToolRunner.ts
```diff
@@ -57,6 +87,35 @@ export class ToolRunner {
...
+    }
+
+    const signals: NodeJS.Signals[] = ['SIGINT', 'SIGTERM', 'SIGKILL'];
+
+    for (const signal of signals) {
+      if (this.child!.kill(signal)) {
+        this.killedByMe = true;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
YongseopKim: I guess that
1. this code tries killing with a signal
2. if successaful, break;
3. else, it treies killing with another signals

src/Project/ToolRunner.ts
```diff
@@ -81,21 +140,22 @@ export class ToolRunner {
   }
 
   public getRunner(name: string, tool: string, toolargs: ToolArgs, path: string, root?: boolean) {
+    this.killedByMe = false;
+
     return new Promise<string>((resolve, reject) => {
       Logger.info(this.tag, 'Running: ' + name);
-      let cmd = undefined;
       if (root) {
```
YongseopKim, line 147: For my record, this code could be refactor.

hyunsik-yoon: I am not fully understand what this comment meant for this PR. Could you elaborate more about what I need to do?

src/Project/ToolRunner.ts
hyunsik-yoon: This is necessary. L63 to L71 handles when `code === null`. Code from L75 handles when `code !== null`

hyunsik-yoon:
@jyoungyun 

> In this PR, it doesn't consider multi requests at the same time. this.child is only one variable and is shared between different processes. Please check this part. :)

To avoid this, I used `isRunning()` method. I also added a test for lots of simultaneous calling of `getRunner()`.

PTAL

jyoungyun (approved):
LGTM

llFreetimell (approved):
LGTM

hyunsik-yoon:
@llFreetimell 
Since @YongseopKim is on his vacation, could you merge? Thanks.

@YongseopKim
Please leave comment in this PR even after this is merged. I will monitor this PR for a while.

src/Project/ToolRunner.ts
YongseopKim: Sorry for the very late reply. Now I read this comment. I'd like to apologize. 😭 

My meaning is like that

as-is
```ts
      if (code === null) {
        Logger.debug(this.tag, `Child process was killed (signal: ${signal!})`);

        if (this.killedByMe) {
          resolve('SelfKilled');
        } else {
          reject(signal!);
        }
        return;
      }
```

to-be
```ts
      if (code === null) {
        Logger.debug(this.tag, `Child process was killed (signal: ${signal!})`);

        if (this.killedByMe) {
          return resolve('SelfKilled');
        } else {
          return reject(signal!);
        }
      }
```

src/Project/ToolRunner.ts
YongseopKim: Sorry for my ambiguous comment. I have a plan to revise `cmd` for `root` and `not root`. That's why I wrote the comment. Please ignore it. Sorry 😃

PR #722, opened 2022-05-27
[Utils] Move MultiStepInput out of external dir

jyoungyun:
This commit moves MultiStepInput class out of external dir.
This commit contains the code chage about MultiStepInput.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

Draft PR: #699 (Big draft PR...)

PR #724, opened 2022-05-27
[Toolchain] Refresh toolchain list

jyoungyun:
This commit adds refresh button and refresh the toolchain list by
updating apt source list.

### DIFF
--- a/src/View/InstallQuickInput.ts
+++ b/src/View/InstallQuickInput.ts
@@ -19,10 +19,13 @@ import * as vscode from 'vscode';
 import {Toolchain} from '../Backend/Toolchain';
 import {JobCallback} from '../Project/Job';
 import {gToolchainEnvMap, ToolchainEnv} from '../Toolchain/ToolchainEnv';
+import {Balloon} from '../Utils/Balloon';
 import {Logger} from '../Utils/Logger';
 import {MultiStepInput} from '../Utils/MultiStepInput';
 
 export async function showInstallQuickInput() {
+  const logtag = showInstallQuickInput.name;
+
   interface State {
     title: string;
     step: number;
@@ -104,7 +107,7 @@ export async function showInstallQuickInput() {
       shouldResume: shouldResume
     });
     if (version instanceof InnerButton) {
-      console.log('press the refresh button');
+      Logger.info(logtag, 'press the refresh button');
       return (input: MultiStepInput) => updateBackend(input, state);
     }
     state.version = version;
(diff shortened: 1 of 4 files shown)


jyoungyun:
Depends on #722

src/View/InstallQuickInput.ts
```diff
@@ -87,17 +93,57 @@ export async function showInstallQuickInput() {
...
       title,
       step: 3,
       totalSteps: 3,
       placeholder: 'Pick toolchain version',
       items: versionGroups,
+      buttons: [updateButton],
       shouldResume: shouldResume
     });
+    if (version instanceof InnerButton) {
+      console.log('press the refresh button');
```
jyoungyun, line 106: I will change this log to `Logger` in other PR.

src/View/InstallQuickInput.ts
```diff
@@ -87,17 +93,57 @@ export async function showInstallQuickInput() {
...
+      return (input: MultiStepInput) => updateBackend(input, state);
```
hyunsik-yoon, line 107: It's kinda difficult to read this code to me. Could you explain a bit?
- `if (version instanceof InnerButton) {`
  - Is this case covers when `InnerButton` was pressed? If so, maybe other variable name instead of `version` would be better to read this code.
- L107 returns a function but code after L110 does not return function. I wonder if this is OK.

src/View/InstallQuickInput.ts
```diff
@@ -87,17 +93,57 @@ export async function showInstallQuickInput() {
...
+            // only the log is output and it goes to the `resolve` so that quick input
+            // can be seen normally.
+            resolve('fail');
+          });
+    });
+  }
+
+  async function updateBackend(input: MultiStepInput, state: Partial<State>) {
+    if (state.toolchainEnv === undefined || state.toolchainType === undefined) {
+      throw Error('toolchainenv  is undefined.');
```
hyunsik-yoon, line 134:
```suggestion
      throw Error('toolchainenv is undefined.');
```

src/View/InstallQuickInput.ts
jyoungyun: > Is this case covers when InnerButton was pressed? If so, maybe other variable name instead of version would be better to read this code.

This `showQuickPick` function returns promise and it returns the InnerButton if user clicks the refresh button and the `version` QuickPickItem if user clicks the version. I don't know how to name this variable. The L74 is in a similar situation and it uses `type` as name because it returns `type` QuickPickItem. What is a better name than version? Please give me an idea. :)

> L107 returns a function but code after L110 does not return function. I wonder if this is OK.

Yes, it's okay. Because this step is the last step for this installation.
When user selects the version number, this code jumps to L155 to install a new version.

The reason why the L107 returns the function is to show the version list again after updated apt cache. To show this multi step input UI in next scenario, it must need the state and this variable so I forwarded it.

dayo09 (approved):

hyunsik-yoon (approved):

PR #730, opened 2022-05-27
[OneExplorer] Add open cfg commands

dayo09:
This commit adds open cfg commands
- Open with Text Editor
- Open with ONE Cfg Editor (default)

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -529,7 +529,11 @@ export class OneExplorer {
     };
 
     subscribeCommands([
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -529,7 +529,11 @@ export class OneExplorer {
     };
 
     subscribeCommands([
-      vscode.commands.registerCommand('onevscode.open-cfg', (file) => this.openFile(file)),
```
dayo09, line 532: Bug: it won't work

PR #732, opened 2022-05-27
Revise activation events

dayo09:
This commit revises activation events.

NOTE: onStartupFinished will activate our extension
some time after VS Code starts up.
The other events are redundant.

---

FYI,
https://code.visualstudio.com/api/references/activation-events#onStartupFinished

We want to provide our users with our extension always. 
There are two options: '\*' and 'onStartupFinished' as activation events 'always'.
'onStartupFinished' will activate our extension some time aftser VS Code starts up, 
so that it doesn't slow down the vscode's start-up, unlike "*" will do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,13 +12,6 @@
   ],
   "extensionKind ": "workspace",
   "activationEvents": [
-    "onCommand:onevscode.build",
-    "onCommand:onevscode.import",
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

PR #743, opened 2022-05-31
[PartEditor] Revise backend list from extension

seanshpark:
This will revise to fetch backend list items from extension.

### DIFF
--- a/media/PartEditor/index.html
+++ b/media/PartEditor/index.html
@@ -39,9 +39,7 @@ limitations under the License.
   <div class="group-child">
     Backend:<br/>
     <select id="circle-be" name="BE" class="set_bk_list">
-      <option>(1) CPU</option>
-      <option>(2) ACL_CL</option>
-      <option>(3) NPU</option>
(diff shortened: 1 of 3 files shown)


seanshpark:
> Could you describe a bit about the usage of 0th item (default)? Thanks.

`(default)` is the default backend assigned to, when the node is not mentioned in the list.
ex)
```
[partition]
backends=cpu,acl_cl
default=cpu
comply=opcode

[OPCODE]
ADD=acl_cl
```

in this change, `(default)` won't be shown in the ListBox.
All the nodes will have key(index) of 0 for initial value, which is `default` backend.

hyunsik-yoon (approved):

seanshpark:
@hyunsik-yoon , as YOU are the ONLY one interested in this task, I would like to change to `1 approval` and move on.

dayo09 (approved):
LGTM

seanshpark:
@dayo09 , thanks for the review!

PR #744, opened 2022-06-02
[Draft] ExecutionEnv with view update

struss:
This commit is draft for ExecutionEnv

### DIFF
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -18,7 +18,6 @@ import {Command} from '../Backend/Command';
 
 interface ExecutionEnv {
   name(): string;
-  envInfo(): string;
   wrapCommand(command: Command): Command;
 }
 
@@ -32,11 +31,6 @@ interface ExecutionEnv {
 //    could split with '\n'.
 class ExecutionEnvBase implements ExecutionEnv {
   name(): string {
-    throw new Error('Method not implemented.');
-  }
-
-  // This will return some information for ExecutionEnv, like Tizen or some specified value to check Executor could run on this.
-  envInfo(): string {
     throw new Error('Method not implemented.');
   }
 
--- a/src/Execute/ExecutionEnvManager.ts
+++ b/src/Execute/ExecutionEnvManager.ts
@@ -28,13 +28,13 @@ import {ExecutionEnv} from './ExecutionEnv';
 // TODO Handle about remote case.
 // Currently, only local case Handled.
 
-interface ExectionEnvEnvManager {
+interface ExectionEnvManager {
   runInference(envName: string, executor: Executor, modelPath: string): void;
   runProfile(envName: string, executor: Executor, modelPath: string): void;
   refreshExecutionEnv(): void;
 }
 
-class BaseEnvManager implements ExectionEnvEnvManager {
+class BaseEnvManager implements ExectionEnvManager {
   executionEnvs: Map<string, ExecutionEnv>;
   constructor() {
     this.executionEnvs = new Map<string, ExecutionEnv>();


package.json
```diff
@@ -130,8 +130,14 @@
         "icon": "$(x)"
       },
       {
-        "command": "onevscode.register-device",
-        "title": "Register",
+        "command": "onevscode.devices.register",
```
dayo09, line 133: 👍

dayo09, line 133: 👍

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// This is a base ExeuctionEnv.
+// on New ExecutionEnv implements, need to add 3things
+// 1. ExecutionEnv itself
+// 2. ExecutionEnv Creation function
+//    This will return class for Env that extends ExecutionEnvBase
+// 3. ExecutionEnv list get function
+//    This will return Commend that execute on host and return string array of device list,
+//    could split with '\n'.
```
yunjayh, line 34: Then, why `getConnectableEnvs` is not a class method, but defined in each module?

yunjayh: Plus, why does it returns `Command`, not returning the device array after executing the command by itself?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,58 @@
...
+  getExecutors(): Executor[] {
+    return Array.from(this.executors.values());
+  }
+  getExecutor(name: string): Executor|undefined {
+    return this.executors.get(name);
+  }
+  setExecutor(name: string, executor: Executor): boolean {
+    throw new Error(
+        'Check certain Executor installed and could use on this Env.\n If it could, return true and add Executor on ExecutorList.');
+  }
```
yunjayh, line 55: T^T I couldn't get what this method expected to do.. 

1. Check the validity of `executor` which has been passed by param.
2. If it's valid, add it to `this.executors` and return true
3. else return false

Did I get correctly?

yunjayh (commented):
IMHO, as we discussed before, the terms, `executor`, `executionEnv`, `device` and other things make the developer and reviewer a little bit confused. 

It may because of my lack of understanding, but I still couldn't get how the toolchain's simulator can be added as `executionEnv` yet..

src/Execute/DeviceProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+
+
+
+enum NodeType {
+  // This will be either local or remote
+  // TODO: Implement remote type.
+  hostType,
+  executionEnv,
+  executor,
+}
```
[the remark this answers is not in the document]
struss: In my thought, Three components will be a tree node.
I'm not sure this term is right for this position, but 

1. HostType : This means type of host where Device(`ExecutionEnv`) connected.
2. ExecutionEnv : This means type of ExcutionEnv where execution will be done.
3. Executor : This means execution SW defined class where execute on `ExecutionEnv`.

src/Execute/ExecutionEnv.ts
struss: > T^T I couldn't get what this method expected to do..

Sorry for confusing about this.
 
> 1. Check the validity of `executor` which has been passed by param.
> 2. If it's valid, add it to `this.executors` and return true
> 3. else return false
> 
> Did I get correctly?

yes. that's right.

src/Execute/ExecutionEnv.ts
struss: > Then, why `getConnectableEnvs` is not a class method, but defined in each module?

if we define this in class, we need to create Env before it really exists. so, I thought module could check this.

struss: > Plus, why does it returns `Command`, not returning the device array after executing the command by itself?

I thought it could run on remote host, then we should run those in `ExecutionEnvManager` object.

struss:
I think I need to comment about some terms

- ExecutionEnvManager
```
This means HostPC where Device(ExecutionEnv) connected. 
so, it will execute real command on this. so it will use either bash or ssh.
it will contains ExecutionEnv which connected on this.
```

- ExecutionEnv (=Device)
```
This means real Device which inference or real action occur.
but, we can't directly access to this device, so it just throw command to ExecutionEnvManager.
it will contains Executors which this device could use.
```

- Executor
```
Currently defined executor.
```

package.json
```diff
@@ -269,6 +280,10 @@
         "one-vscode.enableCodelens": {
           "type": "boolean",
           "default": true
+        },
+        "one-vscode.targetDeviceInfo": {
+          "type": ["object", "null"],
+          "default": null
         }
```
struss, line 287: This will be used to save Config about `Target Device View` setting.

src/Execute/Envs/Envs.ts
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+// This file will collect all Env type on this.
+export * as Host from './HostPcEnv';
+export * as Tizen from './TizenEnv';
```
struss, line 20: Little bit confused to use like this. :(

hyunsik-yoon:
Could you explain what 'Device' means in your code?

- I think some H/W device such as TV can be a `Device`. Is this correct?
- How about S/W thing such as simulator? Can a simulator be a `Device`?
- How about a PC that connects many TVs? Can PC be a `Device`?
- As far as I know, `executor` is a concept of commands such as ssh command to invoke, e.g., inference. In the code, an executor also considered as a `Device` (`DeviceNode` can have a node type of `executor`). Could you let us the reason why an executor can also be a `Device`?

struss:
> Could you explain what 'Device' means in your code?
> 
> * I think some H/W device such as TV can be a `Device`. Is this correct?

right. I thought all H/W could be device.

> * How about S/W thing such as simulator? Can a simulator be a `Device`?

I thought simulator is combination of Device(`ExecutionEnv`) and S/W(`Executor`), which HostPC and some simulator S/W

> * How about a PC that connects many TVs? Can PC be a `Device`?

Yes. but on list, it will display all TV and single PC component.

> * As far as I know, `executor` is a concept of commands such as ssh command to invoke, e.g., inference. In the code, an executor also considered as a `Device` (`DeviceNode` can have a node type of `executor`). 

um... I thought Executor just display on DeviceView to show which Device could use which Executor. but follow from your opinion, it makes people confused about this.

> Could you let us the reason why an executor can also be a `Device`?

No, Device(`ExecutionEnv`) will refer this when it does some actions. and I add that under Device(`ExecutionEnv`) to show which one could be used.

struss:
if anyone have a new idea about this, I will make this into small PR!

struss:
um....what about split `executor` from `backend`?

src/Execute/ExecutionEnvManager.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+interface ExecutionEnvManagerMap {
+  [key: string]: BaseEnvManager;
+}
+
+let globalManagerMap: ExecutionEnvManagerMap = {};
+// This list will keep envTypes on it.
+let envList = Object.values(Envs);
```
hyunsik-yoon, line 59: Some backend provider may want to implement, e.g., `RaspberryExecutionEnv` in their backend.
In such case `envList` won't contain `RaspberryExecutionEnv`.
I wonder if such situation is OK or not.

YongseopKim:
> um....what about split executor from backend?

if you feel necessary, you can do it with some reasonable reasons and discussion together.

src/Execute/ExecutionEnvManager.ts
struss: I thought anyone could create new EnvType on this, and every time they create new type, they need to add codes on `EnvManager` constructor like below

```
for device in RaspberryExecutionEnv.DeviceList():
    addDeviceList(new RaspberryExecutionEnv(device))
``` 

so, I extract all `ExecutionEnv` on `Env/Envs.ts` to export, and use from `Envs.ts` module.

struss: as talked on offline with @hyunsik-yoon , if Backend Provider want to hide there own `ExecutionEnv`, this example could not cover this. so, I think we need to add a method to add this on private backend.

struss:
https://github.com/Samsung/ONE-vscode/pull/754#discussion_r893178019

This will help to understand this.

jyoungyun:
Through the offline meeting, I could understand the code better.

The hierarchy(?) of these classes is like below (according to what I understand):

```
LocalEnvManager(ExecutionEnvManager) (e.g., Local linux machine, remote ssh linux machine)
   - HostPCEnv(ExecutionEnv)
      - Device (1 Device : 1 ExecutionEnv) (No matching class were found in this PR.)
   - TizenEnv(ExecutionEnv)
      - Device 
```

We still need to think more about the `executor`.

I drew the class diagram for this PR.
executor_uml

I'm not sure I understand exactly. If there is something wrong, please update it. :)

struss:
> Through the offline meeting, I could understand the code better.
> 
> The hierarchy(?) of these classes is like below (according to what I understand):
> 
> ```
> LocalEnvManager(ExecutionEnvManager) (e.g., Local linux machine, remote ssh linux machine)
>    - HostPCEnv(ExecutionEnv)
>       - Device (1 Device : 1 ExecutionEnv) (No matching class were found in this PR.)
>    - TizenEnv(ExecutionEnv)
>       - Device 
> ```
> 
> We still need to think more about the `executor`.
> 
> I drew the class diagram for this PR. executor_uml
> 
> I'm not sure I understand exactly. If there is something wrong, please update it. :) [executor_uml.zip](https://github.com/Samsung/ONE-vscode/files/8877445/executor_uml.zip)

exactly right! Thanks for make this UML

struss:
A little bit worried about `Executor` attached on `Backend`. if `one-infer` introduced, no need to know backend to run inference or profile using `Executor`

YongseopKim:
I and @struss discussed the design and I draft https://github.com/Samsung/ONE-vscode/pull/816.

Issue #715, opened 2022-05-26
Make "Cancel" in onecc progress notification work


hyunsik-yoon:
When we run onecc, onecc progress notification shows "cancel" button.
Let's make this button work.

![cancel-onecc](https://user-images.githubusercontent.com/63768207/170422272-73a67701-8855-461c-b810-cafc97150b48.gif)

PR #749, opened 2022-06-03
Enable 'Cancel' button of progressive notification

hyunsik-yoon:
Previously 'Cancel' button of progressive notification
while running onecc showed NYI error.

With this commit, 'Cancel' button actually terminates
onecc progress.

for For https://github.com/Samsung/ONE-vscode/issues/715
draft #716 

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -22,7 +22,6 @@ import * as vscode from 'vscode';
 
 import {CfgEditorPanel} from '../CfgEditor/CfgEditorPanel';
 import {ToolArgs} from '../Project/ToolArgs';
-import {ToolRunner} from '../Project/ToolRunner';
 import {obtainWorkspaceRoot, RealPath} from '../Utils/Helpers';
 import {Logger} from '../Utils/Logger';
 import {ConfigObj} from './ConfigObject';
@@ -506,32 +505,36 @@ export class OneExplorer {
 //
 
 import {EventEmitter} from 'events';
+import {ExitCode, SelfKilled, ToolRunner} from '../Project/ToolRunner';
 
 class OneccRunner extends EventEmitter {
+  private logTag = this.constructor.name;
+
   private startRunningOnecc: string = 'START_RUNNING_ONECC';
   private finishedRunningOnecc: string = 'FINISHED_RUNNING_ONECC';
 
+  private toolRunner: ToolRunner;
+
   constructor(private cfgUri: vscode.Uri) {
     super();
+    this.toolRunner = new ToolRunner();
   }
 
   /**
    * Function called when onevscode.run-cfg is called (when user clicks 'Run' on cfg file).
    */
   public run() {
-    const toolRunner = new ToolRunner();
-
     this.on(this.startRunningOnecc, this.onStartRunningOnecc);
     this.on(this.finishedRunningOnecc, this.onFinishedRunningOnecc);
 
     const toolArgs = new ToolArgs('-C', this.cfgUri.fsPath);
     const cwd = path.dirname(this.cfgUri.fsPath);
-    let oneccPath = toolRunner.getOneccPath();
+    let oneccPath = this.toolRunner.getOneccPath();
     if (oneccPath === undefined) {
       throw new Error('Cannot find installed onecc');
     }
 
-    const runnerPromise = toolRunner.getRunner('onecc', oneccPath, toolArgs, cwd);
+    const runnerPromise = this.toolRunner.getRunner('onecc', oneccPath, toolArgs, cwd);
     this.emit(this.startRunningOnecc, runnerPromise);
   }
 
@@ -545,14 +548,14 @@ class OneccRunner extends EventEmitter {
     // Show progress UI
     vscode.window.withProgress(progressOption, (progress, token) => {
       token.onCancellationRequested(() => {
-        vscode.window.showWarningMessage(`Error: NYI`);
+        this.toolRunner.kill();
       });
 
       const p = new Promise<void>((resolve, reject) => {
         runnerPromise
-            .then(value => {
+            .then((value: ExitCode|SelfKilled) => {
               resolve();
-              this.emit(this.finishedRunningOnecc);
+              this.emit(this.finishedRunningOnecc, value);
             })
             .catch(value => {
               vscode.window.showErrorMessage(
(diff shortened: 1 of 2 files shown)


hyunsik-yoon:
## GIF
Please go see at https://github.com/Samsung/ONE-vscode/issues/715

src/Project/ToolRunner.ts
```diff
@@ -26,6 +26,7 @@ const K_DATA: string = 'data';
 const K_EXIT: string = 'exit';
 
 // successfully killed by calling kill() method
+export type ExitCode = string;
```
hyunsik-yoon, line 29: `ExitCode` was introduced to read code in easier way.

yunjayh (approved):
LGTM

yunjayh:
😮 not requested yet..?!

src/Project/ToolRunner.ts
```diff
@@ -50,7 +51,7 @@ export class ToolRunner {
       Logger.append(data.toString());
     });
 
-    this.child!.on(K_EXIT, (code: number|null, signal: NodeJS.Signals|null) => {
+    this.child!.on(K_EXIT, (code: ExitCode|null, signal: NodeJS.Signals|null) => {
```
llFreetimell, line 54: `number` is changed to `string`, is it intended?

src/OneExplorer/OneExplorer.ts
```diff
@@ -565,7 +568,14 @@ class OneccRunner extends EventEmitter {
     });
   }
 
-  private onFinishedRunningOnecc() {
-    vscode.window.showInformationMessage(`Successfully completed`);
+  private onFinishedRunningOnecc(val: ExitCode|SelfKilled) {
+    const exitCodeSuccess = '0';
+    if (val === exitCodeSuccess) {
+      vscode.window.showInformationMessage(`Successfully completed.`);
+    } else if (val === 'SelfKilled') {
```
llFreetimell, line 575: Instead of `'0'`, `exitCodeSuccess` is used.
So how about using `SelfKilled` instead of `'SelfKilled'` for consistency? :)

```suggestion
    } else if (val === SelfKilled) {
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -506,32 +505,36 @@ export class OneExplorer {
 //
 
 import {EventEmitter} from 'events';
+import {ExitCode, SelfKilled, ToolRunner} from '../Project/ToolRunner';
 
 class OneccRunner extends EventEmitter {
+  private logTag = this.constructor.name;
```
llFreetimell, line 511: Maybe not used variable?

src/Project/ToolRunner.ts
hyunsik-yoon: Yes. For the reason in https://github.com/Samsung/ONE-vscode/pull/749#discussion_r888612720.
While looking at the code,
```
private onFinishedRunningOnecc(val: Number|SelfKilled) {
```

I thought that `Number` is not very specific and hard to figure out what kinda number should be passed.
So I decided to be more specific for clear understanding by changing `Number` to `ExitCode`.

```
private onFinishedRunningOnecc(val: ExitCode|SelfKilled) {
```

I tried to do this change in this PR since this change was related to signature change introduced with `SelfKilled`.

src/OneExplorer/OneExplorer.ts
hyunsik-yoon: With ```if (val === SelfKilled)```, I got a compilation error. 😢 
```
SelfKilled only refers to a type, but is being used as a value here.ts(2693)
```

src/OneExplorer/OneExplorer.ts
hyunsik-yoon: Removed! Thanks.

yunjayh (dismissed):

dayo09:
How about "The job is cancelled" instead of "Terminated successfully" as a cancellation message? Because the word 'successfully' may bring about a miscomprehension of "A job terminated successfully", not "A cancellation terminated successfully".

dayo09 (dismissed):
How about "The job is cancelled" instead of "Terminated successfully" as a cancellation message? Because the word 'successfully' may bring about a miscomprehension of "A job terminated successfully", not "A cancellation terminated successfully".

LGTM for this PR, I think the message can be changed in the another PR if you like.

hyunsik-yoon:
@[dayo09](https://github.com/dayo09)
Message text was changed. PTAL.

src/OneExplorer/OneExplorer.ts
YongseopKim: I understand that @llFreetimell wants like

```ts
const selfKilled = 'SelfKilled';
// ...
{ else if (val === selfKilled) {
```

dayo09 (approved):

dayo09:
Thanks a lot! :-D

src/OneExplorer/OneExplorer.ts
llFreetimell: I didn't notice that `SelfKilled` is not string but a literal type...

```ts
export type ExitCode = string;
export type SelfKilled = 'SelfKilled';
```

If so, I think `SelfKilled` is not used anywhere in both main branch and current PR... @.@

- Main branch
`SelfKilled` in the `resolve('SelfKilled')` may be handled by `string`, not `SelfKilled`.
https://github.com/Samsung/ONE-vscode/blob/746f1337a13f0c0187fb4be501107b960bca77f0/src/Project/ToolRunner.ts#L69
https://github.com/Samsung/ONE-vscode/blob/746f1337a13f0c0187fb4be501107b960bca77f0/src/Project/ToolRunner.ts#L42

- Current PR
`SelfKilled` in `val === 'SelfKilled'` maybe handled by `ExitCode`(`string`), not `Selfkilled`.
  ```ts
    private onFinishedRunningOnecc(val: ExitCode|SelfKilled) {
      const exitCodeSuccess = '0';
      if (val === exitCodeSuccess) {
        vscode.window.showInformationMessage(`Successfully completed.`);
      } else if (val === 'SelfKilled') {
  ```

Am I understanding correctly?

hyunsik-yoon: Merging `SelfKilled` and `ExitCode` into `ExitCode` also works and I tried that way at the beginning. :-)

However, the reason I wanted to decided to handle this separately is because of the following reason:
https://github.com/Samsung/ONE-vscode/blob/d4f44b8eb382657b6a57574dbb3e6ad63df9d91e/src/Project/ToolRunner.ts#L56-L62

Please refer to https://nodejs.org/api/child_process.html#event-exit

I think that we'd better keep this separation and make the semantics of `ExitCode` just like mentioned in the above reference. 😸

src/Project/ToolRunner.ts
```diff
@@ -26,6 +26,7 @@ const K_DATA: string = 'data';
 const K_EXIT: string = 'exit';
 
 // successfully killed by calling kill() method
+export type ExitCode = string;
 export type SelfKilled = 'SelfKilled';
```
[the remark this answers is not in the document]
hyunsik-yoon: Yes, correct. This type is used to indicating that this process was terminated by `kill()` method. Maybe https://github.com/Samsung/ONE-vscode/pull/749#discussion_r891829478 would help understanding. 😸

hyunsik-yoon:
@llFreetimell mentioned

> SelfKilled in the resolve('SelfKilled') may be handled by string, not SelfKilled.

After discussion, I'd like to rewrite this to 

"SelfKilled in the resolve('SelfKilled') may be handled by `ExitCode` which is `string`, not by `SelfKilled`."

This is valid point. I think that I should fix this.

hyunsik-yoon:
replaced by #810

Issue #698, opened 2022-05-23
[CfgEditor] Upgrade to whole new UI


llFreetimell:
### What?

Current CfgEditor is not that good to users.
Let's upgrade current UI to brand new UI!

### How?

For detailed **change previews**, those can be checked at draft PR or following related PRs.

- UI Upgrade
  - Change Step progress bar
  - ~~Change summary panel UI~~
  - Move 'import model type' into 'Import' step
  - Substitute import/codegen/profile step UI
- Feature Upgrade
  - Apply changes from cfg editor to cfg file
- Etc
  - Bugfix
    - `onecc` is not recognized

---

#### Draft

- #697 

#### Todo/Done

- #719 
- #721 
- #740 
- #750 
- #755 
- #764 
- #765 
- #776

hyunsik-yoon:
Yesterday, I was briefly thinking about how to make value-test panel and
I realize that look'n'feel of value-test panel must be similar to that of cfg editor.
That means that it would be great if each UI component in cfg editor could be easily reused by other panels. :-)
We could reuse at component level or source code level.. either way would be fine. :-)

llFreetimell:
Done!

Issue #746, opened 2022-06-02
Use filesystem via `vscode.worksapce.fs` (not `fs`)


dayo09:
## What?

Let's use filesystem via `vscode.worksapce.fs`, not `fs`

## Why?

They say that `fs` does not consider `ssh` support. It's safe to use `vscode.workspace.fs` instead of `fs` to support our extension on ssh on wheels.

dayo09:
@hyunsik-yoon 

vscode/index.d.ts>vscode.workspace.fs
```js

        /**
         * A {@link FileSystem file system} instance that allows to interact with local and remote
         * files, e.g. `vscode.workspace.fs.readDirectory(someUri)` allows to retrieve all entries
         * of a directory or `vscode.workspace.fs.stat(anotherUri)` returns the meta data for a
         * file.
         */
        export const fs: FileSystem;
```

https://code.visualstudio.com/api/extension-guides/virtual-workspaces#review-that-the-extension-code-is-ready-for-virtual-resources

> Look out for usages of the fs node module for file system operations. If possible, use the vscode.workspace.fs API, which delegates to the appropriate file system provider.

As you can see above, it's ms's recommend to use their api for filesystem approaches.

https://code.visualstudio.com/updates/v1_37#_vscodeworkspacefs

> Call to Action: If your extension is currently using the fs module from Node.js, consider migrating to the new vscode.workspace.fs API

YongseopKim:
I've investigated as I don't know how different it is.

## as-is (nodejs-api)

Buffer.from(): https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_class_method_buffer_from_string_encoding
writeFile(): https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_writefile_file_data_options_callback

```ts
const writeBytes = Buffer.from('Hello Node.js'); // Buffer.from(): nodejs api
const data = new Uint8Array(writeBytes);
fs.writeFile('message.txt', data, (err) => {
  if (err) throw err;
  console.log('The file has been saved!');
});
```

## to-be (vscode-api)

fs: https://code.visualstudio.com/api/references/vscode-api#FileSystem
uri: https://code.visualstudio.com/api/references/vscode-api#Uri

```ts
const writeBytes = Buffer.from('Hello vscode'); // Buffer.from(): nodejs api
// writeFile: fs of vscode-api
// someUri: Uri of vscode-api, used like `Uri.file('message.txt')`
[...]

PR #750, opened 2022-06-03
[CfgEditor] Replace static files as new

llFreetimell:
Until now, `_getHtmlForWebview` function was returning literal HTML source.
In addition, old css and js have been used.
This commit replaces static files to new ones.

---

Parent issue : #698 
Draft : #697

### DIFF
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -85,7 +85,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     };
     vscode.commands.executeCommand('setContext', CfgEditorPanel.viewType, true);
 
-    webviewPanel.webview.html = (await this._getHtmlForWebview(webviewPanel.webview)).toString();
+    webviewPanel.webview.html = await this._getHtmlForWebview(webviewPanel.webview);
 
     const changeDocumentSubscription = vscode.workspace.onDidChangeTextDocument(e => {
       if (e.document.uri.toString() === document.uri.toString()) {
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -40,7 +40,6 @@ Some part of this code refers to
 https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/custom-editor-sample/src/catScratchEditor.ts
 */
 
-import * as fs from 'fs';
 import * as ini from 'ini';
 import * as vscode from 'vscode';
 import {getNonce} from '../Utils/external/Nonce';
@@ -86,7 +85,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     };
     vscode.commands.executeCommand('setContext', CfgEditorPanel.viewType, true);
 
-    webviewPanel.webview.html = this._getHtmlForWebview(webviewPanel.webview);
+    webviewPanel.webview.html = (await this._getHtmlForWebview(webviewPanel.webview)).toString();
 
     const changeDocumentSubscription = vscode.workspace.onDidChangeTextDocument(e => {
       if (e.document.uri.toString() === document.uri.toString()) {
@@ -121,7 +120,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     this.updateWebview(document, webviewPanel);
   };
 
-  private _getHtmlForWebview(webview: vscode.Webview) {
+  private async _getHtmlForWebview(webview: vscode.Webview) {
     const nonce = getNonce();
 
     const toolkitUri = getUri(webview, this.context.extensionUri, [
@@ -143,10 +142,9 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     const cssUri =
         getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.css']);
 
-    // import html
     const htmlPath =
-        getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.html']);
-    let html = fs.readFileSync(htmlPath.fsPath, {encoding: 'utf-8'});
+        vscode.Uri.joinPath(this.context.extensionUri, 'media/CfgEditor/cfgeditor.html');
+    let html = Buffer.from(await vscode.workspace.fs.readFile(htmlPath)).toString();
 
     // Apply js and cs to html
     html = html.replace(/\${nonce}/g, `${nonce}`);
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -39,8 +39,11 @@
 Some part of this code refers to
 https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/custom-editor-sample/src/catScratchEditor.ts
 */
+
+import * as fs from 'fs';
 import * as ini from 'ini';
 import * as vscode from 'vscode';
+import {getNonce} from '../Utils/external/Nonce';
 import {getUri} from '../Utils/external/Uri';
 
 export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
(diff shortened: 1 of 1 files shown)


src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -39,8 +39,11 @@
 Some part of this code refers to
 https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/custom-editor-sample/src/catScratchEditor.ts
 */
+
+import * as fs from 'fs';
```
hyunsik-yoon, line 43: can we use `vscode.workspace.fs`?
ref: https://github.com/Samsung/ONE-vscode/issues/746

hyunsik-yoon (dismissed):

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -83,7 +85,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     };
     vscode.commands.executeCommand('setContext', CfgEditorPanel.viewType, true);
 
-    webviewPanel.webview.html = this._getHtmlForWebview(webviewPanel.webview);
+    webviewPanel.webview.html = (await this._getHtmlForWebview(webviewPanel.webview)).toString();
```
mhs4670go, line 88: I think `toString` in this line can be removed. 

As you already know, but for others information, `Buffer.from` method returns `array<uint8>` type.

But, below L147 is already calling `toString` method.
```ts
Buffer.from(await vscode.workspace.fs.readFile(htmlPath)).toString();
```

So, you've got two option.
1. Remove `toString` method in this line.
2. Make `_getHtmlForWebview` return `array<uint8>` type.

But seems first one looks better because it would be easier to modify html contents before return.

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -134,12 +137,27 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
+    html = html.replace(/\${nonce}/g, `${nonce}`);
+    html = html.replace(/\${webview.cspSource}/g, `${webview.cspSource}`);
+    html = html.replace(/\${toolkitUri}/g, `${toolkitUri}`);
+    html = html.replace(/\${codiconUri}/g, `${codiconUri}`);
+    html = html.replace(/\${jsUri}/g, `${jsUri}`);
+    html = html.replace(/\${cssUri}/g, `${cssUri}`);
+
+    return html;
+
+    // TODO Remove following unused codes
```
YongseopKim, line 159: How about commenting out the entire code?

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -134,12 +137,27 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
     ]);
-    const jsUri =
-        getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.js']);
+    const jsUri = getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'index.js']);
 
     const cssUri =
         getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.css']);
 
+    const htmlPath =
+        vscode.Uri.joinPath(this.context.extensionUri, 'media/CfgEditor/cfgeditor.html');
```
YongseopKim, line 146: oh, we use also uri in vscode?

src/CfgEditor/CfgEditorPanel.ts
YongseopKim: Aha below `return /*html*/` makes you just leave the comment. Ignore my comment.

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -118,8 +120,9 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     this.updateWebview(document, webviewPanel);
   };
 
-  // TODO Create separate HTML generator class
-  private _getHtmlForWebview(webview: vscode.Webview) {
+  private async _getHtmlForWebview(webview: vscode.Webview) {
```
mhs4670go, line 123: This function is very good to have. With this, existing static html file would be flexible and extensible. :+1: 👍

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -134,12 +137,27 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
-    const jsUri =
-        getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.js']);
+    const jsUri = getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'index.js']);
 
     const cssUri =
         getUri(webview, this.context.extensionUri, ['media', 'CfgEditor', 'cfgeditor.css']);
 
+    const htmlPath =
+        vscode.Uri.joinPath(this.context.extensionUri, 'media/CfgEditor/cfgeditor.html');
+    let html = Buffer.from(await vscode.workspace.fs.readFile(htmlPath)).toString();
```
YongseopKim, line 147: Hmm, I think this `Buffer` should be handled with exceptions like file handle and buffer(overflow/underflow). Is it okay not to care about exceptions?

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: Thanks for your suggestions :)
Actually, it's my mistake and you caught it :D

Since my original intention was the first option you said, I will fix this by removing `toString` method!

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: Can I ask you about the meaning? :sweat_smile: 

- "Oh, we also use `vscode.Uri` directly in vscode?"
- "Oh, we also use URI class in vscode?"

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: I couldn't understand your saying but... as you used overflow and underflow words, are you talking about buffer index error? :sweat_smile:

YongseopKim: Cases
1. What happen if fs cannot read the file?
2. What happen if fs can read the file and then cannot allocate memory for the file?

src/CfgEditor/CfgEditorPanel.ts
YongseopKim: sorry for ambiguous expression.. 😞 

I'm asking that
- oh, we have to use also vscode.uri class for vscode.workspace.fs?
- if so, all of `nodejs.uri` vars for `nodejs.fs` should be replaced by `vscode.uri` for `vscode.workspace.fs`?

The reason why I'm asking it is...
- if we have to change from `nodejs.fs` to `vscode.workspace.fs`,
- we also have to change components that `vscode.workspace.fs` needs

llFreetimell: > we have to use also vscode.uri class for vscode.workspace.fs?

Yes, parameter of `vscode.workspace.fs` is `vscode.Uri`

> all of nodejs.uri vars for nodejs.fs should be replaced by vscode.uri for vscode.workspace.fs?

AFAIK, yes...!

> we also have to change components that vscode.workspace.fs needs

IIUC, this may be handled in #746

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: > What happen if fs cannot read the file?

If `await vscode.workspace.fs.readFile(htmlPath)` returns `null` or `undefined`, `Buffer.from` raises Exception.
But I didn't consider this because we load static file..!

> What happen if fs can read the file and then cannot allocate memory for the file?

I didn't consider this because HTML file is really small...!
And after loading is finished, variable is not used anymore :)

YongseopKim (approved):
LGTM

hyunsik-yoon (approved):

src/CfgEditor/CfgEditorPanel.ts
mhs4670go: @llFreetimell Sounds very reasonable.

(optional) But,
1. I think exception is an exception because we couldn't expect and the way the html contents are loaded could be changed. It would be better to handle exception on this case.
2. There is `Utils/Uri.ts` in our project where helper functions for URI exist. How about make this line a single function and call it for maintainability and readability?

mhs4670go (approved):
LGTM

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: > we couldn't expect and the way the html contents are loaded could be changed. It would be better to handle exception on this case.

I agree :)

> How about make this line a single function and call it for maintainability and readability?

I think it is good idea to create separate function with exception handling! That would be definitely good for maintainability and readibility :D
And also I think it would be better for newbies to use `vscode.workspace.fs.readFile` :)
Let's do it in another PR!

Issue #751, opened 2022-06-03
[CfgEditor] Add Optimize Step


llFreetimell:
### What?

Current CfgEditor does not support Optimize step.
Let's implement it!

### How?

- Design & Implement Optimize Step
- Introduce json to gather most of optimization paramters
  - Exclude following options. These are future feature lists.
    - mute_warnings
    - disable_validation
    - generate_profile_data
    - change_outputs
    - sparsify_tensor
    - sparsify_traversal_order
    - sparsify_format
    - sparsify_block_size
    - sparsify_block_map
- Implement event listeners for optimization elements

---

#### Draft

- #697 

#### Todo/Done

- #752

PR #752, opened 2022-06-03
[CfgEditor] Introduce Optimize step

llFreetimell:
This commit adds HTML and static files for Optimize step.

---

Parent Issue : #751 
Draft : #697 

---

#### Before & After

| UI | Optimize Step | :x: NYI | optimize |

### DIFF
--- a/media/CfgEditor/cfgeditor.html
+++ b/media/CfgEditor/cfgeditor.html
@@ -212,7 +212,38 @@ limitations under the License.
             </div>
         </div>
         <div class="options" id="optionOptimize">
-            <!-- TODO Implement Optimize step -->
+            <div class="basic">
+                <div class="title">
+                    Basic Options
+                </div>
+                <div class="option">
+                    <vscode-text-field id="optimizeInputPath" startIcon="true" size="50" placeholder="">
+                        Input Path
+                        <span slot="end" class="codicon codicon-search" style="cursor: pointer"></span>
+                    </vscode-text-field>
+                </div>
+                <div class="option">
+                    <vscode-text-field id="optimizeOutputPath" startIcon="true" size="50" placeholder="">
+                        Output Path
+                        <span slot="end" class="codicon codicon-search" style="cursor: pointer"></span>
+                    </vscode-text-field>
+                </div>
+            </div>
+            <div class="advanced">
+                <div class="title">
+                    Optimization Parameters
+                    <span slot="end" class="codicon codicon-question" style="cursor: pointer"></span>
+                </div>
+                <div class="option">
+                    <vscode-data-grid id="basicOptimizeTable" generate-header="default" aria-label="Default Header" grid-template-columns="5% 40% 55%">
+                        <vscode-data-grid-row row-type="header">
+                            <vscode-data-grid-cell cell-type="columnheader" grid-column="1">&nbsp;</vscode-data-grid-cell>
+                            <vscode-data-grid-cell cell-type="columnheader" grid-column="2">Name</vscode-data-grid-cell>
+                            <vscode-data-grid-cell cell-type="columnheader" grid-column="3">Description</vscode-data-grid-cell>
+                        </vscode-data-grid-row>                        
+                    </vscode-data-grid>
+                </div>                
+            </div>
         </div>
         <div class="options" id="optionQuantize">
             <!-- TODO Implement Quantize step -->
--- a/media/CfgEditor/displaycfg.js
+++ b/media/CfgEditor/displaycfg.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import oneOptimizationList from './one-optimizations.json' assert {type : 'json'};
 import {updateImportUI} from './updateUI.js';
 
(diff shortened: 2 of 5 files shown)


hyunsik-yoon:
IMHO 'ON' (or 'OFF') text looks same with other _static_ texts, e.g., optimization name or description.
To me, it's kinda difficult to know weather 'ON' text is static or it can be changed to 'OFF' by clicking it. Maybe we'd better have some indicator(? e.g., icon? check box? blue underline?) to let users that they can click `ON`.

dayo09:
Q. what happens when the user edit the property on the text editor and save it?

llFreetimell:
@hyunsik-yoon 

> How about having some kinda indicator(? e.g., icon? check box? blue underline?) to let users that they can click ON?

For now, cursor style is `pointer` for users to know that ON/OFF is click-able.
But I agree that something additional UI would be better :(

llFreetimell:
@dayo09 

"Editing the property on the editor" is equal to "editing the values on the text editor".
It can be checked at the gif files :D
- editing property on the editor --> the change is applied to file content but not saved
- save --> the changed file content is saved

hyunsik-yoon:
> But I agree that something additional UI would be better :(

It's up to you to proceed this work in this PR or in future PR. :-) Just let us know. 😸

dayo09:
@llFreetimell Oh, I couldn't see that as it was so small, now I opened it on another tab and can see them. thanks.

dayo09 (dismissed):

dayo09:
Q. 
As I can see in the gif, the behavior is:

1. Switching 'ON' adds the optimization property and sets it as True.
2. Switching to 'OFF' remove the optimization property.
3. When setting the optimization property as False, it switches to 'OFF'

It seems that the original .cfg file defines these three states :
 a) property defined ans set as True
 b) property defined and set as False
 c) property not defined

And Cfg editor defines two states:
 A) Set as ON
 B) Set as OFF

So are they equievalant as below?
``` 
A<-> a

B <-> b <-> c
```

llFreetimell:
@dayo09 

> it takes me a massive concentration on the cfg editor to notice the difference

Sorry for attacking your eyesight... :cry: 

> how about highlighting the option when it's changed?

I will find a better way to highlight the changes :)

> I'd suggest to make them strictly equivalent as making True/False/None instead of ON/OFF matcher.

At first, your understanding is correct. b and c are exactly same and they are equal with B.

```
b) property defined and set as False
c) property not defined
B) Set as OFF

Answer : (b == c) <--> B
```

Unfortunately, I am against for following suggestions T.T

1. Let's assume that we introduce NONE.

    I agree that current implementation can cause some confusion too :(
    However, following two expression is exactly same for ONE.

    ```
    [one-optimize]
    input_path=input
    output_path=output
    fold_add_v2=False

    [one-optimize]
    input_path=input
    output_path=output
    ```

    In this point of view, we cannot answer for "What is different for both NONE and FALSE?".
    It means that NONE can cause even worse confusion for users, even above two have no differences.

2. Let's assume that we remove `c) property not defined`, which means that all properties should be defined.

    This may reduce confusion for users.
    However, I worry about two things

    - The number of properties
        - Currently, the number of all the properties is larger than 70~80.
        - It means that defining all the properties may create really long file.
    - Sample cfg files in ONE
        - Most of sample cfg files in ONE are not defining all of the properties.
        - It means that ONE is allowing `c) property not defined`.

Of course, it is only my sole opinion.
Therefore, if most of members agree to apply one of above suggestions, I will apply it :)
Or if you have better suggestion, please let me know :D

dayo09:
@llFreetimell Thanks for the explanation.
Considering your comments, I suggest this: 

1. Use 'True / False' instead of 'On / Off'. Because the less terms will make the less confusions.
2. Put ❓ at the 'switch' button, just like 'optimization parameters' have. And give an explanation that "${option}=False is equivalent to an empty option"

jyoungyun (approved):
LGTM

My concern is that if optimize option is changed, it will need to be modified in several places, including this file. However, for now, the current way seems to be the best solution. Let's think about how to manage it efficiently in the future. :)

llFreetimell:
@dayo09 

> Use 'True / False' instead of 'On / Off'. Because the same words will make less confusion.

I removed the word and used checkbox :D

> Put question at the 'switch' button, just like 'optimization parameters' have. And give an explanation that "${option}=False is equivalent to an empty option"

Because of technical issue, codicon is not inserted inside of vscode-data-grid :(
I will find alternative solution later T.T

llFreetimell:
GIF is updated, PTAL :D

dayo09:
Looks better, and I think it resolves @hyunsik-yoon 's worry too.

> IMHO 'ON' (or 'OFF') text looks same with other static texts, e.g., optimization name or description.
> To me, it's kinda difficult to know weather 'ON' text is static or it can be changed to 'OFF' by clicking it. How about having some kinda indicator(? e.g., icon? check box? blue underline?) to let users that they can click ON?
> 

Thanks.

dayo09 (approved):

jyoungyun (approved):
LGTM

PR #754, opened 2022-06-07
[Device] Add ExecutionEnv

struss:
This commit will add `ExecutionEnv`.

#

From #744

after adding this, I will add each envType.

### DIFF
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -25,7 +25,7 @@ interface ExecutionEnv {
 // This is a base ExeuctionEnv.
 // on New ExecutionEnv implements, need to add 3 things
 // 1. ExecutionEnv class itself
-// 2. ExecutionEnv Creation function : named `function getEnv(name: string): ExecutionEnv`
+// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
 //    This will return class for Env that extends ExecutionEnvBase
 // 3. ExecutionEnv list get function : named `function getConnectableEnvs(): Command`
 //    This will return Commend that execute on host and return string array of device list,
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -16,12 +16,18 @@
 
 import {Command} from '../Backend/Command';
 
+// ExecutionEnv is a class for each Device.
+// It contains information about how to run action on this Device.
+// name() will return a unique identifier for Device.
+// wrapCommand() update command to run on certain Device.
+// TODO: Rename this with right meaning.
 interface ExecutionEnv {
   name(): string;
+  // TODO: Rename this.
   wrapCommand(command: Command): Command;
 }
 
-// This is a base ExeuctionEnv.
+// This is a base class for ExeuctionEnv.
 // To write a class that inherits ExecutionEnvBase, the following needs to be added
 // 1. ExecutionEnv class itself
 // 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string):
@@ -33,7 +39,6 @@ class ExecutionEnvBase implements ExecutionEnv {
   name(): string {
     throw new Error('Method not implemented.');
   }
-  // This function will wrap command with bridge command.
   wrapCommand(command: Command): Command {
     throw new Error('Method not implemented.');
   }
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -25,10 +25,11 @@ interface ExecutionEnv {
 // This is a base ExeuctionEnv.
 // To write a class that inherits ExecutionEnvBase, the following needs to be added
 // 1. ExecutionEnv class itself
-// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
+// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string):
+// ExecutionEnv`
 //    This will return class for Env that extends ExecutionEnvBase
-// 3. ExecutionEnv list get function : named `function getConnectableEnvs(): Command`
-//    This will return Commend that execute on host and return string array of device list,
+// 3. ExecutionEnv list get function : named `function getDeviceListCommand(): Command`
+//    This will return Command that is executed on host and returns string array of device list,
 //    could split with '\n'.
 class ExecutionEnvBase implements ExecutionEnv {
   name(): string {
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -23,7 +23,7 @@ interface ExecutionEnv {
 }
 
 // This is a base ExeuctionEnv.
-// on New ExecutionEnv implements, need to add 3 things
+// To write a class that inherits ExecutionEnvBase, the following needs to be added
 // 1. ExecutionEnv class itself
 // 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
 //    This will return class for Env that extends ExecutionEnvBase
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -18,7 +18,6 @@ import {Command} from '../Backend/Command';
 
 interface ExecutionEnv {
   name(): string;
-  envInfo(): string;
   wrapCommand(command: Command): Command;
 }
 
@@ -32,9 +31,6 @@ interface ExecutionEnv {
 //    This will return Command that is executed on host and returns DeviceList string.
 class ExecutionEnvBase implements ExecutionEnv {
   name(): string {
-    throw new Error('Method not implemented.');
-  }
-  envInfo(): string {
     throw new Error('Method not implemented.');
   }
   // This function will wrap command with bridge command.
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -29,8 +29,7 @@ interface ExecutionEnv {
 // ExecutionEnv`
 //    This will return class for Env that extends ExecutionEnvBase
 // 3. ExecutionEnv list get function : named `function getDeviceListCommand(): Command`
-//    This will return Command that is executed on host and returns string array of device list,
-//    could split with '\n'.
+//    This will return Command that is executed on host and returns DeviceList string.
 class ExecutionEnvBase implements ExecutionEnv {
   name(): string {
     throw new Error('Method not implemented.');
--- a/src/Execute/ExecutionEnv.ts
+++ b/src/Execute/ExecutionEnv.ts
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Command} from '../Backend/Command';
+
+interface ExecutionEnv {
+  name(): string;
+  envInfo(): string;
+  runCommand(command: Command): Command;
+}
+
+// This is a base ExeuctionEnv.
+// on New ExecutionEnv implements, need to add 3 things
+// 1. ExecutionEnv class itself
+// 2. ExecutionEnv Creation function : named `function getEnv(name: string): ExecutionEnv`
+//    This will return class for Env that extends ExecutionEnvBase
+// 3. ExecutionEnv list get function : named `function getConnectableEnvs(): Command`
+//    This will return Commend that execute on host and return string array of device list,
+//    could split with '\n'.
+class ExecutionEnvBase implements ExecutionEnv {
+  name(): string {
+    throw new Error('Method not implemented.');
+  }
+  envInfo(): string {
+    throw new Error('Method not implemented.');
+  }
+  runCommand(command: Command): Command {
+    throw new Error('Method not implemented.');
+  }
+}
+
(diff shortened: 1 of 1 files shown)


src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,45 @@
...
+import {Command} from '../Backend/Command';
+
+interface ExecutionEnv {
+  name(): string;
+  envInfo(): string;
+  runCommand(command: Command): Command;
+}
+
+// This is a base ExeuctionEnv.
+// on New ExecutionEnv implements, need to add 3 things
```
hyunsik-yoon, line 26:
```suggestion
// on new ExecutionEnv implements, need to add 3 things
```

hyunsik-yoon: Does _new ExecutionEnv_ mean a class that inherits `ExecutionEnvBase`?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Command} from '../Backend/Command';
+
+interface ExecutionEnv {
+  name(): string;
+  envInfo(): string;
```
hyunsik-yoon, line 21: Could you add comment about what `envInfo()` should return? Comment with some example would be great.

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Command} from '../Backend/Command';
+
+interface ExecutionEnv {
```
hyunsik-yoon, line 19: Can I double check on `ExecutionEnv`?

Assuming the following dev environment,
- Linux PC, _A_, running ONE-vscode, and has connection with one TV named _TV1_.
- another PC, _B_, that has connection with two more TVs named _TV2_ and _TV3_.

As far as I understand, 4 `ExecutionEnv` (for_TV1_,  _TV2_, _TV3_, and _B_) will be instantiated to use all TVs. Could you let me know if this is correct?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,45 @@
...
+  name(): string;
+  envInfo(): string;
+  runCommand(command: Command): Command;
```
hyunsik-yoon, line 22: Is the role of this function running `command` param?
Then, should return type be some result (e.g., success, fail) after running command ?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// This is a base ExeuctionEnv.
+// on New ExecutionEnv implements, need to add 3 things
+// 1. ExecutionEnv class itself
+// 2. ExecutionEnv Creation function : named `function getEnv(name: string): ExecutionEnv`
```
hyunsik-yoon, line 28: 1.  `create` would be better than `get` since it creates execution env
2. In our code, there are another `Env`. More specific term would be better, I think, .e.g., `Env` -> `ExecutionEnv`

https://github.com/Samsung/ONE-vscode/blob/d4f44b8eb382657b6a57574dbb3e6ad63df9d91e/src/Toolchain/ToolchainEnv.ts#L181

```suggestion
// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
```

hyunsik-yoon, line 28: Can I ask a question?

Assume that we have, e.g, `RaspberryExecutionEnv` that inherits `ExecutionEnvBase`.
One way to create an instance is 

```
let r = new RaspberryExecutionEnv('raspberry 1');
```
and you're recommending

```
let r = getEnv('raspberry 1');
```

Could you explain why you designed like the latter?

src/Execute/ExecutionEnv.ts
struss: Oh, I made a mistake. I thought Env just make a command to run, not running command. I will fix it.

I thought `Executor` make `Command` for inference or profile, and `ExecutionEnv` wrap this with bridge command.
I will change a name of function. sorry for confusing.

src/Execute/ExecutionEnv.ts
struss: I thought that could be 5.

in that case it will show DeviceView below

```
Local (PC A, EnvManager)
L HostPC (PC A, ExecutionEnv)
L TV1(ExecutrionEnv)
Remote (PC B, EnvManager)
L HostPC (PC B, ExecutionEnv)
L TV2(ExecutrionEnv)
L TV3(ExecutrionEnv)
```

In case `PC B`, it will handle as a remote connection

hyunsik-yoon: Oh. I see. Can I have one more question?
what if PC A or B does _not_ have a toolchain installed? Is PC A or B still a ExecutionEnv?

src/Execute/ExecutionEnv.ts
hyunsik-yoon: Aha. I understand now.
Then how about something like `wrapCommand()`?

Maybe some comment would help people.

src/Execute/ExecutionEnv.ts
struss: Oh I will fix it.

src/Execute/ExecutionEnv.ts
struss: 👍 I will fix it. thanks.

src/Execute/ExecutionEnv.ts
struss: yes. just those are ExecutionEnv.

src/Execute/ExecutionEnv.ts
struss: thank! I will fix it!

src/Execute/ExecutionEnv.ts
yunjayh: > yes. just those are ExecutionEnv.

As the user timeline using ONE-vscode
  * (1) There was nothing in `target devices` view
  * (1)-1 Or, as default, `Local PC` is registered without any `ExecutionEnv` below.
  * (2) User can add `ExecutionEnv` by his/her manual registration step. 
  * (3) Let's assume user is going to register TV#2 which is wire-connected with PC2
  * (4) User selects remote PC as `EnvManager` and register the IP of PC2
  * (5) User selects `ExecutionEnv` type as TV and select TV2
  * (6) Then, new `EnvManager` PC2 will be appear in `Target Devices` view with `ExecutionEnv` TV2 as its child node.

I'm not sure how PC A and PC B still a ExecutionEnv..

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,46 @@
...
+  wrapCommand(command: Command): Command;
+}
+
+// This is a base ExeuctionEnv.
+// To write a class that inherits ExecutionEnvBase, the following needs to be added
+// 1. ExecutionEnv class itself
+// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
+//    This will return class for Env that extends ExecutionEnvBase
+// 3. ExecutionEnv list get function : named `function getConnectableEnvs(): Command`
+//    This will return Commend that execute on host and return string array of device list,
```
hyunsik-yoon, line 31:
```suggestion
//    This will return Command that is executed on host and return string array of device list,
```

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,46 @@
...
+  envInfo(): string;
+  wrapCommand(command: Command): Command;
+}
+
+// This is a base ExeuctionEnv.
+// To write a class that inherits ExecutionEnvBase, the following needs to be added
+// 1. ExecutionEnv class itself
+// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string): ExecutionEnv`
+//    This will return class for Env that extends ExecutionEnvBase
+// 3. ExecutionEnv list get function : named `function getConnectableEnvs(): Command`
```
hyunsik-yoon, line 30: IMHO this name is confusing since `getConnectableEnvs()` sounds like it would return list of envs.
How about something like `getDeviceListCommand() : Command` ?

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,46 @@
...
+//    This will return Commend that execute on host and return string array of device list,
+//    could split with '\n'.
```
hyunsik-yoon, line 32: Could you explain a bit about this line? 
Line 31 mentioned string array will be returned. You mean each element in string array could contain '\n' and could splits again?

src/Execute/ExecutionEnv.ts
struss: um... A little bit differ.

if `ExecutionEnv` inherited class added on, I thought those could be added automatically. 

for example, if user add a remote `ExecutionEnvManager`, then it automatically make `ExecutionEnv` search and add it.

src/Execute/ExecutionEnv.ts
struss: Oh, sorry. I write invalid information. I mean `single line, split by '\n'. each token will be device name.`. I will fix this comment.

src/Execute/ExecutionEnv.ts
yunjayh: Aha, now I understand how `ExecutionEnv` are automatically registered. Thank you for comment.

AFAIU, as you said `it automatically make ExecutionEnv search and add it`, `EnvManager` will search for its child `ExecutionEnv` with the result of command from `wrapCommand()`.
If there is no toolchain installed in PC B, then IMHO the command should not return anything and PC B cannot registered as `ExecutionEnv`. 

(I think this discussion is a little bit out of scope for this PR. API itself doesn't matter for this conversation and it looks good to me.)

src/Execute/ExecutionEnv.ts
struss: Um.... currently not sure, but I think this function will be used to get information about this and check certain executor run on this ExecutionEnv. I think I will add this on other PR that handle about `Executor` to check about this.

hyunsik-yoon: How about introducing `envInfo()` when you really need?
ref: https://ko.wikipedia.org/wiki/YAGNI

src/Execute/ExecutionEnv.ts
hyunsik-yoon: So AFAIU, a device can have `ExecutionEnv` if
- the device has toolchain installed (having simulator)
- or the device can run commands, e.g., `infer`
- or the device has connection with other devices that have `ExecutionEnv`

How about adding such comment for `ExecutionEnv`?

Since I and @yunjayh does not understand `ExecutionEnv` clearly and I am afraid that other people would not understand clearly either.

src/Execute/ExecutionEnv.ts
struss: if we use like former one, then, every time some provider add new `ExecutionEnv`, they need to add some code that check and add this.(https://github.com/Samsung/ONE-vscode/pull/744#discussion_r891852735). so, I make a common factor as a module export function and use this.

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// This is a base ExeuctionEnv.
+// To write a class that inherits ExecutionEnvBase, the following needs to be added
+// 1. ExecutionEnv class itself
+// 2. ExecutionEnv Creation function : named `function createExecutionEnv(name: string):
+// ExecutionEnv`
+//    This will return class for Env that extends ExecutionEnvBase
+// 3. ExecutionEnv list get function : named `function getDeviceListCommand(): Command`
+//    This will return Command that is executed on host and returns string with '\n'.
+//    Each token splited by '\n' is device name.
```
hyunsik-yoon, line 33: IMHO information related to `\n` is too deep to describe here. I think you need to write this again when you introduce code for `getDeviceListCommand()`....
How about removing information related to `\n` here?

src/Execute/ExecutionEnv.ts
yunjayh: > It's also hard to review such function because reviewers cannot know

I absolutely agree with this. I prefer to introduce a dummy backend and ExecutionEnv to your draft. It will help both reviewers and code writer because explaining with example make things much easier. :smile:

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Command} from '../Backend/Command';
+
+interface ExecutionEnv {
```
YongseopKim, line 19: AFAIU, ExecutionEnv
- a target device: a local pc(simulator), a tv, ...
- This does not act(like install, run, uninstall, ...). This has info(Command) for the env. This is a kind of getter for Command of the specific env.
- This is a kind of data/info class for the specific env.
- This is a kind of CommandBuilder for the specific env.
- ExecutionEnv can

src/Execute/ExecutionEnv.ts
```diff
@@ -0,0 +1,42 @@
...
+  name(): string;
+  wrapCommand(command: Command): Command;
```
YongseopKim, line 21: As I told @struss offline, I disagree with this name. `ExecutionEnv` can be not related to bridge envs.

YongseopKim (commented):
The naming problem takes too much long time. If there is a problem with naming, let's write the `TODO: Rename this` comment first.

src/Execute/ExecutionEnv.ts
struss: I make this TODO on this. I will make some nice name for this.

struss:
Thanks for @jyoungyun , UML for this completed.
https://github.com/Samsung/ONE-vscode/pull/744#issuecomment-1152136506

hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

PR #755, opened 2022-06-08
[CfgEditor] Handling messages from webview

llFreetimell:
This commit adds message handlers to process messages from webview.

---

Parent Issue : #698 
Draft: #697

### DIFF
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -147,37 +147,20 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
             }
           }
 
-          // TODO Optimize this to modify only changed lines
           if (isSame === false) {
+            // cfg file is written along with the order of array elements
             let sortedCfg: any = {};
-            if (this._oneConfig['onecc'] !== undefined) {
-              sortedCfg['onecc'] = this._oneConfig['onecc'];
-            }
-            if (this._oneConfig['one-import-tf'] !== undefined) {
-              sortedCfg['one-import-tf'] = this._oneConfig['one-import-tf'];
-            }
-            if (this._oneConfig['one-import-tflite'] !== undefined) {
-              sortedCfg['one-import-tflite'] = this._oneConfig['one-import-tflite'];
-            }
-            if (this._oneConfig['one-import-bcq'] !== undefined) {
-              sortedCfg['one-import-bcq'] = this._oneConfig['one-import-bcq'];
-            }
-            if (this._oneConfig['one-import-onnx'] !== undefined) {
-              sortedCfg['one-import-onnx'] = this._oneConfig['one-import-onnx'];
-            }
-            if (this._oneConfig['one-optimize'] !== undefined) {
-              sortedCfg['one-optimize'] = this._oneConfig['one-optimize'];
-            }
-            if (this._oneConfig['one-quantize'] !== undefined) {
-              sortedCfg['one-quantize'] = this._oneConfig['one-quantize'];
-            }
-            if (this._oneConfig['one-codegen'] !== undefined) {
-              sortedCfg['one-codegen'] = this._oneConfig['one-codegen'];
-            }
-            if (this._oneConfig['one-profile'] !== undefined) {
-              sortedCfg['one-profile'] = this._oneConfig['one-profile'];
-            }
+            const sections = [
+              'onecc', 'one-import-tf', 'one-import-tflite', 'one-import-bcq', 'one-import-onnx',
+              'one-optimize', 'one-quantize', 'one-codegen', 'one-profile'
+            ];
+            sections.forEach((section) => {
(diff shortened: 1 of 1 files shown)


src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -111,7 +111,80 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
       switch (e.type) {
-        // TODO Add more webview modification handlers
+        case 'setParam':
+          if (this._oneConfig[e.section] === undefined) {
+            this._oneConfig[e.section] = {};
+          }
+          if (this._oneConfig[e.section][e.param] === undefined) {
+            this._oneConfig[e.section][e.param] = '';
+          }
+          this._oneConfig[e.section][e.param] = e.value;
```
YongseopKim, line 121: `e.value`'s type is always `string`?

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -111,7 +111,80 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
+          break;
+        case 'setSection':
+          this._oneConfig[e.section] = ini.parse(e.param);
```
YongseopKim, line 124: if `ini.parse()` is failed?

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -111,7 +111,80 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
+          break;
+        case 'updateDocument':
+          let isSame = true;
+          const iniDocument = ini.parse(document.getText());
```
YongseopKim, line 128: Hmm... below code can be extracted to a function. 

(Optional) However, any better way? like using `diff`?

YongseopKim: For it, of course, rules like whether using `whitespace` or not should be aligned.

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -111,7 +111,80 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
+              }
+              isSame = false;
+              break;
+            }
+          }
+
+          // TODO Optimize this to modify only changed lines
+          if (isSame === false) {
+            let sortedCfg: any = {};
+            if (this._oneConfig['onecc'] !== undefined) {
```
YongseopKim, line 153: I think that these can be done by

```ts
arr = ['onecc', ...]
// use for each
```

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: Yes, `e.section`, `e.param`, `e.value` are always string..!

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: As any documentation is found, I looked into the implementation of `ini.parse`.
But I cannot find any error/exception handling in the function.
So I think `ini.parse` returns wrongly parsed results when `e.param` is invalid..!

src/CfgEditor/CfgEditorPanel.ts
YongseopKim: I got why my idea cannot. the sections and its keys can have own orders. /cc @llFreetimell

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -147,37 +147,20 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
-              sortedCfg['one-quantize'] = this._oneConfig['one-quantize'];
-            }
-            if (this._oneConfig['one-codegen'] !== undefined) {
-              sortedCfg['one-codegen'] = this._oneConfig['one-codegen'];
-            }
-            if (this._oneConfig['one-profile'] !== undefined) {
-              sortedCfg['one-profile'] = this._oneConfig['one-profile'];
-            }
+            const sections = [
+              'onecc', 'one-import-tf', 'one-import-tflite', 'one-import-bcq', 'one-import-onnx',
```
YongseopKim, line 154: (In the future) we can place these const strings in one file.

YongseopKim (approved):
LGTM

dayo09 (approved):

src/CfgEditor/CfgEditorPanel.ts
YongseopKim: for example,

supported_sections.txt
```
onecc
one-import-tf
...
```
or

supported_sections.ts
```
const sections = [ 'onecc', ... ];
```

llFreetimell: I agree :)
It would be better to create some meta files for ONE compiler like following in the future :D
```
media/ONE/optimize_options.ts
media/ONE/sections.ts
...
```

Issue #728, opened 2022-05-27
Format command ids


dayo09:
## What?

Let's format our command names as 
`one.\<module\>.\<command\>`

1. `one.` prefix
- Use `one` instead of `onevscode` as our extension's official name is 'one'
2. `\<module\>.\<command\>`
- Currently, you can see that vscode defines their own commands as this.
  - workbench.view.earch (View: Show Search)
  - workbench.extensions.action.installExtensions (Extensions: Install Extensions)

#### Current command lists
```
onevscode.refresh-one-explorer ---> one.exploer.refresh
onevscode.create-cfg 
onevscode.flip-cfg-to-gui ----------> one.cfg-editor.open-gui
onevscode.flip-cfg-to-ini
onevscode.infer-model
onevscode.run-cfg
onevscode.rename-on-oneexplorer
onevscode.refresh-toolchain -------> one.toolchain.refresh
onevscode.install-toolchain
onevscode.uninstall-toolchain
onevscode.register-device ----------> one.devices.register 
onevscode.build
onevscode.import
onevscode.json-tracer
onevscode.configuration-settings
onevscode.toggle-codelens
onevscode.circle-tracer
onevscode.circle-graphview
onevscode.refresh-one-explorer
onevscode.refresh-toolchain
onevscode.install-toolchain
onevscode.register-device
onevscode.run-cfg
onevscode.rename-on-oneexplorer
onevscode.open-text-editor-on-one-explorer
onevscode.create-cfg
onevscode.uninstall-toolchain
onevscode.flip-cfg-to-gui
onevscode.flip-cfg-to-ini
```

## Why?
1. For our convinience 
-> There are basic operations like "add / move / delete" which one explorer / toolchain view/ device view will implement. Let's separate their namespace for convinience.

2. To have sync with the naming rule of other vscode commands.

dayo09:
### Namespace decisions

It's clear about explorer/toolchain/device view's commands,
but it's hard to decide pre-existing function's workspace names.. Please suggest any ideas on this table.

Before | After
-- | --
onevscode.refresh-one-explorer | one.explorer.refresh
onevscode.create-cfg | one.explorer.createCfg
onevscode.run-cfg | one.explorer.run
onevscode.rename-on-oneexplorer | one.explorer.rename
onevscode.refresh-one-explorer | one.explorer.refresh
onevscode.open-text-editor-on-one-explorer | one.explorer.openTextEditor
  |  
[...]

PR #759, opened 2022-06-08
[CfgEditor] Rename command names

dayo09:
This commit renames command names from
onevscode.* to one.editor.* and so on.

Related to #728

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,13 +93,13 @@
         "icon": "$(add)"
       },
       {
-        "command": "onevscode.flip-cfg-to-gui",
+        "command": "one.editor.open",
         "title": "Open cfg in GUI",
         "category": "ONE",
         "icon": "$(go-to-file)"
       },
       {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -93,13 +93,13 @@
         "icon": "$(add)"
       },
       {
-        "command": "onevscode.flip-cfg-to-gui",
+        "command": "one.editor.open",
```
llFreetimell, line 96: If we introduce new type of custom editor, we cannot distinguish from it.
So I think adding `cfg` word seems better :)

```suggestion
        "command": "one.editor.openCfg",
```

dayo09: Do we plan to? I thought we plan to add more backend but not a cfg editor..

dayo09: What about renaming it later then? :-D

jyoungyun (approved):
LGTM

llFreetimell (approved):
LGTM

PR #765, opened 2022-06-09
[CfgEditor] Add more condition for onDidChangeTextDocument

llFreetimell:
Because of VSCode design, `onDidChangeTextDocument` can be called even when there are no changes.
It is not intended behavior that `updateWebview` is called even if document is not changed.
This commit will fix it by checking whether there are actual document changes.

---

Parent Issue : #698 
Draft : #697

### DIFF
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -88,7 +88,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     webviewPanel.webview.html = await this._getHtmlForWebview(webviewPanel.webview);
 
     const changeDocumentSubscription = vscode.workspace.onDidChangeTextDocument(e => {
-      if (e.document.uri.toString() === document.uri.toString()) {
+      if (e.contentChanges.length > 0 && e.document.uri.toString() === document.uri.toString()) {
         this.updateWebview(document, webviewPanel);
       }
     });


src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -88,7 +88,7 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
     webviewPanel.webview.html = await this._getHtmlForWebview(webviewPanel.webview);
 
     const changeDocumentSubscription = vscode.workspace.onDidChangeTextDocument(e => {
-      if (e.document.uri.toString() === document.uri.toString()) {
+      if (e.contentChanges.length > 0 && e.document.uri.toString() === document.uri.toString()) {
```
YongseopKim, line 91: Hmm... if `onDidChangeTextDocument can be called even when there are no changes.`, can we be sure `e` is not null? (99% it is not null)

YongseopKim (approved):
LGTM

src/CfgEditor/CfgEditorPanel.ts
YongseopKim: Q. `e.document.uri.toString() === document.uri.toString()` can be replaced by `e.document.uri === document.uri`?

dayo09: @YongseopKim I tested uri comparison before, and URIs with the same paths were not the same by === operator....

llFreetimell: > can we be sure e is not null? (99% it is not null)

I don't know the exact implementation of VSCode internal logic, but I am sure because VSCode sample extension is written exactly same :D

> Q. e.document.uri.toString() === document.uri.toString() can be replaced by e.document.uri === document.uri?

I think this can be dangerous because `object === object` compares their memory address, not member variables..!

dayo09 (approved):

YongseopKim:
FYI, I asked ` e.document.uri === document.uri`. My question was not about `object === object`(object's reference or hashvalue === object's reference or hashvalue). In C++, we can see `operator==` so that I guessed it. I considered fsPath of Uri but it doesn't have scheme. I've read this and then I realize it's safer way than anything.

```ts
        /**
         * Returns a string representation of this Uri. The representation and normalization
         * of a URI depends on the scheme.
         *
         * * The resulting string can be safely used with {@link Uri.parse}.
         * * The resulting string shall *not* be used for display purposes.
         *
         * *Note* that the implementation will encode _aggressive_ which often leads to unexpected,
         * but not incorrect, results. For instance, colons are encoded to `%3A` which might be unexpected
         * in file-uri. Also `&` and `=` will be encoded which might be unexpected for http-uris. For stability
         * reasons this cannot be changed anymore. If you suffer from too aggressive encoding you should use
         * the `skipEncoding`-argument: `uri.toString(true)`.
         *
         * @param skipEncoding Do not percentage-encode the result, defaults to `false`. Note that
         *    the `#` and `?` characters occurring in the path will always be encoded.
         * @returns A string representation of this Uri.
         */
        toString(skipEncoding?: boolean): string;
```

In short, `e.document.uri.toString() === document.uri.toString()` is the right way 😄 .

Issue #641, opened 2022-05-16
[PartitionEditor] GUI Editor for .part file 


seanshpark:
This issue is to share/discuss about _PartitionEditor_ extension that will
- provide a GUI editor for `*.part` file defined in ONE repo

Features for this editor as inital version
- define available partition backends (key for making partioned models)
- set default backend
- select how to partition (this is comply key, by operator code (`OPCODE`) or by oeprator name (`OPNAME`)
   - by operator index (actually tensor index) will be added later
- select nodes for one of the backend

Things uncertain as of now
- which model is input the input of _one-partiton_ ?

seanshpark:
As initial version of _PartitionEditor_ user scenario will be like this
- user adds a backend (we will assume this as default backend)
- initial `comply` will show `OPCODE`
- as there is only one backend, operator codes list, add/remove buttons are disabled
- user adds another backend
- operator codes list is enabled
- user clicks `add` button
   - available OPCODE list is shown, user selects one and then OK button
   - opcode is added to operator codes list
- user clicks `del` button
   - selectd opcode item is removed
- user clicks `SAVE` button
   - stores to `.part` file

For `OPNAME`
- user selects `OPNAME` for `comply` button
- if there was only one `backend`, operator name list, add/remove buttons are disabled
- user clicks on any `backend list` item
- operator name list reflects the backend
   - `(default)` for defaut backend, `edit` button is disabled
   - operator names for other backends, `edit` button is enabled
- user clicks `edit` button
   - jump to `CircleGraph` extension
   - there should be `Return to PartitonEditor(?)` menu item to return
   - Close `CircleGraph` in the middle of editing?
      - ask save?

NOTE
- `OPCODE` list and `OPNAME` list are stored separately
- Showing `PartitionEditor` and `CircleGraph` may confuse the user and the operation itself
   - while editing selection with `CircleGraph` user can close `PartitionEditor`
   - `CircleGraph` cannot continue the operator as caller is closed

seanshpark:
Something like https://docs.microsoft.com/en-us/cpp/mfc/dialog-data-exchange?view=msvc-170 can help building UI...

seanshpark:
Need to check PartitionEditor can invoke CircleGraph, act like a modal dialog
- [ ] prepare dummy PartitionEditor with a button
- [ ] show CircleGraph with the button

seanshpark:
Name: `PartitionEditor` or `PartEditor`?
[...]

PR #772, opened 2022-06-09
[PartEditor] Relocate inital BackendNames

seanshpark:
This will relocate initialization of _backEndNames to
resolveCustomTextEditor method as current position is called multiple
times while editing.


seanshpark:
related #641

jyoungyun (approved):
LGTM

PR #773, opened 2022-06-09
[PartEditor] Use normal string for template literal types

seanshpark:
This will revise to use normal string instead of using template literal types.


seanshpark:
related #641 
usage is simple case so just using normal string is enough.

dayo09 (approved):

PR #777, opened 2022-06-10
[PartEditor] Set backend code in Operators listbox

seanshpark:
This will apply backend code from the document to Operators listbox.

### DIFF
--- a/media/PartEditor/index.js
+++ b/media/PartEditor/index.js
@@ -43,6 +43,8 @@ editor.Editor = class {
     this.document = window.document;
     this.backends = [];   // array of editor.Backend
     this.operators = [];  // array of editor.Operator
+    this.partition = {};  // current partition from document
+    this.beToCode = {};   // backend string to code, for search performance
   }
 
   initialize() {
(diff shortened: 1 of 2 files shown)


seanshpark:
for #641, from draft #647

jyoungyun (dismissed):
LGTM

media/PartEditor/index.js
```diff
@@ -124,7 +135,87 @@ editor.Editor = class {
...
+
+  /**
+   * @brief apply document partition to this.operators code
+   */
+  updateOperatorsBackend() {
+    // must clear first as this.partition.OPNAME may not have all nodes
+    this.clearOperatorsCode();
+
+    for (let name in this.partition.OPNAME) {
+      if (this.partition.OPNAME.hasOwnProperty(name)) {
```
mhs4670go, line 159: Does this if condition ever become 'false'? Because `name` is an element of for loop that iterates `this.partition.OPNAME`.

seanshpark: You are right! I think this is some trace while implementation.

mhs4670go (approved):
LGTM

jyoungyun (approved):
LGTM

PR #779, opened 2022-06-10
[Backend] Use ProblemResolver to install older packages

jyoungyun:
This commit uses ProblemResolver to install older packages.

### DIFF
--- a/src/Backend/ToolchainImpl/DebianToolchain.ts
+++ b/src/Backend/ToolchainImpl/DebianToolchain.ts
@@ -70,9 +70,14 @@ class DebianToolchain implements Toolchain {
     // According to man(8) aptitude
     // -q: suppress all incremental progress indicators
     // -y: assume that the user entered "yes"
+    // -o: Set a configuration file option directly
+    //   * Aptitude::ProblemResolver::SolutionCost
+    //        : Describes how to determine the cost of a solution.
+    //          ref: https://tools.ietf.org/doc/aptitude/html/en/ch02s03s04.html
     this.prepare();
     let cmd = new Command('aptitude');
     cmd.push('install');
+    cmd.push(`-o Aptitude::ProblemResolver::SolutionCost='100*canceled-actions,200*removals'`);
     let pkg: string = this.info.name;
     if (this.info.version !== undefined) {
       pkg = `${pkg}=${this.info.version.str()}`;
--- a/src/Tests/Backend/ToolchainImpl/DebianToolchain.test.ts
+++ b/src/Tests/Backend/ToolchainImpl/DebianToolchain.test.ts
@@ -58,7 +58,9 @@ suite('Backend', function() {
         test('', function() {
           let dt = new DebianToolchain(info);
           let cmd = dt.install();
-          const expectedStr = `sudo aptitude install ${name}=${version.str()} -q -y`;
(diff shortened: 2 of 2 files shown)


jyoungyun:
The TRIV toolchain releases their package using different channels. One is for the official channel and the other is for the nightly channel. Nightly channel is released every day, but official channel is released at a certain time, so packages that are dependent on the official version are not the latest packages. The apt system does not allow to install older package, so installation of official release is failed.

In this case, `aptitude` can resolve this issue by using `ProblemResolver::SolutionCost`.

The `ProblemResolver::SolutionCost` describes how to determine the cost of a solution.
The default cost, sorting solutions by their safety cost, then by their apt pin priority.
In this PR, it changes the cost like below:
```
Aptitude::ProblemResolver::SolutionCost='100*canceled-actions,200*removals'
```
It means:
- by increasing `canceled-actions` counter, it does not keep packages at their current version.
- by increasing `removals` counter, it does not delete packages that we want to keep.

ref: https://www.debian.org/doc/manuals/aptitude/ch02s03s04.en.html

Several projects use the same approach to install a particular version.
https://www.intel.com/content/www/us/en/developer/articles/guide/installing-intel-parallel-studio-xe-runtime-2020-using-apt-repository.html
https://www.npmjs.com/package/moncoin-crypto

YongseopKim:
That's a great solution. I have a question. Why are they `100` and `200`? Is it a kind of heuristics? or a kind of threshold?

jyoungyun:
> That's a great solution. I have a question. Why are they `100` and `200`? Is it a kind of heuristics? or a kind of threshold?

Actually I don't know about this much. Several projects that need to support to install old version packages use this approach. However there is not much information the details about the counter value of `removals` and `canceled-actions`.

My estimate is that this solution allows to downgrade by prioritizing the `canceled-actions` package and then `removals` package. Here, the values of `100` and `200` are estimated to be specific values arbitrarily determined by the user rather than threshold values.

YongseopKim (approved):
LGTM

src/Backend/ToolchainImpl/DebianToolchain.ts
```diff
@@ -70,9 +70,14 @@ class DebianToolchain implements Toolchain {
...
     // -q: suppress all incremental progress indicators
     // -y: assume that the user entered "yes"
+    // -o: Set a configuration file option directly
+    //   * Aptitude::ProblemResolver::SolutionCost
+    //        : Describes how to determine the cost of a solution.
+    //          ref: https://tools.ietf.org/doc/aptitude/html/en/ch02s03s04.html
     this.prepare();
     let cmd = new Command('aptitude');
     cmd.push('install');
+    cmd.push(`-o Aptitude::ProblemResolver::SolutionCost='100*canceled-actions,200*removals'`);
```
YongseopKim, line 80: (Optional)

```ts
// NOTE: SolutionConst can be fixed
const canceledActions = 100;
const removals = 200;
const solutionConst = `'${canceledActions}*canceled-actions,${removals}*removals'`;
cmd.push(`-o Aptitude::ProblemResolver::SolutionCost=${solutionConst}`);
```

YongseopKim:
1 approval?

hyunsik-yoon (approved):

Issue #569, opened 2022-05-09
[UX] Update toolchain view


jyoungyun:
Let's try to finish the following within May. 😨 😅 🥵 🍺 

- [x] Enhanced "show installable toolchains"
  - [x] #573
    - [x] `toolchainTypes()` 
      - returns, e.g., `['official', 'nightly']`
    - [x] `toolchains()` -> `toolchains(toolchainType: string, from: number, count: number)` 
      - why? there could be, e.g., tons of versions in some toolchain type, e.g., `nightly`.
      - e.g., `toolchains('nightly', 10, 20)` returns 20 nightly versions starting from recent 10th version
    - [ ] remove deprecated code
  - [x] Frontend
     - [x] #684
  - [ ] unit test (if necessary)
    - [ ] backend impl
  - [x] #578
  - [x] Solve prerequisite to use a specific debian repo
    - e.g., registrer a specific debian repo, installing required packages, e.g., `ca-certificates`, `wget`, `gnupg2`, etc.
    - [x] #627
- [x] Show already-installed toolchain
  - [x] #659
  - [ ] function & view UI
     - [x] show list (draft: #657)
        - [x] #658 
     - [ ] show dependency information (draft: #688)
        - [x] #690
  - [ ] unit test (if necessary)
    - [ ] view UI
    - [x] backend impl
- [x] #577
   - [x] #589
   - [x] #634
   - [x] #636
- [x] Uninstall toolchain (draft: #699)
  - [x] check backend API 
     - [x] Add automatic yes and quiet option to uninstall command (PR 54)
  - [x] function & view UI
     - [x] #665 
     - [x] #707
  - [ ] unit test (if necessary)
    - [ ] view UI
- [x] #653
  - [x] check backend API
     - [x] Set install and uninstall to root (PR 45)
     - [x] Use aptitude tool when installing toolchain (PR 51)
  - [x] function & view UI
     - [x] #652
  - [ ] unit test (if necessary)
    - [ ] view UI

### Improvement point
- [ ] Revise backend API
   - [ ] `class Compiler` -> `class Toolchain`
- [ ] Toolchain view
   - [ ] Improve dependency information UI
- [ ] Toolchain API
   - [x] #704
- [ ] Install
   - [ ] frontend (one-vscode)
     - [ ] Improve UI
       - [x] #706
       - [ ] Improve progress UI
       - [ ] Encrypt user password
     - [ ] advanced functions
       - [ ] Install latest official toolchain (#779)
[...]

PR #781, opened 2022-06-10
[Toolchain] Uninstall package before installation

jyoungyun:
Debian package allows to install only one package in one machine.
This commit removes package if the same package is already installed
through interaction with users.

### DIFF
--- a/src/Project/JobRunner.ts
+++ b/src/Project/JobRunner.ts
@@ -98,6 +98,7 @@ export class JobRunner extends EventEmitter {
 
   private onCleanup() {
     this.running = false;
+    process.env.userp = '';
   }
 
   public start(path: string, jobs: WorkJobs) {
--- a/src/Project/ToolRunner.ts
+++ b/src/Project/ToolRunner.ts
@@ -161,7 +161,6 @@ export class ToolRunner {
         // TODO(jyoung): Need password encryption
         tool = `echo ${process.env.userp} | sudo -S ` + tool;
         this.child = cp.spawn(tool, toolargs, {cwd: path, shell: true});
-        process.env.userp = '';
       } else {
         this.child = cp.spawn(tool, toolargs, {cwd: path});
       }
--- a/src/View/InstallQuickInput.ts
+++ b/src/View/InstallQuickInput.ts
@@ -17,7 +17,10 @@
 import * as vscode from 'vscode';
 
 import {Toolchain} from '../Backend/Toolchain';
-import {JobCallback} from '../Project/Job';
+import {DebianToolchain} from '../Backend/ToolchainImpl/DebianToolchain';
+import {Job, JobCallback} from '../Project/Job';
+import {JobInstall} from '../Project/JobInstall';
+import {JobUninstall} from '../Project/JobUninstall';
 import {gToolchainEnvMap, ToolchainEnv} from '../Toolchain/ToolchainEnv';
 import {Balloon} from '../Utils/Balloon';
 import {Logger} from '../Utils/Logger';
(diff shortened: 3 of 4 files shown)


src/Project/ToolRunner.ts
```diff
@@ -161,7 +161,6 @@ export class ToolRunner {
         // TODO(jyoung): Need password encryption
         tool = `echo ${process.env.userp} | sudo -S ` + tool;
         this.child = cp.spawn(tool, toolargs, {cwd: path, shell: true});
-        process.env.userp = '';
```
jyoungyun, line 164: If there is already installed package, it should be removed first.
In this case, this PR requests the two requests using one job to be entered the password from the user only once. However, this code resets the user password before all requests are complete. So I remove this line and reset the password after all requests are done. (JobRunner.ts L102)

jyoungyun:
* Already installed toolchain
   - Show information message `Backend toolchain can be installed only one. Do you want to remove the existing installed toolchain?`
   - If yes, remove the existing installed toolchain and then install the selected toolchain
   - install_uninstall_yes
   - If no, show canceled information message `Installation is canceled.`
   - install_uninstall_no

from https://github.com/Samsung/ONE-vscode/issues/569#issuecomment-1150720962

src/View/InstallQuickInput.ts
```diff
@@ -157,21 +160,51 @@ export async function showInstallQuickInput() {
...
+        resolve(toolchain);
+      };
+      const failed: JobCallback = function() {
+        reject();
+      };
+
+      const installed =
+          toolchainEnv.listInstalled().filter(value => value instanceof DebianToolchain);
+      if (installed.length > 1) {
+        throw Error('One or more debian toolchains cannot be installed');
```
dayo09, line 176: How about "Installed debian toolchain must be unique"?

src/Project/JobRunner.ts
```diff
@@ -98,6 +98,9 @@ export class JobRunner extends EventEmitter {
 
   private onCleanup() {
     this.running = false;
+    if (process.env.userp !== '') {
```
YongseopKim, line 101: (Optional) This looks okay when `if condition` is removed even.

src/View/InstallQuickInput.ts
```diff
@@ -157,21 +160,51 @@ export async function showInstallQuickInput() {
...
+      const success: JobCallback = function() {
+        resolve(toolchain);
+      };
+      const failed: JobCallback = function() {
+        reject();
+      };
+
+      const installed =
+          toolchainEnv.listInstalled().filter(value => value instanceof DebianToolchain);
+      if (installed.length > 1) {
```
YongseopKim, line 175: This looks a little complicated. How about this?

```ts
if (installed.length > 1) {
    throw Error('One or more debian toolchains cannot be installed');
}

if (installed.length === 0) {
    toolchainEnv.install(toolchain, success, failed);
    return;
}

const answer = await vscode.window.showInformationMessage(
    'Backend toolchain can be installed only one. Do you want to remove the existing installed toolchain?',
    'Yes', 'No');
if (answer === 'Yes') {
    const jobs: Array<Job> = [];
    const uninstallJob = new JobUninstall(installed[0].uninstall());
    uninstallJob.failureCallback = failed;
    jobs.push(uninstallJob);
    const installJob = new JobInstall(toolchain.install());
    installJob.successCallback = success;
    installJob.failureCallback = failed;
    jobs.push(installJob);
    toolchainEnv.request(jobs);
} else {
    Balloon.info('Installation is canceled.');
    reject();
}
```

src/Project/JobRunner.ts
jyoungyun: Yes, right. I just want to express it more accurately. :) I will update it.

jyoungyun:
@dayo09 @YongseopKim PTAL

src/View/InstallQuickInput.ts
YongseopKim: FYI

YongseopKim (approved):
LGTM

PR #782, opened 2022-06-13
[PartEditor] Revise default backend array

seanshpark:
This will revise default backend array to give 'CPU' as initial value.


seanshpark:
for #641, from draft #647

this change will make upcoming changes a bit simple.

PR #783, opened 2022-06-13
[PartEditor] Handle Backend combobox change

seanshpark:
This will set backend combobox change to reflect default checkbox.

### DIFF
--- a/media/PartEditor/index.js
+++ b/media/PartEditor/index.js
@@ -73,6 +73,11 @@ editor.Editor = class {
           break;
       }
     });
+
+    // backend combobox change
+    this.document.getElementById('circle-be').addEventListener('change', () => {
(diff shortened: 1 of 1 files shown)


seanshpark:
for #641, from draft #647

`default` checkbox is to set default backend to the backend of combobox
- `check` is to change default to the backend of combobox
- `uncheck` is to change default to the 0'th item (CPU as of now)
- `uncheck` to CPU is not possible as it is logical error
- current PR disables `check` button if default is CPU

`check` is disable as current default is CPU
image

`check` is enabled as current default is not ACL_CL
image

jyoungyun (approved):
LGTM

mhs4670go (approved):
LGTM

PR #785, opened 2022-06-14
[PartEditor] Introduce makeDefaultPartiton method

seanshpark:
This will introduce Partition class and makeDefaultPartiton method
for default 'partition' section.

### DIFF
--- a/src/PartEditor/PartEditor.ts
+++ b/src/PartEditor/PartEditor.ts
@@ -24,6 +24,12 @@ import {Balloon} from '../Utils/Balloon';
 import {getNonce} from '../Utils/external/Nonce';
 import {Logger} from '../Utils/Logger';
 
+type Partition = {
+  backends?: {};
+  default?: {};
+  comply?: {};
+};
+
 export class PartEditorProvider implements vscode.CustomTextEditorProvider {
   public static readonly viewType = 'onevscode.part-editor';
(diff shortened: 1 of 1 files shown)


seanshpark:
for #641, from draft #647

src/PartEditor/PartEditor.ts
```diff
@@ -126,6 +132,21 @@ export class PartEditorProvider implements vscode.CustomTextEditorProvider {
...
 
+  private makeDefaultPartiton() {
+    let partition: Partition = {};
+    let backends = '';
+    for (let idx = 1; idx < this._backEndNames.length; idx++) {
+      if (backends.length) {
+        backends = backends + ',';
+      }
+      backends = backends + this._backEndNames[idx];
+    };
```
mhs4670go, line 143: How about this one?
```suggestion
    let backends = this._backEndNames.join(",");
    };
```

seanshpark: `let idx = 1;` starts from second item.
is there something like copy array from second item?

mhs4670go: Oh, my bad. Then, how about this one?
```typescript
let backends = this._backEndNames.slice(1, this._backEndNames.length).join(',');
```

mhs4670go (approved):
LGTM

jyoungyun (approved):
LGTM

PR #787, opened 2022-06-14
[PartEditor] Handle default check with update document

seanshpark:
This will handle default check button to set default backend
and update the document with changed default backend value.


seanshpark:
for #641, from draft #647

before click on checkbox
image

after clinck on checkbox
image

jyoungyun (approved):
LGTM

mhs4670go (approved):
LGTM

PR #791, opened 2022-06-15
[CircleGraphCtrl] Introduce load finish handler

seanshpark:
This will add handler for model has finished loading model and graph is ready.


seanshpark:
for #641, from draft #647

PR #792, opened 2022-06-16
[CircleGraphCtrl] Revise initGraphCtrl to accept model path

seanshpark:
This will revise initGraphCtrl to accept model path.


seanshpark:
for #641, from draft #647

this change is needed for PartGraphSelector as it makes model path from part file path.

mhs4670go (approved):
LGTM

PR #793, opened 2022-06-16
[CircleGraphCtrl] Revise member scope to protected

seanshpark:
This will revise member scope to protected.


mhs4670go (approved):
LGTM

PR #797, opened 2022-06-16
[PartEditor] Handle Set with graph view button

seanshpark:
This will enable to handle 'Set with graph view' button.
Actual switching to graph view is left todo.


seanshpark:
for #641, from draft #647

mhs4670go (approved):
LGTM

PR #798, opened 2022-06-16
[CircleGraphCtrl] Introduce CircleGraphEvent

seanshpark:
This will introduce CircleGraphEvent interface so that owner of
CircleGraphCtrl can handle events(loading) of the ctrl.


seanshpark:
for #641, from draft #647

mhs4670go (approved):
LGTM

PR #800, opened 2022-06-16
[PartEditor] Notify for selection change

seanshpark:
This will enable PartGraphSelPanel to send selection change notification to owner.


seanshpark:
for #641, from draft #647

PR #803, opened 2022-06-17
[PartEditor] More commands for PartGraphSelector

seanshpark:
This will add more commands as
- close PartGraphSelector by command
- refresh selection state by command


seanshpark:
for #641, from draft #647

Issue #813, opened 2022-06-21
[Toolchain] Support default toolchain


jyoungyun:
## What

Let's define the default toolchain.

jyoungyun:
Done!

PR #818, opened 2022-06-21
[Toolchain] Introduce default toolchain

jyoungyun:
This commit introduce default toolchain.
    
Related issue: #813

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,6 +151,11 @@
         "icon": "$(x)"
       },
       {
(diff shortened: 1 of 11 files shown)


jyoungyun:
The blue circle icon in front of `triv2-toolchain-latest` in Toolchain view means that this is a default toolchain. If there were many toolchains here, only one toolchain would have a blue circle icon.

jyoungyun:
This draft is not valid anymore. I will close it.

PR #814, opened 2022-06-21
DRAFT CircleGraphCtrl view mode

seanshpark:
on-going draft to provide view mode as viewer or selector of CircleGraphCtrl.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #790, opened 2022-06-15
[CircleGraph] as circle model graph viewer


seanshpark:
Current `CircleGraphPanel` was introduced as development for GUI node selection like Partition Editor.

As basic feature is to show `.circle` model, we can provide this panel as `.circle` model viewer.

Roughly we need to 
- add a mode in `CircleGpaphCtrl` as just a viewer and as a node selector
- provide same feature like `netron` for viewer mode

seanshpark:
I'll work on this after partition editor feature is enabled.

seanshpark:
How to set mode?
- mode related switches are for menu
- menu creation is done at startup
- this, need to set mode in global when loading

seanshpark:
CircleViwer to do features
- [ ] update view when file has changed
- [ ] ...

seanshpark:
done for now

PR #819, opened 2022-06-21
[CircleGraph] Introduce initial view mode

seanshpark:
This will introduce initial view mode global variable, to set
CircleGraph as viewer or node selector, as of now.


seanshpark:
fro #790, from draft #814

Issue #824, opened 2022-06-22
[PartEditor] About backend names


seanshpark:
As extension for OneRT is not under development, current backend names(strings) are fixed and will be revised when that extension is available.

Current settings are `CPU/ACL_CL/NPU` but this NPU now has its name `TRIX`.

PR #825, opened 2022-06-22
[PartEditor] USe TRIX

seanshpark:
This will use TRIX to replace NPU in backend names.


seanshpark:
For #824

jyoungyun (approved):
LGTM

PR #827, opened 2022-06-22
[CircleGraph] Introduce host view mode

seanshpark:
This will introduce view mode in host object.


seanshpark:
for #790, from draft #814

jyoungyun (approved):
LGTM

PR #829, opened 2022-06-22
[CircleGraph] Apply condition for selector view mode

seanshpark:
This will apply condition for codes that should work for selector view mode.


seanshpark:
for #790, from draft #814

PR #832, opened 2022-06-22
[Toolchain] Add runCfg command in toolchain provider

jyoungyun:
This commit adds runCfg command in toolchain provider.
And it hides this command from command palette because it can't
work without cfg file.

Draft PR: #818

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,6 +151,10 @@
         "icon": "$(x)"
       },
       {
+        "command": "one.toolchain.runCfg",
+        "title": "Run"
+      },
+      {
         "command": "one.device.register",
         "title": "Register",
         "category": "ONE",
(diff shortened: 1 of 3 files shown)


YongseopKim (approved):
LGTM

package.json
```diff
@@ -150,6 +150,10 @@
         "category": "ONE",
         "icon": "$(x)"
       },
+      {
+        "command": "one.toolchain.runCfg",
+        "title": "Run"
```
struss, line 155:
```suggestion
        "command": "one.toolchain.runCfg",
        "title": "Run",
        "category": "ONE"
```
I think category missed?

llFreetimell (approved):
~~LGTM~~

package.json
jyoungyun: The `category` does not need in this case. This is because it is not be provided in the command palette. :)

jyoungyun:
Please merge this PR /cc @llFreetimell. /cc @struss  :)

Issue #563, opened 2022-05-04
[CfgEditor] Introduce Quantize tab in Advanced Options


llFreetimell:
#### Before

- `Quantize` tab in `Advanced Options` was not implemented

#### After

- UI for `Quantize` tab in `Advanced Options` is added
- Content in `.cfg` file is displayed in `Quantize` tab

---

#### Todo/Done

- Add panel UI for `Quantize` tab
- Display `one-quantize` content in `.cfg` file

llFreetimell:
### Considerations for UI

- Parameter Lists

```
[Normal Quantization]
<quantDequantPass>
verbose=T/F
input_model_dtype, input_dtype
quantized_dtype
granularity
input_path
generate_profile_data=T/F

<recordMinMax>
verbose=T/F
input_data
input_data_format
min_percentile
max_percentile
mode
generate_profile_data=T/F

<quantizeWithMinMax>
verbose=T/F
input_model_dtype, input_dtype
quantized_dtype
granularity
TF-style_maxpool=T/F
input_type
output_type
quant_config
output_path
generate_profile_data=T/F

[ForceQuant]
verbose=T/F
tensor_name
scale
zero_point
input_path
output_path

[COpyQUant]
verbose=T/F
src_tensor_name
dst_tensor_name
input_path
output_path
```

llFreetimell:
### Parameters in `circle-quantizer`

#### Action

only one of following option can be used

- quantize_dequantize_weights
  - Mandatory
    - input_model_dtype (float32)
    - output_model_dtype (uint8)
    - granularity (layer, channel)
  - Optional
    - config
- quantize_with_minmax
  - Mandatory
    - input_model_dtype (float32)
    - output_model_dtype (uint8)
    - granularity (layer, channel)
  - Optional
    - input_type (float32, uint8, int16)
    - output_type (float32, uint8, int16)
    - TF-style_maxpool
    - config
- requantize
  - Mandatory
    - input_model_dtype
    - output_model_dtype
- force_quantparam
  - Mandatory
    - source_tensor_name
    - scale
    - zero_point
- copy_quantparam
  - Mandatory
    - source_tensor_name
    - destination_tensor_name
- fake_quantize

#### Etc

- Mandatory
  - input
  - output
- Optional
  - generate_profile_data
  - verbose

llFreetimell:
Done!

PR #838, opened 2022-06-23
[CfgEditor] Add Quantize Step UI

llFreetimell:
This commit adds Quantize Step UI.

---

Parent Issue : #563

---

#### Before & After

| UI | Quantize Tab | :x: NYI | quantize |

### DIFF
--- a/media/CfgEditor/cfgeditor.css
+++ b/media/CfgEditor/cfgeditor.css
@@ -151,7 +151,9 @@ label {
 }
 
 #optionImport .basic,
-#optionImport .advanced {
+#optionImport .advanced,
+#optionQuantize .basic,
+#optionQuantize .advanced {
   display: none;
 }
 
--- a/media/CfgEditor/cfgeditor.html
+++ b/media/CfgEditor/cfgeditor.html
@@ -246,7 +246,208 @@ limitations under the License.
             </div>
         </div>
         <div class="options" id="optionQuantize">
-            <!-- TODO Implement Quantize step -->
+            <div class="prerequisite">
+                <div class="title">
+                    Basic Options
+                </div>
+                <div class="option">
+                    <vscode-radio-group id="quantizeActionType">
+                        <label slot="label">Quantization Action Type</label>
+                        <vscode-radio value="defaultQuant" checked="true">Quantization</vscode-radio>
+                        <vscode-radio value="forceQuant">Force Quantization Param</vscode-radio>
+                        <vscode-radio value="copyQuant">Copy Quantization Param</vscode-radio>
+                    </vscode-radio-group>
+                </div>
+            </div>
+            <div class="basic" id="optionQuantizeDefaultQuantBasic">
+                <div class="option">
+                    <vscode-text-field id="DefaultQuantInputPath" startIcon="true" size="50" placeholder="">
+                        Input Path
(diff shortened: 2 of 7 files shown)


jyoungyun (approved):
LGTM

YongseopKim (approved):
LGTM

jinevening:
The UI looks very good to me. I'd like to say just a few things.

1. Mode: Why `moving_average` mode was removed?
2. Input Type, Output Type: It would be better to explain what `default` means. For example `default (same with quantized data type)` or a question mark like the one next to TF_Style_Maxpool.
3. Advanced Options: What kind of options come here? I think `mode` (and `min_percentile`, `max_percentile`) can come here. Those options are not adjusted in most cases.
4. Quantization Config File: It would be better to explain what is the role of config file. For example `Quantization Config File (for mixed-precision quantization)` or a question mark.
5. Input Model Data Type: It has only one option (float32), so I think this can be removed.
6. Input Data: It would be better to explain what is the role of input data. For example, using a question mark. `Input Data (?)` If cursor is placed on the question mark, the explanation may pop up like `Dataset used to quantize activation. If not given, random data will be used for quantization, which will seriously degrade the accuracy of the model.`
7. Input Data Format: It would be better to add explanation of each format. Maybe using question marks or external links.

llFreetimell:
> Mode: Why moving_average mode was removed?

My mistake :crying_cat_face: 

> Input Type, Output Type: It would be better to explain what default means. For example default (same with quantized data type) or a question mark like the one next to TF_Style_Maxpool.

I changed `default` to `quantized data type` :)

> Advanced Options: What kind of options come here? I think mode (and min_percentile, max_percentile) can come here. Those options are not adjusted in most cases.

My intention is following :)

- Basic options : Mandatory/strongly required options. Options which are directly related with quantization.
- Advanced options : Usually not used options. Not important options. Options which are not related with quantization.

So I put `mode` at basic options.
However, it is not that often used, I will move it to Advanced Options!
(I didn't notice that `min_percentile` and `max_percentile` are only used for `percentile`..!)

> Quantization Config File: It would be better to explain what is the role of config file. For example Quantization Config File (for mixed-precision quantization) or a question mark.

Added question mark :D

> Input Model Data Type: It has only one option (float32), so I think this can be removed.

If there will be only one option(float32) in the future, removing would be better.
However, if it is not that sure, I think leaving as it is seems not that bad :)

> Input Data: It would be better to explain what is the role of input data. For example, using a question mark. Input Data (?) If cursor is placed on the question mark, the explanation may pop up like Dataset used to quantize activation. If not given, random data will be used for quantization, which will seriously degrade the accuracy of the model.

Added question mark :D
FYI, all of question marks in the editor are not implemented yet because of other important features T.T

> Input Data Format: It would be better to add explanation of each format. Maybe using question marks or external links.

Added question mark :D

jinevening (approved):
LGTM Thank you very much

YongseopKim (approved):
LGTM 👍

jyoungyun (approved):
LGTM

hyunsik-yoon:
(optional for future) 
0. GUI shows description and cfg shows arg.
- e.g. `quantization config file` in GUI changes `quant_config` in cfg. I was thinking that what if we mention (`quantization config file (--quant_config)` in GUI.

2. The word `input` may mean two things to users: `input model` or `input data`. I hope we use  `input model` or `input data` instead of `input`.

Issue #823, opened 2022-06-22
Provide way to access error log when error happens


hyunsik-yoon:
from https://github.com/Samsung/ONE-vscode/pull/811/files#diff-89282f0acb59e2be3eec68bb6254e840704023508e3c3499708a44bbe5ca9982R23-R28

Some user does not open ONE-vscode output.
In such case, we have to provide them better way to access log.

Would showing 'Show Log' button in notification message box help users?

![showLog](https://user-images.githubusercontent.com/63768207/174924785-cc55d410-f34b-4311-b7a7-a67a67d889b7.gif)

jyoungyun:
When developer call `Logger.show();` function, it changes to focus on our `ONE-vscode` output log. But I'm not sure it is good to show the output log directly even though user does not want to see it.

Related PR: https://github.com/Samsung/ONE-vscode/pull/815

PR #860, opened 2022-06-27
Show `See log` button in Balloon

hyunsik-yoon:
This adds 'OK' and 'See log' button in methods in Balloon.
When 'See log' button is pressed, log window will be opened.

This also changes some methods in OneExplorer.ts
to show this new Balloon. (as a demonstration)

for #823

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -522,6 +522,7 @@ export class OneExplorer {
 //
 
 import {EventEmitter} from 'events';
+import {Balloon} from '../Utils/Balloon';
 
 class OneccRunner extends EventEmitter {
   private startRunningOnecc: string = 'START_RUNNING_ONECC';
@@ -572,8 +573,7 @@ class OneccRunner extends EventEmitter {
               this.emit(this.finishedRunningOnecc, value);
             })
             .catch(value => {
-              vscode.window.showErrorMessage(
-                  `Error occured while running: 'onecc --config ${this.cfgUri.fsPath}'`);
+              Balloon.error(`Error occured while running: 'onecc --config ${this.cfgUri.fsPath}'`);
               reject();
             });
       });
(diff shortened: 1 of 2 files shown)


hyunsik-yoon:
seeLog

YongseopKim (approved):
LGTM

hyunsik-yoon:
@jyoungyun 

> Why don't you get rid of Bubble and use basic api if necessary for each individual module?

To handle button like, `show log`, we need a function that handles click, which can be a common code. Because of this, I think having common class such as `Bubble` is not bad idea when we have `show log`.

> Should 'see logs' be visible in all cases of using Bubble?

In case of error or warning, I hope it to be default button.
In case of info, not having 'show log' could be default option.

What if I change API like the following?

```ts
class Balloon {
  // default is 'show seeLog button'
  error(msg: string, showSeeLogBtn?: boolean = true) { .. }
  warning(msg: string, showSeeLogBtn?: boolean = true) { .. }

  // default is 'do not show seeLog button'
  info(msg: string, showSeeLogBtn?: boolean = false) { .. }
}
```

jyoungyun (dismissed):

jyoungyun:
> What if I change API like the following?
> 
> ```ts
> class Balloon {
>   // default is 'show seeLog button'
>   error(msg: string, showSeeLogBtn?: boolean = true) { .. }
>   warning(msg: string, showSeeLogBtn?: boolean = true) { .. }
> 
>   // default is 'do not show seeLog button'
>   info(msg: string, showSeeLogBtn?: boolean = false) { .. }
> }
> ```

I like your suggestion. 

It would be better if it could be expanded more and used everywhere. I will apply it.
Thank you for considering my comments. :)

hyunsik-yoon:
I will fix and update this PR.

hyunsik-yoon:
Now
Error shows 'See Log` button by default.
Info does NOT show 'See Log' button by default.

seeLog

hyunsik-yoon:
PTAL

yunjayh (approved):

YongseopKim (approved):

jyoungyun (approved):
LGTM

PR #861, opened 2022-06-27
[PartEditor] Revise to use becode

seanshpark:
This will revise attribute code as becode to be more specific as backend code.


mhs4670go (approved):
LGTM

PR #863, opened 2022-06-27
[CfgEditor] Add '--save-allocations' codegen option

ai-moiseev:
This commit adds checkbox in config editor which allows to automatically add or remove `--save-allocations` CLI codegen option in command line text field.

image

Issue: #395 

### DIFF
--- a/media/CfgEditor/cfgeditor.html
+++ b/media/CfgEditor/cfgeditor.html
@@ -467,6 +467,12 @@ limitations under the License.
                         Command
                     </vscode-text-field>           
                 </div>
+                <div class="option">
+                    <vscode-checkbox id="codegenSaveAllocations">
+                        Save allocations trace
+                    </vscode-checkbox>
+                    <span title="Save allocation trace as *.tracealloc.json file for supported backends" class="codicon codicon-question" style="cursor: pointer"></span>
+                </div>
             </div>
         </div>
         <div class="options" id="optionProfile">
--- a/media/CfgEditor/index.js
+++ b/media/CfgEditor/index.js
@@ -366,12 +366,24 @@ function registerCopyQuantOptions() {
 function registerCodegenOptions() {
   const codegenBackend = document.getElementById('codegenBackend');
   const codegenCommand = document.getElementById('codegenCommand');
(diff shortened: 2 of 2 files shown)


llFreetimell:
Thank you for your interest to cfg editor :D
I appreciate for your effort and interest for cfg editor, but because of following reason I am hard to approve this PR... :crying_cat_face: 

- `--save-allocations` option is not in `one-codegen`
  - **This is the most important reason.** cfg editor only shows parameters in cfg files.
  - However, `--save-allocations` options is in neither `one-codegen` and cfg file.
  - In addition, `--save-allocations` option is backend specific, not general.
- automatic completion in cfg editor should be minimum
  - There are no automatic completion feature yet in cfg editor but some will be introduced in the future.
  - However, too much automatic completion or ambiguous features may cause confusion for users.
    - e.g. If backend is `abc`, is this option still valid?
    - e.g. click `--save-allocations`, and then remove "--save-allocations" in `command`. Then command does not have `save-allocations` but checkbox is still clicked. Which one is correct?

How about you? If you still think that this PR is really needed, please let me know your thoughts :)

ai-moiseev:
I agree on all points :) I suppose there's need for various declared common options in `one-codegen` (at least for the `-o` option, I think it's generic enough for most of the backends :smiley:). And for options which may work in one backend, but not available in others, I think it would be nice to have some '_capability_' protocol (like for LSP servers in VSCode) - probably implemented as XML or JSON metadata file which gets shipped alongside codegen binary and states which features can be used. But it requires a defined standard between all backends which may be not easy to achieve, as backends are maintained as a separate projects.

Some time ago I've tried to introduce this option wrapper to `one-codegen`: https://github.com/Samsung/ONE/pull/9104
But it seems the scope of one-codegen is currently only to select backend binary and pass CLI argument string 'as-is'.

With this PR I've thought to introduce a bit of UX discoverability for a various possible backend options to provide users an opportunity to experiment with options. But I agree that it may confuse the user when some options will work and some others will cause compiler to exit with an error. So currently it would be probably better to just ask the user to skim through backend documentation for available options... :disappointed:

PR #864, opened 2022-06-28
[CircleGraph] Tidy 'about' style

seanshpark:
This will tidy 'about' styles which are not used anymore.


mhs4670go (approved):
LGTM

PR #851, opened 2022-06-27
DRAFT Revise style of PartEditor and CircleGraph

seanshpark:
on-going draft to revise style of PartEditor and CircleGraph.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #822, opened 2022-06-22
[PartEditor] UX enhancement


seanshpark:
From a little discussion with @lemmaa for our initial partition editor UI,
we can;
(1) partition editor view and graph view needs some revision as there are strange mismatch in UX.
current initial version didn't regard any UX so this is expected.
(2) discuss how partition editor should be (in long term) for best UX

This issue is to describe about (1) how it would be better to match two views.

seanshpark:
(1) about node listbox selection and graph view node selection
- these two should match
- use netron selection back to node selection
- sync selection state of listbox and graph
- sync selection visibility of listbox and graph
- match color of selection would be better
- use color for each backend, like CPU as RED, GPU as GREEN, NPU as PURPLE
- sync color of backend assignment also to graph nodes
- use default gray for unassigned in graph (not netron color scheme)

(2) about node listbox
- show operator code name too
- add a search edit to filter nodes
- by filter nodes, we can multi-select for same operator code (like select all Conv2D)

seanshpark:
Todo
- [ ] apply different colors for each backend; text color
   - default: black, cpu: cyan, acl_cl: red, trix: green
   - [ ] apply to Partition Editor view; combobox and listbox
   - [ ] apply to Graph view
- [ ] ...

seanshpark:
`view` class structure
- view.Graph = class extends grapher.Graph
   - this.view = view.View
   - this.model = model
- view.Node = class extends grapher.Node
   - this.context = view.Graph

seanshpark:
> selection change of listbox --> selection change of graph
> selection change of graph --> selection change of listbox

currently, there can be multiple views for single document
if selection of listbox is to update selection of graph,
then selection of graph should sync selection of listbox.
thus, for multiple views for one document should sync with all the listbox of the webviews.

try to enable this feature. see hows the UX comes out.

seanshpark:
While working with `selection` information forwarding from `PartEditor` to `PartGraphSelector`...
- there can be multiple `PartEditor` (by split) per document
- let `PartGraphSelector` be 1:1 with `PartEditor`
   - for this, need to revise `PartGraphSelector` panel to be multiple instance
   - for this, `PartEditor` and `PartGraphSelector` should have each others ID to distinguish

seanshpark:
Current `PartEdit` implementation seems wrong.
- `PartEditorProvider` is single instance
- attributes per document should move to separate class

seanshpark:
[...]

PR #867, opened 2022-06-28
[CircleGraphCtrl] Introduce disposed state

seanshpark:
This will introduce disposed state for safety.


seanshpark:
for #822, from draft #851

mhs4670go (approved):
LGTM

Issue #848, opened 2022-06-24
[CircleViewer] Fix to follow color theme of VSCode


seanshpark:
Current circle viewer, as based on netron, follows color theme of desktop not VSCode.
Let's fix this to follow VSCode.

seanshpark:
done

PR #869, opened 2022-06-28
[CirclrGraphCtrl] Enable system theme change

seanshpark:
This will introduce method to handle system theme change.


seanshpark:
for #848, from draft #851

mhs4670go (approved):
LGTM

PR #872, opened 2022-06-28
[CircleGraphCtrl] Revise style to follow vscode

seanshpark:
This will revise css to follow VSCodes dark theme vscode-dark instead of system dark theme.


seanshpark:
for #848, from draft #851

NOTE: 
- `.vscode-dark.default` and `.vscode-dark.welcome` don't have space between `vscode` and `default/welcome`
- this is cause `default` and `welcome` are class name applied to `body`
- thus have to be like `.vscode-dark AND default class applied`
- for others are like `elements child of body` so `.vscode-dark .graph` having space in between

mhs4670go (approved):
LGTM

Issue #870, opened 2022-06-28
Prepare `Release` and `Debug` build


hyunsik-yoon:
Currently we use _debug_ build to build ONE-vscode.
We now need to prepare _release_ build.

What is the difference between debug and release build with ONE-vscode?

- Debug build
  - put everything 
- Release build
  - remove '[debug]' log

What else?

Note: tsc compilers (https://www.typescriptlang.org/docs/handbook/compiler-options.html) does not seems to have optimization flgas.

hyunsik-yoon:
### How to implement:
- env var? (like https://stackoverflow.com/a/59908595)
  - it won't work if ONE-vscode works after installing vsix.

YongseopKim:
> What is the difference between debug and release build with ONE-vscode?

Q. Is it meaning that `npm run compile` **manually** removes `debug log` when it is built on release mode? or tsc compiler automatically removes `debug log`?

hyunsik-yoon:
@YongseopKim,

For now, as far as release-build in a form of vsix does not show `debug` log, it would be great.

(Main purpose of running one-vscode by pressing F5 is for debugging... so we could consider it as debug-build.....)

hyunsik-yoon:
> So, can I understand like below in ONE-vscode?
>  - a kind of debug build: run in debug mode -> Logger.debug() prints
>  - a kind of release build: install/run vsix -> Logger.debug() doesn't print

Yes. 

> And, we don't have to prepare release build? because vsix is done with release build.

I tried to find the difference between release build and debug build in case of JS or TS project. However, it seems that there are not many difference between them.

So I'd like to focus only on `Logger.debug()` in this issue.

hyunsik-yoon:
Let me rename this.

hyunsik-yoon:
logger.debug was handled well. Let me close.

PR #880, opened 2022-06-28
Show debug log only when it runs in vscode debug mode

hyunsik-yoon:
This shows debug log only when it runs in vscode debug mode.
When ONE-vscode is packaged into vsix, debug log won't be shown.

For https://github.com/Samsung/ONE-vscode/issues/870#issuecomment-1168453781

### DIFF
--- a/src/Utils/Logger.ts
+++ b/src/Utils/Logger.ts
@@ -18,6 +18,8 @@ import * as vscode from 'vscode';
 
 type MsgList = (number|boolean|string|object)[];
 
+const isDebugMode = process.env.VSCODE_DEBUG_MODE === 'true';
+
 export class Logger {
   static outputChannel = vscode.window.createOutputChannel('ONE-VSCode');
   static firstFocus: boolean;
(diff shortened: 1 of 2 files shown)


hyunsik-yoon:
Result:

code below 
```
  Logger.info(tag, 'one-vscode activate OK');
  Logger.debug(tag, 'Printed by Logger.debug');
  Logger.info(tag, '-----------Does the above was printed?----------');
```
with vsix
image

ran in vscode debug model (after pressing F5)
image

hyunsik-yoon:
https://stackoverflow.com/a/53867084 was referred.

There could be other way by using classes in `vscode.debug` namespace. However, I thought that using `vscode.debug` seems overkill for the purpose of this PR.

src/Utils/Logger.ts
```diff
@@ -18,6 +18,8 @@ import * as vscode from 'vscode';
 
 type MsgList = (number|boolean|string|object)[];
 
+const isDebugMode = process.env.VSCODE_DEBUG_MODE === 'true';
```
YongseopKim, line 21: IHMO, what happens if `process.env.VSCODE_DEBUG_MODE` is undefined? If something bad happens, the code could be
```ts
const isDebugMode = (process.env.VSCODE_DEBUG_MODE !== undefined) && (process.env.VSCODE_DEBUG_MODE === 'true');
```

YongseopKim (approved):
LGTM

src/Utils/Logger.ts
hyunsik-yoon: The following code works well without compilation or runtime error. :-)

```ts
let a = undefined;
console.log(a !== 'true');  // this prints true
```

YongseopKim: 👍

YongseopKim: LGTM!

PR #883, opened 2022-06-29
[PartEditor] Introduce PartEditor class

seanshpark:
This will introduce PartEditor class to hold an instance for each document.


seanshpark:
Codes are copied from `PartEditorProvider` class.

mhs4670go (approved):
LGTM

PR #887, opened 2022-06-29
[OneExplorer] Rewrite using vscode.workspace.fs

dayo09:
Let's rewrite filesystem functions using vscode.workspace.fs.

Reason 1. It can prevents invalid symlink error.
Reason 2. We can believe in vscode-extension which prevents possible, as they recommend.


dayo09:
This commit requires too much changes. Let's introduce a wrapper first. #1021

PR #893, opened 2022-06-30
[CircleGraph] Add keyCtrl and keyShift

seanshpark:
This will add keyCtrl and keyShift members in view.View to utilize key
state of control and shift.


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

PR #895, opened 2022-06-30
[CircleGraph] Use let

seanshpark:
This will revise to use let instead of var.


mhs4670go (approved):
LGTM

PR #896, opened 2022-06-30
[CirclrGraph] set category style only for viewer mode

seanshpark:
This will revise CircleGraphCtrl to set category style of the node only for
viewer mode.


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

PR #897, opened 2022-06-30
[CircleGraph] clang format off of get class

seanshpark:
This will disable clang format for class method getter.


mhs4670go (approved):
LGTM

PR #899, opened 2022-06-30
[PartEditor] Comments for message from web view

seanshpark:
This will add some comments for message from web view.


mhs4670go (approved):
LGTM

PR #901, opened 2022-06-30
[CircleGraph] Remove double host

seanshpark:
This will fix to remove double declare of host const.


media/CircleGraph/view.js
```diff
@@ -1285,9 +1285,6 @@ view.Node = class extends grapher.Node {
             (node.name || node.location);
         const title = header.add(null, styles, content, tooltip);
 
-        // this.context = view.Graph
-        // this.context.view = view
-        const host = this.context.view._host;
```
seanshpark, line 1290: this three lines already exist at the top of this method.

mhs4670go (approved):
LGTM

PR #903, opened 2022-06-30
[PartEditor] Revise backend combobox

seanshpark:
This will revise backend combobox value string generation.


mhs4670go (approved):
LGTM

PR #904, opened 2022-06-30
[CircleGraph] toggleSelect with ctrl key

seanshpark:
This will revise toggleSelect method handle with ctrl key pressed.
- without ctrl: select only the node
- with ctrl: toggle node selection


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

Issue #906, opened 2022-06-30
[OneExplorer] Add delete command


dayo09:
## What?

Let's create delete command for OneExplorer.

### vscode explorer UI inspection

To follow the vscode explorer basic UI, let's inspect the messages first.

#### Delete file
![vscode-delete-file](https://user-images.githubusercontent.com/17171963/176679095-543818bc-a744-4d33-a83e-315695c576f8.png)

```
Are you sure you want to delete 'filename.txt'?

You can restore this file from the Trash.

Cancel / Move to Trash
```

#### Delete directory
![vscode-delete-dir](https://user-images.githubusercontent.com/17171963/176679062-7c158f4c-2ca2-4280-b6bd-6eb626670674.png)

```
Are you sure you want to delete 'name' and its contents?
You can restore this file from the Trash.

Cancle / Move to Trash
```

PR #729, opened 2022-05-27
[OneExplorer] Support keybindings 

dayo09:
This commit supports keybindings 'rename' 'delete' 'refresh'
on OneExplorer.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

PR #907, opened 2022-06-30
[OneExplorer] Add delete command

dayo09:
This commit adds delete command on every directory and files.
Its UI basically looks after vscode's default explorer.
It asks the user once more whether to delete.

---

For #906

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -232,7 +232,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   delete(oneNode: OneNode): void {
     if (oneNode.node.type === NodeType.directory) {
       vscode.window
-          .showWarningMessage(
+          .showInformationMessage(
               `Are you sure you want to delete '${
                   path.parse(oneNode.node.path).base}' and its contents?`,
               {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
@@ -246,7 +246,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
           .then(() => this.refresh());
     } else {
       vscode.window
-          .showWarningMessage(
+          .showInformationMessage(
               `Are you sure you want to delete '${path.parse(oneNode.node.path).base}'?`,
               {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
           .then(ans => {
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -230,30 +230,32 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   }
 
   delete(oneNode: OneNode): void {
+    const isDirectory = (oneNode.node.type === NodeType.directory);
+    const title =
+        `Are you sure you want to delete '${path.parse(oneNode.node.path).base}'` + isDirectory ?
+        'and its contents?' :
+        '?';
+    const detail = `You can restore this file from the Trash.`;
+    const approval = 'Move to Trash';
+    const recursive = isDirectory ? true : false;
+
     if (oneNode.node.type === NodeType.directory) {
-      vscode.window
-          .showInformationMessage(
-              `Are you sure you want to delete '${
-                  path.parse(oneNode.node.path).base}' and its contents?`,
-              {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
+      vscode.window.showInformationMessage(title, {detail: detail, modal: true}, approval)
           .then(ans => {
-            if (ans === 'Move to Trash') {
-              Logger.debug('OneExplorer', `Delete '${path.parse(oneNode.node.path).base}'.`);
+            if (ans === approval) {
+              Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
               return vscode.workspace.fs.delete(
-                  oneNode.node.uri, {recursive: true, useTrash: true});
+                  oneNode.node.uri, {recursive: recursive, useTrash: true});
             }
           })
           .then(() => this.refresh());
     } else {
-      vscode.window
-          .showInformationMessage(
-              `Are you sure you want to delete '${path.parse(oneNode.node.path).base}'?`,
-              {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
+      vscode.window.showInformationMessage(title, {detail: detail, modal: true}, approval)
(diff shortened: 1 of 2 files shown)


dayo09:
### Screenshots

You can see that the messages looks after the vscode default messages 
: https://github.com/Samsung/ONE-vscode/issues/906#issue-1290077415


#### Delete a directory

0630-delete-directory

#### Delete a file

0630-delete-file

yunjayh:
Q) (Just for curiosity for now) Can we bind this command with hot key? for example `delete` key. As we talked last meeting, bind rename to `F2` key is great to have.

dayo09:
@yunjayh Sure :-D I am trying to find how to. I think, at the time I tried https://github.com/Samsung/ONE-vscode/pull/729 before, I had problem getting a node and get uri. Maybe with https://github.com/Samsung/ONE-vscode/pull/908, we can obtain uri from the tree, still. PTAL :-D

src/OneExplorer/OneExplorer.ts
```diff
@@ -229,6 +229,37 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+              `Are you sure you want to delete '${
+                  path.parse(oneNode.node.path).base}' and its contents?`,
+              {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
+          .then(ans => {
+            if (ans === 'Move to Trash') {
+              Logger.debug('OneExplorer', `Delete '${path.parse(oneNode.node.path).base}'.`);
+              return vscode.workspace.fs.delete(
+                  oneNode.node.uri, {recursive: true, useTrash: true});
+            }
+          })
```
yunjayh, line 245: Q) `Balloon.info` has same structure with this code. But the `balloon`'s `showInformationMessage` promise pop up the message at right bottom of vscode window, and this code show the pop up a window which looks like OS-native. May I ask what's the difference..?

yunjayh (approved):
LGTM

jyoungyun:
It would be better to place `Delete` at the bottom of the mouse right-click context menu.

src/OneExplorer/OneExplorer.ts
```diff
@@ -229,6 +229,37 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+  delete(oneNode: OneNode): void {
+    if (oneNode.node.type === NodeType.directory) {
+      vscode.window
+          .showInformationMessage(
+              `Are you sure you want to delete '${
+                  path.parse(oneNode.node.path).base}' and its contents?`,
+              {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
+          .then(ans => {
+            if (ans === 'Move to Trash') {
+              Logger.debug('OneExplorer', `Delete '${path.parse(oneNode.node.path).base}'.`);
```
hyunsik-yoon, line 241: How about chaingint this to `info`?
IMHO, deletion is important event and it would be better to have such log to track state when user encounters some issue with release vsix.

src/OneExplorer/OneExplorer.ts
```diff
@@ -229,6 +229,37 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+          })
+          .then(() => this.refresh());
+    } else {
+      vscode.window
+          .showInformationMessage(
+              `Are you sure you want to delete '${path.parse(oneNode.node.path).base}'?`,
+              {detail: `You can restore this file from the Trash.`, modal: true}, 'Move to Trash')
+          .then(ans => {
+            if (ans === 'Move to Trash') {
+              Logger.debug('OneExplorer', `Delete '${path.parse(oneNode.node.path).base}'.`);
```
hyunsik-yoon, line 254: ditto

src/OneExplorer/OneExplorer.ts
hyunsik-yoon: may because of `modal: true` ?

dayo09: Please look into https://github.com/Samsung/ONE-vscode/pull/907#issuecomment-1171199706. The current delete command behavior looks alike the vscode-native.

src/OneExplorer/OneExplorer.ts
```diff
@@ -229,6 +229,37 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+              return vscode.workspace.fs.delete(
+                  oneNode.node.uri, {recursive: false, useTrash: true});
+            }
+          })
```
hyunsik-yoon, line 258: can we define a function to reduce code duplication?
for example,
```
function confirmDeletion(uri: URI, isDirectory) {
  let title = isDirectory? " ...  and its contents?" : " ..."';
  let recursive = isDirectory? true : false;
  // showinformation code below
  ...
}
```

dayo09: Actually, I tried it but eventually the current one looked better... but well, I will try once again.

src/OneExplorer/OneExplorer.ts
yunjayh: > maybe because of `modal: true` ?

This made the change! Thank you!!

src/OneExplorer/OneExplorer.ts
dayo09: @hyunsik-yoon PTAL,

src/OneExplorer/OneExplorer.ts
```diff
@@ -229,6 +229,39 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    if (oneNode.node.type === NodeType.directory) {
+      vscode.window.showInformationMessage(title, {detail: detail, modal: true}, approval)
+          .then(ans => {
+            if (ans === approval) {
+              Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
+              return vscode.workspace.fs.delete(
+                  oneNode.node.uri, {recursive: recursive, useTrash: true});
+            }
+          })
+          .then(() => this.refresh());
```
hyunsik-yoon, line 251: This is identical to L253 to L261.

how about changing the current code

```
if (oneNode.node.type === NodeType.directory) {
  vscode.window.showInformationMessage(...);
} else {
  vscode.window.showInformationMessage(...);  // same with the above showInformationMessage
}
```

to  
```
vscode.window.showInformationMessage(...);
```

dayo09: Cool, I missed it. Thanks :-D

hyunsik-yoon (approved):

yunjayh (approved):
LGTM

PR #908, opened 2022-06-30
[OneExplorer] Save tree root

dayo09:
This commit saves tree root.
When it needs to rebuild a part of the tree by 'refresh',
build the sub-tree under the given node.

----

From #738

COMMENT: This feature is required to implement `reveal` command and `open in vscode-explorer view` command. Currently, we cannot search through the root node as it doesn't keep it.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -168,7 +168,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
    */
   refresh(oneNode?: OneNode): void {
     if (!oneNode) {
-      // refresh the root
+      // Reset the root in order to build from scratch (at OneTreeDataProvider.getTree)
       this.tree = undefined;
     }
 
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -150,17 +150,30 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   private fileWatcher =
       vscode.workspace.createFileSystemWatcher(`**/*.{cfg,tflite,onnx,circle,tvn}`);
 
+  private tree: Node|undefined;
+
   constructor(private workspaceRoot: vscode.Uri) {
     const fileWatchersEvents =
         [this.fileWatcher.onDidCreate, this.fileWatcher.onDidChange, this.fileWatcher.onDidDelete];
 
     for (let event of fileWatchersEvents) {
-      event(() => this._onDidChangeTreeData.fire());
+      event(() => this.refresh());
     }
   }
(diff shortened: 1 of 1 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -150,17 +150,30 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
 
+  private tree: Node|undefined;
+
   constructor(private workspaceRoot: vscode.Uri) {
     const fileWatchersEvents =
         [this.fileWatcher.onDidCreate, this.fileWatcher.onDidChange, this.fileWatcher.onDidDelete];
 
     for (let event of fileWatchersEvents) {
-      event(() => this._onDidChangeTreeData.fire());
+      event(() => this.refresh());
```
dayo09, line 160: This change is required. 

FYI, what `_onDidChangeTreeData.fire()` does:
- it's an event api given by vscode. (vscode.TreeViewProvider)
- it calls `getTree()`, which will remain return the same tree if we don't reset `this.tree`.
- NOW, `this.refresh()` does what we want: reset `this.tree` and fires

src/OneExplorer/OneExplorer.ts
```diff
@@ -150,17 +150,30 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-    this._onDidChangeTreeData.fire();
+  /**
+   * Refresh the tree under the given oneNode
+   * @param oneNode A start node to rebuild. The sub-tree under the node will be rebuilt.
+   *                If not given, the whole tree will be rebuilt.
+   */
+  refresh(oneNode: OneNode|void): void {
+    if (!oneNode) {
+      // refresh the root
+      this.tree = undefined;
```
YongseopKim, line 172: Is it right?
- oneNode -> undefined
- tree -> undefined
- oneNode -> undefeind

Is it your intention?

hyunsik-yoon: ref: https://github.com/Microsoft/TypeScript/issues/25481#issuecomment-403082159

how about `refresh(oneNode?: OneNode): void {` ?

dayo09: > Is it your intention?

Yes :-D

YongseopKim: IHMO, if so,

```ts
refresh(oneNode: OneNode|undefined): void {
// ...
if (oneNode === undefined) { // if you want to clarify more 
  this.tree = undefined;
  // and I don't think it's necessary that oneNode = undefined;
}
```

But, it's optional as always. 😃

hyunsik-yoon (approved):

YongseopKim (approved):

PR #911, opened 2022-07-01
[PartEditor] Show filename as panel title

seanshpark:
This will revise panel title to show filename instead.


mhs4670go (approved):
LGTM

PR #912, opened 2022-07-01
[PartEditor] Revise update without backend

seanshpark:
This will revise update from PartEditor to PartGraphSelector to remove
backend as it is not used anymore.


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

PR #914, opened 2022-07-01
[CircleGraph] Add graph node style for backend assignments

seanshpark:
This will add graph node style for backend assignmments with
element generation as color is decided from backend extensions.


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

PR #917, opened 2022-07-01
[PartEditor] Do not check _document

seanshpark:
_document is now always valid object so there is no need to check.


seanshpark:
for #822 from draft #851

mhs4670go (approved):
LGTM

PR #920, opened 2022-07-03
[PartGraphSelector] Set node backend assignment after load

seanshpark:
This will set node backend assignment after load is finished.


seanshpark:
for #822 from draft #851

GrapView will show color nodes with this change

image

mhs4670go (approved):
LGTM

PR #924, opened 2022-07-04
[PartEditor] Revise select element usage

seanshpark:
This will revise select element usage.
- use `add()` not `options.add()`
- use `selectedOptions`


seanshpark:
Refernce
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/add
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions

mhs4670go (approved):
LGTM

PR #925, opened 2022-07-04
[CircleGraph] Scroll to selected nodes with flag

seanshpark:
This will introduce '_scrollToSelected' flag member as scroll to selected
nodes when selection has changed.


mhs4670go (approved):
LGTM

PR #927, opened 2022-07-04
[CircleGraph] Revise setSelection graph traverse

seanshpark:
This will revise setSelection method graph traverse to use nodes.keys()
for using forEach as this method is used in existing codes.


mhs4670go (approved):
LGTM

PR #934, opened 2022-07-04
[ToolRunner] Use Promise.prototype.catch instead of try-catch

llFreetimell:
To correctly test `Promise`, we should use `await` or `Promise.prototype.*`.
However, since other testcases are already using `Promise.prototype.*`,
let's use it.

### DIFF
--- a/src/Tests/Project/ToolRunner.test.ts
+++ b/src/Tests/Project/ToolRunner.test.ts
@@ -104,12 +104,10 @@ suite('Project', function() {
         for (let i = 0; i < simultaneousTrials; i++) {
           let args = new ToolArgs();
           args.push('1');  // 1 sec
-          try {
-            toolRunner.getRunner('sleep', 'sleep', args, '.');
-          } catch (err) {
-            // cannot make process
+          const runner = toolRunner.getRunner('sleep', 'sleep', args, '.');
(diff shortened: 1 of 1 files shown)


src/Tests/Project/ToolRunner.test.ts
```diff
@@ -104,12 +104,10 @@ suite('Project', function() {
         for (let i = 0; i < simultaneousTrials; i++) {
           let args = new ToolArgs();
           args.push('1');  // 1 sec
-          try {
-            toolRunner.getRunner('sleep', 'sleep', args, '.');
```
llFreetimell, line 108: :bell: **Details** :bell: 

`await` or `Promise.prototype.catch` should be used

hyunsik-yoon: This err is raised before returning promise.

https://github.com/Samsung/ONE-vscode/blob/c861c744a178975d07742a67c76624bb0e781335/src/Project/ToolRunner.ts#L160-L170

llFreetimell: Error raising is expected and that's what this test want to check.
My intention was catch that error in `catch()` but not caught...

llFreetimell:
Close this because this warning is not caught easily and related codes may be changed in the near future :(

PR #816, opened 2022-06-21
[draft] DESIGN DEVICE API

YongseopKim:
ONE-vscode-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

PR #935, opened 2022-07-04
[DeviceAPI] Add Device API on Execute

struss:
This commit will add `Device` API on Execute

### DIFF
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -20,7 +20,7 @@ import {DeviceSpec} from '../Backend/Spec';
  * `Device` class
  * a class to handle and access Device.
  * Each Device has single `DeviceSpec`, which could be same with other device.
- * Eash Device has a name to specify with other device with same `DeviceSpec`.
+ * Each Device has a name to specify with other device with same `DeviceSpec`.
  *
  * for example, Device A, B, C connected on, and A and B are TizenTV and C is HostPC.
  * then we could access this as below
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -16,12 +16,20 @@
 
 import {DeviceSpec} from '../Backend/Spec';
 
-// What is `RealDeviceInstance` class?
-// Suppose a case that some tizen tvs are. dtv0, dtv1, ... dtvn_1
-// the dtv's spec can be just only one
-// However, each dtv should be handled one by one
-// At the time, this class expresses an instance for a real device
-class DeviceInstance {
+/**
+ * `Device` class
+ * a class to handle and access Device.
+ * Each Device has single `DeviceSpec`, which could be same with other device.
+ * Eash Device has a name to specify with other device with same `DeviceSpec`.
+ *
+ * for example, Device A, B, C connected on, and A and B are TizenTV and C is HostPC.
+ * then we could access this as below
+ *
+ * const deviceA = new Device("A", TizenDeviceSpec);
+ * const deviceB = new Device("B", TizenDeviceSpec);
+ * const deviceC = new Device("C", HostPCSpec);
+ */
+class Device {
   name: string;
   spec: DeviceSpec;
   constructor(name: string, spec: DeviceSpec) {
@@ -30,4 +38,4 @@ class DeviceInstance {
   }
 };
 
-export {DeviceInstance};
+export {Device};
--- a/src/Execute/DeviceInstance.ts
+++ b/src/Execute/DeviceInstance.ts
@@ -21,7 +21,7 @@ import {DeviceSpec} from '../Backend/Spec';
 // the dtv's spec can be just only one
 // However, each dtv should be handled one by one
 // At the time, this class expresses an instance for a real device
-class RealDeviceInstance {
+class DeviceInstance {
   name: string;
   spec: DeviceSpec;
   constructor(name: string, spec: DeviceSpec) {
@@ -30,4 +30,4 @@ class RealDeviceInstance {
   }
 };
 
-export {RealDeviceInstance};
+export {DeviceInstance};
--- a/src/Execute/RealDeviceInstance.ts
+++ b/src/Execute/RealDeviceInstance.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


hyunsik-yoon:
I guess that draft for this PR is https://github.com/Samsung/ONE-vscode/pull/816

src/Execute/RealDeviceInstance.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import {DeviceSpec} from '../Backend/Spec';
+
+// What is `RealDeviceInstance` class?
+// Suppose a case that some tizen tvs are. dtv0, dtv1, ... dtvn_1
+// the dtv's spec can be just only one
+// However, each dtv should be handled one by one
+// At the time, this class expresses an instance for a real device
+class RealDeviceInstance {
```
hyunsik-yoon, line 24: I am not sure the meaning of `Real` and `Insatance` here. Can we just call this `Device`?
Thinking about `Real`, will there be, e.g., `VirtualDeviceInstance` later? I wonder what other `DeviceInstance` will be in future. Could you share some example?
Similarly, thinking about `Instance`, will there be, e.g., `RealDeviceClass` later?

YongseopKim: It makes sense. However, `Device` can be ambiguous. The situation can be explained as comments.

```
// Suppose a case that some tizen tvs are. dtv0, dtv1, ... dtvn_1
// the dtv's spec can be just only one
// However, each dtv should be handled one by one
// At the time, this class expresses an instance for a real device
```

I can agree on excepting `Real` but I think `Instance` is necessary.

struss: I also agree about `Real` part. I think `Real` means nothing about this. I will remove that expression. 
but about `Instance` I think remaining this could be more nice. I will change to `DeviceInstance` would look more nice.

struss: > Similarly, thinking about `Instance`, will there be, e.g., `RealDeviceClass` later?

No, `DeviceClass` will not implemented. I think `DeviceInstance` could classified by `DeviceSpec`.

src/Execute/DeviceInstance.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {DeviceSpec} from '../Backend/Spec';
+
+// What is `RealDeviceInstance` class?
```
yunjayh, line 19:
```suggestion
// What is `DeviceInstance` class?
```

src/Execute/DeviceInstance.ts
```diff
@@ -0,0 +1,33 @@
...
+// Suppose a case that some tizen tvs are. dtv0, dtv1, ... dtvn_1
+// the dtv's spec can be just only one
+// However, each dtv should be handled one by one
+// At the time, this class expresses an instance for a real device
```
yunjayh, line 23:
```suggestion
// Suppose the case of multiple tizen tvs - dtv0, dtv1, ... dtvn_1
// Every dtv instances shares one device spec, but should be handled separately
// This class expresses an instance for each of such real devices
```

(Just suggestion) When I read the comment, it took time to understand what this `DeviceInstance` implemented for. I'm not good at writing down the explanation, but could you please consider about update the comment a little bit? (I'm not confident to the suggested version too.)

src/Execute/RealDeviceInstance.ts
hyunsik-yoon: I am still wonder why `Instance` is needed in naming. So, to me it's kinda hard to understand what this class is for... Could you explain why it is _necessary_ and why it makes code _more nice_? 

IMHO, 
When a class is instantiated,  for example, `let a = new Foo()`.
everyone know `a` is an instance of `Foo`. So we don't normally use `FooInstance` as class name. 

By naming `FooInstance` (not `Foo`),
for me, I assume that an instance of class `Foo` is mapped with one instance of `FooInstance`...
It's kinda complex to understand.
Instead of creating `FooInstance`, can we move methods and vars of `FooInstance` into `Foo` and remove `FooInstance`?

I guess maybe I am missing something here... More explanation would help me to understand why tthis class is needed.

YongseopKim: Okay.
agree on `instance`: +1 (@YongseopKim)
disagree on `instance`: +1 (@hyunsik-yoon)

Now it depends on you, @struss. If you disagree on `instance` and others would be okay, I'll be okay too.

struss: um... I use `Instance` to emphasize A single Device with `DeviceSpec`. but follow from @hyunsik-yoon opinion, make class name with `Instance` will make redundant meaning. 

so, I will add one point to @hyunsik-yoon 's opinion.

src/Execute/DeviceInstance.ts
struss: I will revise this comment with change class name `Device`. I think class name could make more people hard to understand this concept, as this contains duplicated meaning.

src/Execute/Device.ts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import {DeviceSpec} from '../Backend/Spec';
+
+/**
+ * `Device` class
+ * a class to handle and access Device.
+ * Each Device has single `DeviceSpec`, which could be same with other device.
+ * Eash Device has a name to specify with other device with same `DeviceSpec`.
```
hyunsik-yoon, line 23:
```suggestion
 * Each Device has a name to specify with other device with same `DeviceSpec`.
```

hyunsik-yoon (approved):
Thanks. :-)

YongseopKim (approved):
LGTM!
(Could you change the title?)

struss:
@yunjayh PTAL

seanshpark:
FYI https://github.com/Samsung/ONE-vscode/pull/935/commits

commit still has `Instance` in title

src/Execute/Device.ts
```diff
@@ -0,0 +1,41 @@
...
+ * a class to handle and access Device.
+ * Each Device has single `DeviceSpec`, which could be same with other device.
+ * Each Device has a name to specify with other device with same `DeviceSpec`.
+ *
+ * for example, Device A, B, C connected on, and A and B are TizenTV and C is HostPC.
+ * then we could access this as below
+ *
+ * const deviceA = new Device("A", TizenDeviceSpec);
+ * const deviceB = new Device("B", TizenDeviceSpec);
+ * const deviceC = new Device("C", HostPCSpec);
```
yunjayh, line 30: :+1:

yunjayh (dismissed):

hyunsik-yoon (approved):

yunjayh (approved):

YongseopKim (approved):

YongseopKim:
@llFreetimell

Issue #928, opened 2022-07-04
[CircleGraph] Fix node color for backend assignment


seanshpark:
Backend assignment color for graph nodes are fixed for fast end-to-end development.
Let's fix this to follow values from partition editor extension.

seanshpark:
from #822

seanshpark:
done

PR #940, opened 2022-07-04
DRAFT node color from extension

seanshpark:
on-going draft to get graph node color from extension.


seanshpark:
for #928

seanshpark:
done

PR #943, opened 2022-07-05
[CircleGraphCtrl] Introduce pageloaded event

seanshpark:
This will introduce 'pageloaded' eventh when page is loaded.


seanshpark:
for #928, from #940

PR #944, opened 2022-07-05
[PartEditor] Send backends with color as object

seanshpark:
This will revise resultBackends to send backend name and color in one object.


seanshpark:
for #928, from #940

mhs4670go (approved):
LGTM

PR #945, opened 2022-07-05
[CircleGraph] Backend colors from extension

seanshpark:
This will revise extension to send backend colors to web view.


seanshpark:
for #928, from #940

mhs4670go (approved):
LGTM

Issue #923, opened 2022-07-03
[PartEditor] Provide search input to filter operators


seanshpark:
Let's provide a search editbox to filter operators with name and opcode.

seanshpark:
from #822

seanshpark:
done

PR #948, opened 2022-07-05
[PartEditor] input for find operators

seanshpark:
This will add input element for find operators by name and code.


seanshpark:
for #923, from #942

mhs4670go (approved):
LGTM

PR #949, opened 2022-07-05
[PartEditor] Find operator with input

seanshpark:
This will enable find operator with input text for name and opcode.


seanshpark:
for #923, from #942

mhs4670go (approved):
LGTM

Issue #950, opened 2022-07-05
[ToolchainEnv] Enable even workspace is not set yet


yunjayh:
### What

Let's enable many functions related to `toolchainEnv` even if the vscode workspace is not set yet.

### Why

`toolchainEnv` is independent to the workspace but only related with the execution environment.

yunjayh:
AFAIU, the `WorkFlow` takes a workspace as its member variable, and execute the job commands with setting the `workspace`. as its `cwd`.

However, `Toolchain` related commands doesn't related with the working directory. So, let's set it as home directory.

hyunsik-yoon:
> Let's enable many functions related to toolchainEnv even if the vscode workspace is not set yet.

When you meant by `many functions`, can you list those functions?

yunjayh:
Following methods of `ToolchainEnv` class cannot be called if workspace is not defined.

  * `ToolchainEnv.execute` 
  * `ToolchainEnv.request`
  * `ToolchainEnv.prerequisites`
  * `ToolchainEnv.install`
  * `ToolchainEnv.uninstall`
  * `ToolchainEnv.compiler`

But all of the above methods are able to run independent to current working directory. That's why I fired this issue.

hyunsik-yoon:
I guess that `ToolchainEnv.compile()` can be called when OneExplorerView is available ('available' means a case when workspace is set)

jyoungyun:
I have confirmed that the above functions work without workspace through simple modification. Did you check the draft PR I suggested?

yunjayh:
Thanks to @jyoungyun, #1002 solved this issue! :bow:

Issue #932, opened 2022-07-04
Provide welcome page to fresh user


yunjayh:
Currently, the user who installed our `ONE-vscode` extension cannot get any information.

Since we have a `ONE view container`, it can be a great place to give a welcome message and a brief guide. 

I'm going to refer to [this link](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome).

yunjayh:
😮 I totally forgot about it. Thank you for your information..!!

hyunsik-yoon:
@yunjayh Note that #348 code was revoked for some reason. So we don't have the code right now in main branch. :-)

yunjayh:
**[The texts are work in progress]** IMHO with this way, user can get what to do after installing our extension, or opening new window without workspace opened.
![image](https://user-images.githubusercontent.com/50489513/177274738-12818230-4501-46c7-aafd-3c41afaebd72.png)

_Originally posted by @yunjayh in https://github.com/Samsung/ONE-vscode/issues/953#issuecomment-1174727869_

jyoungyun:
Can you use this PR, #972 ?

yunjayh:
> I don't understand why we need `Register Installed Toolchain` button. If the system has toolchain, it will search and show it automatically. Is there any reason you defined this?

The contents are mostly copied from #348. I agree with your thought.

Sure for #972 :smile:

jyoungyun:
> The contents are mostly copied from https://github.com/Samsung/ONE-vscode/pull/348. I agree with your thought.

At that time, we though that the function would be necessary because the actual implementation was not implemented. But for now, I don't think it's necessary because it's automatically searched installed toolchain.

I didn't consider the welcome view in OneExplorerView. However, if you comment out the error throw code in OneExplorerView, you can show the ToolchainView working.

yunjayh:
> you can show the ToolchainView working even though workspace is not set.

The codes for showing `ToolchainView` without workspace is merged thanks to @jyoungyun, but because of timing issue, sometimes `ToolchainView` doesn't show anything even if backend is registered and toolchain has been installed. 
The best way to resolve the issue is find way to remove timing issue. For temporal solution, let's give a guide to user to refresh their toolchain view.

hyunsik-yoon:
>  For temporal solution, let's give a guide to user to refresh their toolchain view.

IMHO 
If solving timing issue won't take too long,
[...]

PR #951, opened 2022-07-05
[Toolchain] Change toolchain workspace as `homedir`

yunjayh:
This commit changes toolchain workspace as `hoemdir`. Since `toolchain` related
jobs are independent from vscode-workspace, it's able to set the command's workspace
as user `homedir`.

---

for: #950

### DIFF
--- a/src/Toolchain/ToolchainEnv.ts
+++ b/src/Toolchain/ToolchainEnv.ts
@@ -33,7 +33,7 @@ import {showPasswordQuickInput} from '../View/PasswordQuickInput';
 class Env implements BuilderJob {
   logTag = 'Env';
   workFlow: WorkFlow;  // our build WorkFlow
-  currentWorkspace: string = '';
+  workingDirectory: string = '';
   isPrepared: boolean = false;
 
   constructor() {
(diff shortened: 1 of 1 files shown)


yunjayh:
Even though this pr is landed, it's blocked to enable `one-vscode` extension with *NO workspace opened* status. It will be done while #932.

src/Toolchain/ToolchainEnv.ts
```diff
@@ -33,7 +33,7 @@ import {showPasswordQuickInput} from '../View/PasswordQuickInput';
 class Env implements BuilderJob {
   logTag = 'Env';
   workFlow: WorkFlow;  // our build WorkFlow
-  currentWorkspace: string = '';
+  currentWorkspace: string = require('os').homedir();
```
yunjayh, line 36: Since `workspace` is defined in vscode, it can make confusion to others. Let's change this variable name to `workingDirectory`

hyunsik-yoon:
I am still not sure why home dir is used for cwd...

From https://nodejs.org/api/child_process.html#child_processspawncommand-args-options
```
Use cwd to specify the working directory from which the process is spawned. If not given, 
the default is to inherit the current working directory. 
```

yunjayh:
(From offline talked with @hyunsik-yoon) The current codes makes reader or developer confused about why cwd is defined as home directory. 

I changed this to home dir since cws is **DONT CARE** parameter for toolchain related child processes. However it can make misunderstanding such as *it should be home dir* or *there is some reason for this home dir*. 

To prevent this, there are two options.
  * Add comment to explain this situation
  * Change workspace as optional in `workflow.start` or something

Since second option will need more effort, I'll add comment about this discussion and add todo to update workflow module.

jyoungyun (approved):
LGTM

yunjayh:
#1002 substitute this pr.

PR #953, opened 2022-07-05
[Draft] Provide welcome guide via ONE view container

yunjayh:
[WIP]

This commit provides welcome guide via ONE view container.

---

for #932

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,16 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+		"viewsWelcome": [
+      {
+        "view": "OneExplorerView",
+        "contents": "Open folder which includes models and configuration files.\n[Open Folder](Command:workbench.action.files.openFolder)\nFollowing types of models will be displayed.\n- *.tflite, *pb and *.onnx\n- Additional file types which are described by backend.\n\nTo learn more about ONE-vscode extension usage [read our docs](https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md)."
+      },
+      {
+        "view": "ToolchainView",
+        "contents": "Register already installed toolchain or install a new toolchain.\n[Register Toolchain](Command:one.toolchain.refresh)\n\n[Install Toolchain](Command:one.toolchain.install)"
+      }
+    ],
     "viewsContainers": {
       "activitybar": [
         {


yunjayh:
**[The texts are work in progress]** IMHO with this way, user can get what to do after installing our extension, or opening new window without workspace opened.
image

package.json
```diff
@@ -16,6 +16,16 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+		"viewsWelcome": [
+      {
+        "view": "OneExplorerView",
+        "contents": "Open folder where model files are stored. Models files are one of the following format:\n- *.tflite, *.pb, Keras model or saved model for Tensorflow​\n- *.onnx or *.pth for PyTorch\n[Open Folder](Command:onevscode.fileTreeView-open)\n"
```
dayo09, line 22:
```suggestion
        "contents": "Open folder where model files are stored.\nModel files are one of the following format:\n- *.tflite, *.pb, Keras model or saved model for Tensorflow​\n- *.onnx or *.pth for PyTorch\n[Open Folder](Command:onevscode.fileTreeView-open)\n"
```

yunjayh: Does `Keras model` or `saved model` and `.pth` are visible in one-explorer? I read the codes, but I'm not sure about it.
/cc @dayo09

yunjayh:
image

src/OneExplorer/OneExplorer.ts
```diff
@@ -501,10 +501,21 @@ input_path=${modelName}.${extName}
...
 
   constructor(context: vscode.ExtensionContext) {
+    try {
+      this.workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+    } catch (e: unknown) {
+      let errmsg = 'Failed to obtain workspace root';
+      if (e instanceof Error) {
+        errmsg = e.message;
+      }
+      Logger.debug('OneExplorer', errmsg);
```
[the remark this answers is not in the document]
yunjayh: Yes I intended it. I wrote that code when I was trying to test how to show welcome view in one-explorer view. Now, I'm working on how to treat *no workspace error* in proper way in my local draft. Anyway, thank you for your advice!!

yunjayh:
@dayo09 Could you please take a look at `OneExplorer.ts`?
I modified it not to throw error and not to pop-up the error message balloon. Is there any threat that I didn't consider or expect?

AFAIK, the extension's `activation` function will be executed in every vscode window start up ([source](https://code.visualstudio.com/api/references/activation-events#onStartupFinished)). 

If vscode cannot obtain the workspace, `OneTreeDataProvider` returns empty children. It leads to show welcome view which defined at `package.json`.
Else, if workspace is defined by user, `OneExplorerView` works well as expected.

PR #961, opened 2022-07-05
[CfgEditor] Revisit CfgEditorPanel class

YongseopKim:
Let's revisit CfgEditorPanel class.

### DIFF
--- a/src/CfgEditor/CfgData.ts
+++ b/src/CfgEditor/CfgData.ts
@@ -21,6 +21,19 @@ const sections = [
   'one-optimize', 'one-quantize', 'one-codegen', 'one-profile'
 ];
 
+// NOTE: Why is not function overloadding used? Its maintain costs expensive.
+// Compared to C++, TS supports function overloading very compilcated like
+// class TSClass {
+//   // declare function signatures to overload
+//   set(sth: A);
+//   set(sth: B);
+//   // implementation is the only one
+//   set(sth: unknown) {
+//     if (sth isInstanceOf(A)) { ... }
+//     else if (sth isInstanceOf(B)) { ... }
+//   }
+// }
+//
 export class CfgData {
   private _oneConfig: any = undefined;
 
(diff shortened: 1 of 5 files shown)


YongseopKim:
TODO
- [ ] Separate thing about WebView(+HTML)
- [ ] TEST

YongseopKim:
This is doing refactoring for CfgEditorPanel 
- Separate data class and webview builder class
- Now the only data class has own config instance
- Now making html for webview is separated

YongseopKim:
PTAL @llFreetimell and @Samsung/one-vscode

YongseopKim:
Now is it possible?

src/CfgEditor/CfgHtmlBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+export class CfgHtmlBuilder {
+  webview: vscode.Webview;
+  extensionUri: vscode.Uri;
+
+  constructor(webview: vscode.Webview, extensionUri: vscode.Uri) {
+    this.webview = webview;
+    this.extensionUri = extensionUri;
+  }
+
+  async build(): Promise<string> {
```
llFreetimell, line 31: This function is changed to return `Promise<string>`, not `string`.
But I cannot find anywhere to use `Promise`.
Is there any reason to return `Promise` even it is not used?

src/CfgEditor/CfgData.ts
```diff
@@ -0,0 +1,111 @@
...
+            this._oneConfig[sectionName][paramName] === param) {
+          continue;
+        }
+        return false;
+      }
+    }
+    return true;
+  }
+
+  getSorted(): CfgData {
```
llFreetimell, line 99: Function name is `getSorted` but it is doing both `get` and `set`.
So I want to suggest following idea.

```ts
  sort(): void {
    ...
  }

  const cfgData = CfgData();
  cfgData.sort();
  let someVar = cfgData.getOneConfg();
```

YongseopKim: Oh, that's a good point! I'll fix it soon.

src/CfgEditor/CfgHtmlBuilder.ts
YongseopKim: It's because this function is `async`. And the reason that uses `async` is using `await this.getHtml()`.

src/CfgEditor/CfgData.ts
YongseopKim: done

llFreetimell:
Sorry for bothering but please rebase with latest commit!

src/CfgEditor/CfgData.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  getStringfied(): string {
+    return ini.stringify(this._oneConfig);
+  }
+
+  updateWithParsedConfig(cfg: any): void {
+    this._oneConfig = cfg;
+  }
+
+  updateWithStringifiedText(text: string): void {
```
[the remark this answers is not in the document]
YongseopKim: great idea!

YongseopKim:
PTAL @llFreetimell and @Samsung/one-vscode

llFreetimell (approved):
LGTM

jyoungyun (approved):
LGTM

PR #963, opened 2022-07-06
Test PR for PR/NO Merge

yunjayh:
Since this commit doesn't have signed-off message, the test should be broken.
However, adding `PR/NO TEST` label should prevent the test running.

---

06 Jul 22 10:16 -> create pr without any label to check action is working
11 Jul 22 12:44 -> removed `check-label` job which remarks as failure if `PR/NO TEST` is set

### DIFF
--- a/src/Config/README.md
+++ b/src/Config/README.md
@@ -5,6 +5,8 @@ This extension adds a GUI configuration file setting for ONE developers. This in
 - import cfg file and edit it
 - export cfg file
 
+🤧 
+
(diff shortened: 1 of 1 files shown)


yunjayh:
[Without `PR/NO TEST` label] The workflow run successfully (which doesn't mean it succeed)
image

llFreetimell:
@yunjayh 

As the first purpose of this feature is to disable tests in draft PR. I think remarking failure is not needed :)
Inversely, remarking failure seems "test failed" even the label is "no test"

PR #965, opened 2022-07-06
[PartEditor] Introduce types for commands

seanshpark:
This will introduce types for commands to make command and function call simple.


seanshpark:
from draft #964

mhs4670go (approved):
LGTM

Issue #962, opened 2022-07-06
[PartEditor] UX enhancement/fixes no.2


seanshpark:
Second UX enhancement/fixes of PartEditor todo.

seanshpark:
- [ ] fix right click on GraphView

user intension is to change backend...

(1) change context menu
- show context menu with `Set backend to CPU`, `Set backend to ACL_CL`, ...

(2) disable context menu on right click
- ?

seanshpark:
Add: Graph edit view popup menu position
- when button on the toolbar is clicked, menu shows up on top of the button
- which first item can easly be clicked
- it would be better to move the popup menu to show on bottom of the button

seanshpark:
done for now

PR #970, opened 2022-07-06
[PartEdit] Revise find with opcode name

seanshpark:
This will revise to find opcode name as shown in listbox.


seanshpark:
for #962

PR #972, opened 2022-07-06
[Toolchain] Support welcome view

jyoungyun:
This commit do not throw an error when workspace is not defined.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -495,12 +495,17 @@ input_path=${modelName}.${extName}
 
 export class OneExplorer {
   // TODO Support multi-root workspace
-  public workspaceRoot: vscode.Uri = vscode.Uri.file(obtainWorkspaceRoot());
+  public workspaceRoot: vscode.Uri;
   public treeView: vscode.TreeView<OneNode|undefined>|undefined;
 
   constructor(context: vscode.ExtensionContext) {
+    let workspaceRoot = obtainWorkspaceRoot();
+    if (!workspaceRoot) {
+      throw new Error('Need workspace');
+    }
+    this.workspaceRoot = vscode.Uri.file(workspaceRoot);
     // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
-    const oneTreeDataProvider = new OneTreeDataProvider(this.workspaceRoot!);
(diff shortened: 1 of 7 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -495,12 +495,17 @@ input_path=${modelName}.${extName}
...
-  public workspaceRoot: vscode.Uri = vscode.Uri.file(obtainWorkspaceRoot());
+  public workspaceRoot: vscode.Uri;
   public treeView: vscode.TreeView<OneNode|undefined>|undefined;
 
   constructor(context: vscode.ExtensionContext) {
+    let workspaceRoot = obtainWorkspaceRoot();
+    if (!workspaceRoot) {
+      throw new Error('Need workspace');
+    }
+    this.workspaceRoot = vscode.Uri.file(workspaceRoot);
```
yunjayh, line 506: I pulled this pr and run as debug mode. Did it activate successfully? It failed to activate the extension because `new OneExplorer(context)` in extension.ts threw an error.

IMHO it is needed to add error handler in `activate` function in `extension.ts`.

```ts
export function activate(context: vscode.ExtensionContext) {
...
  // as-is
  new OneExplorer(context);

  // to-be
  try {
    new OneExplorer(context);
  } catch (e: unknown) {
    let errmsg = 'Failed to obtain workspace root';
    if (e instanceof Error) {
      errmsg = e.message;
    }
    Logger.debug('OneExplorer', 'Workspace is unset yet.');
  }
...
```

yunjayh (commented):
May I ask what's your intention in https://github.com/Samsung/ONE-vscode/issues/932#issuecomment-1176029309? When I read the comment, I understood as you are going to work with it. Did I get correctly?

src/OneExplorer/OneExplorer.ts
jyoungyun: I tried to show that the ToolchainView works even without workspace. Now this code still fails without workspace because the OneExplorer throws an error if there is no workspace. If you comment out L504 or catch an error related to OneExplorer, you can see that ToolchainView is working without workspace.

jyoungyun: I expect you to be able to show the Welcome View when OneExplorer does not have workspace instead of throwing error.

PR #973, opened 2022-07-06
[CfgEditor] Integrate multiple js into one

llFreetimell:
There were multiple js files in `media` for maintenance.
However, it causes unintended error in remote-ssh environment.
To resolve this, and for better performance,
multiple js files should be integrated as one.


llFreetimell:
#### For reviewers

The code and logic itself is not changed at all.
Just four files are integrated to one `index.js`.

hyunsik-yoon (approved):

hyunsik-yoon (approved):

seanshpark:
> it causes unintended error in remote-ssh environment.

I think this error is that current web view doesn't show initial .cfg content, right?
If, so, I think the issue is about timing between (1) web view takes time to load (2) extension sends message 'displayCfgToEditor' to load the cfg file content.
IMHO, (2) should happen after (1) is finished. (current code seems to send `displayCfgToEditor` when extension is activated)

One solution can be, (step-a) at the end of (1), web view send request message to extension (step-b) extension send cfg file content when message from (step-a) is received.

> and for better performance,

This is what I got with 0.2.0 release version from my test machine. do you have any results before and after?

image

YongseopKim (approved):
LGTM

llFreetimell:
@seanshpark 
Really appreciate for your perfect solution...! I didn't know neither the `network` tab nor detailed cause but all resolved :)
And the total loading time was decreased from 219ms(89ms+130ms) to 147ms but it is no more meaningful :D

---

I will close this and create new PR for applying perfect solution!

PR #976, opened 2022-07-07
[Toolchain] Fix SonarLint related issues

jyoungyun:
This commit fixes the problem pointed out in SonarLint.

### DIFF
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -17,8 +17,7 @@
 import * as vscode from 'vscode';
 
 import {Toolchain} from '../Backend/Toolchain';
-import {DebianToolchain} from '../Backend/ToolchainImpl/DebianToolchain';
-import {Job, JobCallback} from '../Project/Job';
+import {Job} from '../Project/Job';
 import {JobInstall} from '../Project/JobInstall';
 import {JobUninstall} from '../Project/JobUninstall';
 import {Logger} from '../Utils/Logger';
@@ -30,6 +29,8 @@ enum NodeType {
   backend,
   toolchain,
 }
+
+type ToolchainTreeData = ToolchainNode|undefined|void;
 
 export class ToolchainNode extends vscode.TreeItem {
   constructor(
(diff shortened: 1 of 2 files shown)


YongseopKim (approved):
LGTM

src/Toolchain/ToolchainProvider.ts
```diff
@@ -161,7 +159,7 @@ export class ToolchainProvider implements vscode.TreeDataProvider<ToolchainNode>
     });
   }
 
-  run(cfg: string) {
+  run(_cfg: string) {
```
hyunsik-yoon, line 162: Could you let us know why this change is needed? Thanks in advance.

jyoungyun: This changes came from SonarLint. If this variable is not used in this function, SonarLint suggests removing this variable or adding `_` prefix before the word.

SonarLint warning:
> Remove the unused function parameter "cfg" or rename it to "_cfg" to make intention explicit.sonarlint(typescript:S1172)

src/Toolchain/ToolchainProvider.ts
```diff
@@ -60,12 +61,9 @@ export class ToolchainNode extends vscode.TreeItem {
 export class ToolchainProvider implements vscode.TreeDataProvider<ToolchainNode> {
   tag = this.constructor.name;  // logging tag
 
-  private _onDidChangeTreeData: vscode.EventEmitter<ToolchainNode|undefined|void> =
-      new vscode.EventEmitter<ToolchainNode|undefined|void>();
```
jyoungyun, line 64: ToolchainNode|undefined|void

> Replace this union type with a type alias. [+2 locations]sonarlint(typescript:S4323)

<details>
<summary> SonarLint Rule Description </summary>

```
Type aliases should be used (typescript:S4323)
CODE_SMELL  Code Smell  MINOR  Minor
Union and intersection types are convenient but can make code harder to read and maintain. So if a particular union or intersection is used in multiple places, the use of a type alias is recommended.

Noncompliant Code Example
function foo(x:string|null|number) { // Noncompliant
  // ...
}
function bar(x:string|null|number) {
  // ...
}
function zoo(): string|null|number {
  return null;
}
Compliant Solution
type MyType = string | null | number;

function foo(x: MyType) {
  // ...
}
function bar(x: MyType) {
  // ...
}
function zoo(): MyType {
  return null;
}
```

</details>

src/Toolchain/ToolchainProvider.ts
```diff
@@ -60,12 +61,9 @@ export class ToolchainNode extends vscode.TreeItem {
...
-  readonly onDidChangeTreeData?: vscode.Event<ToolchainNode|undefined|void> =
-      this._onDidChangeTreeData.event;
-
-  constructor() {}
```
jyoungyun, line 68: > Unexpected empty constructor.sonarlint(typescript:S1186)

hyunsik-yoon (approved):

PR #978, opened 2022-07-07
[OneExplorer] Introduce ArtifactLocator

dayo09:
This commit refactors ConfigObj class by introducing Artifact and ArtifactLocator

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -147,7 +147,7 @@ export class LocatorRunner {
    * A helper function to grep a filename ends with 'ext' within the given 'content' string.
    */
   public static searchWithExt = (ext: string, content: string): string[] => {
-    // Don't remove this. It's to prevent an 'content.cplit is not a function' error.
+    // Don't remove this. It's to prevent 'content.split is not a function' error.
     // TODO Find more straightforward way to resolve an error
     content = content + '';
 
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -20,6 +20,191 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 
 /**
+ * 'Artifact' means includes
+ *   - existing files to run ONE config (base model)
+ *   - result files after running ONE config (derived models, )
+ */
+interface Artifact {
+  /**
+   * An artifact's type
+   */
+  type: string;
+
+  /**
+   * A file extension
+   */
+  ext: string;
+
+  /**
+   * A full path in file system
+   */
+  path?: string
+}
+
+/**
+ * 'ArtifactLocator' is to find artifact's absolute path
+ */
+export class ArtifactLocator {
+  /**
+   * The section of ini to find the targetted file
+   * If not given, locator searches the whole section
+   */
+  section?: string;
+
+  /**
+   * The key inside section to find the targetted file
+   * If not given, locator searches the whole key
+   */
+  key?: string;
+
+  /**
+   * (optional)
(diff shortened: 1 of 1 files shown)


src/OneExplorer/ConfigObject.ts
```diff
@@ -19,6 +19,191 @@ import * as ini from 'ini';
...
+   */
+  artifact: Artifact;
+
+  /**
+   * A default search function
+   * @param content [this.section][this.key] value
+   * @returns string file name array
+   */
+  private searchEndsWithExt = (content: string): string[] => {
+    // Don't remove this. It's to prevent an 'content.cplit is not a function' error.
```
YongseopKim, line 80: Could you describe this error in detail? Hmm. The code doesn't seem to have the problem.

src/OneExplorer/ConfigObject.ts
```diff
@@ -19,6 +19,191 @@ import * as ini from 'ini';
 import * as path from 'path';
 import * as vscode from 'vscode';
 
+/**
+ * 'Artifact' means includes
```
YongseopKim, line 23: Is it a kind of group unit? logically? not physically?

src/OneExplorer/ConfigObject.ts
YongseopKim: I cannot reproduce this.

dayo09: If you remove the line and run `npm run unittest`, you can see the error.

BTW, there is a typo
'content.cplit' -> 'content.split'

src/OneExplorer/ConfigObject.ts
dayo09: Just a logical one :-D

src/OneExplorer/ConfigObject.ts
```diff
@@ -19,6 +19,178 @@ import * as ini from 'ini';
...
+    this.artifactLocators.push({
+      artifact: {type: 'circle', ext: '.circle'},
+      locator: new Locator((value: string) => LocatorRunner.searchWithExt('.circle', value))
+    });
+
+    this.artifactLocators.push({
+      artifact: {type: 'tvn', ext: '.tvn'},
+      locator: new Locator((value: string) => LocatorRunner.searchWithExt('.tvn', value))
+    });
+  }
```
dayo09, line 168:
```suggestion
    this.artifactLocators.push({
      artifact: {type: 'log', ext: '.circle.log'},
      locator: new Locator(
          (value: string) => {return LocatorRunner.searchWithExt('.circle', value)
                                  .map(filepath => filepath.replace('.circle', '.circle.log'))})
    });
  }
```
FOR EXAMPLE,
Above addition will support .circle.log

src/OneExplorer/ConfigObject.ts
YongseopKim: Very interesting!

I can reproduce
```
###True
###True
###/dummy/dummy/../../model.tflite
###/dummy/dummy/../../model.circle
###/model.circle
###/dummy/dummy/../..//model.q8.circle
###true
```

The contents' values are also `True` or `true`?

Anyway, the code seems okay on vscode npm watch. However, running unittests causes this error.

I guess
- On ts, the content is assumed as string type
- On js, content cannot be sure as string type because it is on js world
- the content is boolean type or functioin about boolean, I guess
- So, content(`not string type`) cannot have split function

I think that `content = content + ''` can be good option because it can make it sure the type is string.

YongseopKim (approved):
LGTM

dayo09:
@Samsung/one-vscode PTAL :-D

llFreetimell (approved):
LGTM

PR #981, opened 2022-07-07
[CirclrGraph] Reset model load data after loading

seanshpark:
This will reset model load data array after loading.


seanshpark:
from #980
- This may release memory used.

mhs4670go (approved):
LGTM

PR #982, opened 2022-07-07
[CircleGraph] Enable reload message

seanshpark:
This will enable reload message to reload the model after it may have
changed.


seanshpark:
from  #980

mhs4670go (approved):
LGTM

PR #983, opened 2022-07-07
[CircleGraphCtrl] Introduce reloadModel

seanshpark:
This will introduce reloadModel method to invoke view to reload the model.


seanshpark:
from  #980

mhs4670go (approved):
LGTM

Issue #929, opened 2022-07-04
[PartGraphSelector] Find dialog node selection


seanshpark:
Original click on find dialog nodes or edges selects that node or edge.
Current action is just to scroll to that item but actually it's not clear which item is selected as there is no graphical feedback.

Let's fix this by showing some feedback.
But,
- selecting on edges (arrows) are meaning less for us.

seanshpark:
while working with draft...
- find dialog is almost same as `Find` in `PartEditor`...
- maybe don't provide `find` feature for `selector` mode...

PR #989, opened 2022-07-07
DRAFT revise find dialog

seanshpark:
on-going draft to revise find dialog for better UX.


seanshpark:
for #929

PR #991, opened 2022-07-07
[PartEditor] Reset Operators listbox before fill

seanshpark:
This will reset Operators listbox before filling.


mhs4670go (approved):
LGTM

PR #993, opened 2022-07-07
[PartEditor] Reload model when changed

seanshpark:
This will enable to reload the model when changed from outside.


mhs4670go (approved):
LGTM

PR #854, opened 2022-06-27
[Toolchain] Run compile using default toolchain

jyoungyun:
This commit runs compile with cfg using default toolchain.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #1002, opened 2022-07-08
[Project] Introduce workDir in Job interface

jyoungyun:
This commit modifies the execution location of JobRunner to see the home
directory of os instead of seeing workspaceRoot.

Draft PR: #854 

**NOTE**
This modification helps each job run in a different location(cwd).

### DIFF
--- a/src/Project/JobCommand.ts
+++ b/src/Project/JobCommand.ts
@@ -29,6 +29,7 @@ class JobCommand implements Job {
   tool: string;
   toolArgs: ToolArgs;
   root: boolean;
+  workDir: string;
   successCallback?: JobCallback;
   failureCallback?: JobCallback;
 
@@ -37,6 +38,7 @@ class JobCommand implements Job {
     this.jobType = Job.Type.tUndefined;
     this.name = '';
     this.valid = false;
+    this.workDir = require('os').homedir();
 
     // init by cmd
     assert(cmd.length > 0);
--- a/src/Project/JobRunner.ts
+++ b/src/Project/JobRunner.ts
@@ -18,6 +18,7 @@ import {EventEmitter} from 'events';
 import * as vscode from 'vscode';
 
 import {Balloon} from '../Utils/Balloon';
+import * as helpers from '../Utils/Helpers';
 import {Logger} from '../Utils/Logger';
 
 import {Job} from './Job';
@@ -31,7 +32,6 @@ const K_CLEANUP: string = 'cleanup';
 export class JobRunner extends EventEmitter {
   tag = this.constructor.name;  // logging tag
   jobs: WorkJobs = [];
-  cwd: string = '';
   running: boolean = false;
   toolRunner: ToolRunner;
   private progressTimer?: NodeJS.Timeout;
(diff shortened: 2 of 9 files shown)


jyoungyun:
/cc @yunjayh
This PR removes the dependency with `obtainWorkspaceRoot()` function in Job runner.
If there is no workspace, you can see the ToolchainView and several operation(install, uninstall) works well by handling the `OneExplorer` error.
What I suggested before(#972) was not enough to handle toolchain operation. I'm sorry to bother you.

Related PR: #951

src/Project/JobRunner.ts
```diff
@@ -45,9 +45,10 @@ export class JobRunner extends EventEmitter {
     this.on(K_CLEANUP, this.onCleanup);
   }
 
-  private invoke(job: Job, path: string) {
+  private invoke(job: Job) {
     let tool: string = job.tool;
     let toolArgs: ToolArgs = job.toolArgs;
+    let workDir: string = job.workDir;
```
YongseopKim, line 51: How about handling if job.wordDir is empty?

YongseopKim (approved):
LGTM

src/Project/JobCommand.ts
```diff
@@ -37,6 +38,7 @@ class JobCommand implements Job {
     this.jobType = Job.Type.tUndefined;
     this.name = '';
     this.valid = false;
+    this.workDir = require('os').homedir();
```
yunjayh, line 41: I felt assigning as homedir is better to implemented in each child classes' constructor.

+) I'm not familiar about the history of `JobCommand`. Does `JobCommand` is designed to be a parent class of *workspace independent* commands only, such as `install`, `uninstall`, etc.? If it is, never mind the suggestion above.

src/Project/JobRunner.ts
jyoungyun: There are two cases of implementing `Job` interface.
- JobCommand
- JobBase

JobCommand class defines the `workDir` as `require('os').homedir()` function because this class usually is used by Toolchain or Device APIs to request install, uninstall or register somethings.
JobBase class defines the `workDir` as '' null string. It will be replaced by `workDir = helpers.obtainWorkspaceRoot();` value at the point when it is really executing.

As you know, we usually use `JobCommand` class and `JobBase` is a remnant of old codes.

YongseopKim: 👍

src/Project/JobCommand.ts
jyoungyun: > Does `JobCommand` is designed to be a parent class of *workspace independent* commands only, such as `install`, `uninstall`, etc.? 

The original design of `JobCommand` depended on workspace. Because it was only used to run compiler. There was no workflow about Toolchain at the initial version. However, the design has been changed. We use `JobCommand` to request job related to Toolchain and to run compile. If it requests to run Toolchain job, it can run workspace independent. And if it requests to run Compile job, it should run at the `cfg` location. So I defined the `workDir` of `JobCommand` as os home directory and only `JobConfig` job has different workDir as per its cfg location.

I will apply your suggestion.

yunjayh: AFAIU `JobCommand` is designed for toolchain workflow, and it can run workspace independent. 

IMO it's okay for current status. :smile: Now, the child classes should define `workDir` if the child process needs specific working directory. Let's go which this.

yunjayh (approved):
LGTM Thank you!!!

src/Project/JobCommand.ts
jyoungyun: After talking with @yunjayh, we agreed to keep the code as it is.

PR #1005, opened 2022-07-11
[CfgEditor] Fix to use acquireVsCodeApi only once

llFreetimell:
Currently, acquireVsCodeApi was called in `index.js` and it cannot be referenced from other js files.
However, calling `acquireVsCodeApi` more than one time causes an error.
("An instance of the VS Code API has already been acquired")
To resolve these issues, this commit introduces our own API to use VsCodeApi.

### DIFF
--- a/media/CfgEditor/index.js
+++ b/media/CfgEditor/index.js
@@ -14,12 +14,11 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


YongseopKim (approved):
LGTM

jyoungyun (approved):
LGTM

PR #1009, opened 2022-07-11
[Toolchain] Introduce DefaultToolchain class

jyoungyun:
This commit introduces DefaultToolchain class.

Draft PR: #854

### DIFF
--- a/docs/HowToUse.md
+++ b/docs/HowToUse.md
@@ -4,13 +4,13 @@ Let's compile an example by using ONE-vscode.
 
 ## Install the toolchain
 
-First, you need the ONE toolchain. Let's install the ONE toolchain. (You should install the in-house backend to follow this.)
+You need the ONE toolchain. Let's install the ONE toolchain. (You should install the in-house backend to follow this.)
 
 <img src="https://user-images.githubusercontent.com/7223627/174947559-71ab213c-68c4-4700-bcec-e55f6a2203bf.gif" width=600 />
 
 ## Set default toolchain
 
-To compile model using ONE toolchain, you must specify the default toolchain. To specify a default toolchain, if you mouse up the toolchain item you want to use in Toolchain view, :heavy_check_mark: icon appears. Click this icon to specify it as default toolchain. If toolchain is specified as default toolchain, the icon changes.
+To compile model using ONE toolchain, you must specify the default toolchain. Default toolchain refers to which toolchain you will use to run ONE related tools. To specify a default toolchain, if you mouse up the toolchain item you want to use in Toolchain view, :heavy_check_mark: icon appears. Click this icon to specify it as default toolchain. If toolchain is specified as default toolchain, the icon changes.
 
 | Toolchain | Default Toolchain |
 |-----------|-------------------|
@@ -18,7 +18,7 @@ To compile model using ONE toolchain, you must specify the default toolchain. To
 
 ## Prepare a NN model and a config file
 
-Second, you need a NN model and a config file to run compiling. Now you can get one from [res/modelDir/truediv](../res/modelDir/truediv/). Add it to your workspace. Now you can see a model and a config.
+You need a NN model and a config file to run compiling. Now you can get one from [res/modelDir/truediv](../res/modelDir/truediv/). Add it to your workspace. Now you can see a model and a config.
 
 <img src="https://user-images.githubusercontent.com/10216715/174798969-eee44fea-bd71-4e6a-8e2c-9e1de37ad74a.gif" width=600 />
 
--- a/docs/HowToUse.md
+++ b/docs/HowToUse.md
@@ -8,6 +8,14 @@ First, you need the ONE toolchain. Let's install the ONE toolchain. (You should 
 
 <img src="https://user-images.githubusercontent.com/7223627/174947559-71ab213c-68c4-4700-bcec-e55f6a2203bf.gif" width=600 />
 
+## Set default toolchain
+
+To compile model using ONE toolchain, you must specify the default toolchain. To specify a default toolchain, if you mouse up the toolchain item you want to use in Toolchain view, :heavy_check_mark: icon appears. Click this icon to specify it as default toolchain. If toolchain is specified as default toolchain, the icon changes.
+
+| Toolchain | Default Toolchain |
+|-----------|-------------------|
+| <img src="https://user-images.githubusercontent.com/7223627/177962615-71e64f52-e684-432d-a053-07896a20a479.png" width=300 /> | <img src="https://user-images.githubusercontent.com/7223627/177963453-d22998a0-be21-4531-a70b-9057a204e7eb.png" width=300 /> |
+
 ## Prepare a NN model and a config file
 
 Second, you need a NN model and a config file to run compiling. Now you can get one from [res/modelDir/truediv](../res/modelDir/truediv/). Add it to your workspace. Now you can see a model and a config.
--- a/src/Toolchain/DefaultToolchain.ts
+++ b/src/Toolchain/DefaultToolchain.ts
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as vscode from 'vscode';
+
+import {Toolchain} from '../Backend/Toolchain';
+import {Logger} from '../Utils/Logger';
+import {ToolchainEnv} from './ToolchainEnv';
+
+class DefaultToolchain {
+  private static _instance: DefaultToolchain;
+  private _toolchainEnv?: ToolchainEnv;
+  private _toolchain?: Toolchain;
+
+  private constructor() {
+    // This is private constructor
+  }
+
+  static getInstance() {
+    return this._instance || (this._instance = new this());
+  }
+
+  async ask(toolchainEnv: ToolchainEnv, toolchain: Toolchain): Promise<boolean> {
+    return new Promise<boolean>((resolve, reject) => {
+      const name = `${toolchain.info.name}-${toolchain.info.version ?.str()}`;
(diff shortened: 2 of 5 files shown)


jyoungyun:
default_toolchain_operation

jyoungyun:
@hyunsik-yoon I will implement cancel job after merged this PR.

YongseopKim (dismissed):
LGTM

src/Toolchain/DefaultToolchain.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  getToolchain(): Toolchain|undefined {
+    return this._toolchain;
+  }
+
+  getToolchainEnv(): ToolchainEnv|undefined {
+    return this._toolchainEnv;
+  }
+
+  equal(toolchain: Toolchain) {
```
dayo09, line 70: (optional) equal -> equals

It's very trivial, think it as an optional.

src/Toolchain/ToolchainProvider.ts
```diff
@@ -113,62 +120,65 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
...
-                }
-              });
-        },
-        () => {
-          Logger.info(this.tag, 'Installation is canceled.');
-        });
+    const notifyInstalled = (toolchainEnv: ToolchainEnv, toolchain: Toolchain) => {
+      const name = `${toolchain.info.name}-${toolchain.info.version ?.str()}`;
+      vscode.window.showInformationMessage(`Install ${name} successfully`);
+      DefaultToolchain.getInstance().ask(toolchainEnv, toolchain).then(() => this.refresh());
```
dayo09, line 136: (optional)
It looks handy that user is asked whether to set the toolchain as `default toolchain` when installation is done.

I think users may expect the toolchain to remember the default toolchain between vscode's on and off.
So I suggest the furthur feature: remembering the default toolchain

dayo09 (approved):
LGTM, I left several comments but it's okay to skip them

src/Toolchain/ToolchainProvider.ts
jyoungyun: > So I suggest the furthur feature: remembering the default toolchain

I think so. I have a plan to save the default toolchain information in vscode. So I asked you about the related content a few days ago. :)
I made an issue. #1015

hyunsik-yoon:
I was trying to think over future UI consideration as follows:

When I run one-vscode, I see there is one toolchain installed.
When I click to run cfg, error shows
image

🐕 First question is, "Do all user understand the meaning of `default toolchain`?" I am afraid that some user might not be sure about the meaning.

🐕 Next question at this moment is... "I have only one toolchain installed. Why don't one-vscode make it as default toolchain?"

When I move to `V` icon, it says `"Set default toolchain"`. So I clicked 'V' to make this toolchain as default one.
🐕  I noticed something changed... but I didn't know what the change was.
Later I found that icon was changed but it is kinda difficult what are the difference between icons of default toolchain and non-default toolchain. IMHO there should be noticeable difference to distinguish default one and non-default ones.

Anyway, it seemed that I set the toolchain as default one. But I was not sure... I wanted to check the toolchain is now default one. 🐕 However there is no way for me to find the answer for this question.

I also tried to find the way to make it non-default but couldn't find how.

Some suggestions..... are.
- When there is only one toolchain installed, we'd better make it as default one.
- We need some noticeable icon when a toolchain is default one. (instead of 
image
how about something like ✔️ to show that this toolchain was _selected_?) 
- Hovering the default toolchain name may show that that this toolchain was selected as default toolchain.

src/Toolchain/ToolchainProvider.ts
```diff
@@ -182,15 +192,51 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
...
+
+    if (!activeToolchainEnv || !activeToolchain) {
+      this.error(
+              'Default toolchain is not set. Please install toolchain and set the default toolchain.',
+              'Instruction', 'OK')
+          .then((value) => {
+            if (value === 'Instruction') {
+              Logger.info(this.tag, 'Open ONE-vscode HowToUse document');
+              vscode.env.openExternal(vscode.Uri.parse(
+                  'https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md'));
```
hyunsik-yoon, line 216:
```suggestion
                  'https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md#set-default-toolchain'));
```

docs/HowToUse.md
```diff
@@ -8,6 +8,14 @@ First, you need the ONE toolchain. Let's install the ONE toolchain. (You should
...
+
+To compile model using ONE toolchain, you must specify the default toolchain. To specify a default toolchain, if you mouse up the toolchain item you want to use in Toolchain view, :heavy_check_mark: icon appears. Click this icon to specify it as default toolchain. If toolchain is specified as default toolchain, the icon changes.
+
+| Toolchain | Default Toolchain |
+|-----------|-------------------|
+| <img src="https://user-images.githubusercontent.com/7223627/177962615-71e64f52-e684-432d-a053-07896a20a479.png" width=300 /> | <img src="https://user-images.githubusercontent.com/7223627/177963453-d22998a0-be21-4531-a70b-9057a204e7eb.png" width=300 /> |
+
 ## Prepare a NN model and a config file
 
 Second, you need a NN model and a config file to run compiling. Now you can get one from [res/modelDir/truediv](../res/modelDir/truediv/). Add it to your workspace. Now you can see a model and a config.
```
hyunsik-yoon, line 21: We may not need `First,` and `Second,` at L7 and L21. Or we could add `second,` at L13.

src/Toolchain/ToolchainProvider.ts
```diff
@@ -182,15 +192,51 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
...
       this.refresh();
     });
   }
 
-  run(_cfg: string) {
-    throw Error('Not implemented yet');
+  run(cfg: string, backend?: string, toolchain?: Toolchain) {
+    const activeToolchainEnv =
+        backend ? gToolchainEnvMap[backend] : DefaultToolchain.getInstance().getToolchainEnv();
+    const activeToolchain = toolchain ? toolchain : DefaultToolchain.getInstance().getToolchain();
```
hyunsik-yoon, line 206: When we have two backends (`foo` and `bar`) and default toolchain is `bar 1.0`,
I am afraid that, e.g., `run(a.cfg, foo)` where param `toolchain === undefined` could lead mismatch env and toolchain (`activeToochainEnv === gToolchainEnvMap[foo]` and `activeToolchain === bar 1.0`).

jyoungyun: I already handled the mismatch case. 

```
    if (!activeToolchainEnv?.listInstalled().find(value => JSON.stringify(value.info) === JSON.stringify(activeToolchain.info))) {
      this.error('Invalid Toolchain');
      return;
    }
```

Does this code resolve your concern?

jyoungyun:
> - When there is only one toolchain installed, we'd better make it as default one.

If we set automatically when there is one toolchain installed, the user will not be aware of the default toolchain concept. What about specifying the toolchain the user want to use at first, save this value in the workspace, and set it automatically the next time you start?

>- We need some noticeable icon when a toolchain is default one. (instead of 
image
how about something like ✔️ to show that this toolchain was _selected_?) 

Okay, I will update this change in another PR.

>- Hovering the default toolchain name may show that that this toolchain was selected as default toolchain.

I will also apply this changes in another PR.

jyoungyun:
Please review again, @hyunsik-yoon @YongseopKim @dayo09

src/Toolchain/DefaultToolchain.ts
jyoungyun: I modified the function name to `isEqual`. :)

src/Toolchain/ToolchainProvider.ts
hyunsik-yoon: I am afraid that `listInstalled()` might be somewhat heavy for this check.,

If there is no fancy way,
how about
- ` run(cfg: string, backend?: string, toolchain?: Toolchain)` ==>  `run(cfg: string)` if we don't have serious requirements with `backend` and `toolchain` params
- or adding `//TODO ...` T.T

jyoungyun: I thought there would be a UX where the user directly selects and calls the toolchain without using the default toolchain. For that reason, it was designed to receive more arguments. But.. I will fix it.

src/Toolchain/ToolchainProvider.ts
dayo09: I remember that. Now I understand the context. :-D

src/Toolchain/DefaultToolchain.ts
dayo09: That one is rather about the verb's form. It was inspired by https://github.com/Samsung/ONE-vscode/pull/1012#discussion_r918471558, like he says, it seems the other library functions uses `exists` instead of `exist` as the object is singular and it uses the verb.

dayo09: > I modified the function name to isEqual. :)

I like this too :_D

YongseopKim:
Conflicted

jyoungyun:
> When users install Debian-based toolchain only, I am not sure whether they should know about concept of default toolchain. Since there is only one installed, ONE-vscode could easily know the installed one is default one... Users may think default toolchain concept is redundant.
> 
> One other thing regarding user-experience is that... 10 out of 10 users like me, who expect to see pleasant 'compilation was successful' message, will see error message instead.

I see. If the installed toolchain is just one, I will set it as a default toolchain.

hyunsik-yoon:
> /cc @hyunsik-yoon How about this changes?

I personally think that it is somewhat redundant to show `Set ... as a default toolchain [OK]` when only one toolchain is installed. But it's my personal preference. I think current UI is good. :-)

Can I suggest one more thing? Once a user selects a default toolchain, it would be great if the selection could be persistent - meaning that default toolchain selected previously could be continue to be a default toolchain in next running of one-vscode without showing `Set ... as a default toolchain [OK]` again. :-)

hyunsik-yoon (approved):
Thanks.

jyoungyun:
> Can I suggest one more thing? Once a user selects a default toolchain, it would be great if the selection could be persistent - meaning that default toolchain selected previously could be continue to be a default toolchain in next running of one-vscode without showing Set ... as a default toolchain [OK] again. :-)

I think your suggestion is like https://github.com/Samsung/ONE-vscode/pull/1009#discussion_r918501823, am I right?

If yes, I also think it would be better to keep the default toolchain value regardless of restart. This issue will be handled in a separated PR.

jyoungyun:
PTAL /cc @dayo09 @YongseopKim

YongseopKim (approved):
LGTM

dayo09 (approved):

PR #1010, opened 2022-07-11
[DeviceAPI] revisit Executor to make this use on Device API

struss:
This commit will revisit `Executor` for some reason

about `name()`: if we add None Backend Executor, we need to add Executor name on this.
                          otherwise, follow Backend Name.
about `require()`: to use on `satisfied()` on Device API.

### DIFF
--- a/src/Backend/Executor.ts
+++ b/src/Backend/Executor.ts
@@ -15,9 +15,13 @@
  */
 
 import {Command} from './Command';
+import { DeviceSpec } from './Spec';
 import {Toolchains} from './Toolchain';
 
 interface Executor {
+  // Executor unique name
+  name(): string;
+
   // exetensions of executable files
   getExecutableExt(): string[];
 
@@ -26,10 +30,17 @@ interface Executor {
 
   // TODO: use cfg path to run onecc after one-infer landed
   runInference(_modelPath: string, _options?: string[]): Command;
+  
+  // return deviceSpec that required for this Executor to run
+  require(): DeviceSpec;
 }
 
 // General excutor uses onecc so default jobs can be used
 class ExecutorBase implements Executor {
+  name(): string {
+    throw new Error('Name are not determinded.');
+  }
+
   // exetensions of executable files
   getExecutableExt(): string[] {
     throw Error('Invalid toolchains call');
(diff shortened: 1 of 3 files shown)


YongseopKim (approved):
LGTM

src/Backend/Executor.ts
```diff
@@ -26,10 +30,17 @@ interface Executor {
...
   runInference(_modelPath: string, _options?: string[]): Command;
+
+  // return deviceSpec that required for this Executor to run
+  require(): DeviceSpec;
 }
 
 // General excutor uses onecc so default jobs can be used
 class ExecutorBase implements Executor {
+  name(): string {
+    throw new Error('Name are not determinded.');
```
hyunsik-yoon, line 41: ??

```suggestion
    throw new Error('Name is not determined.');
```

src/Backend/Executor.ts
```diff
@@ -26,10 +30,17 @@ interface Executor {
 
   // TODO: use cfg path to run onecc after one-infer landed
   runInference(_modelPath: string, _options?: string[]): Command;
+
+  // return deviceSpec that required for this Executor to run
```
hyunsik-yoon, line 34:
```suggestion
  // return deviceSpec required for this Executor to run
```

hyunsik-yoon: or 

```suggestion
  // return deviceSpec that is required for this Executor to run
```

src/Backend/Executor.ts
```diff
@@ -55,7 +55,7 @@ class ExecutorBase implements Executor {
   }
 
   require(): DeviceSpec {
-    throw new Error('Invalid require Spec call.');
+    throw new Error('Invalid DeviceSpec call.');
```
hyunsik-yoon, line 58: `DeviceSpec` is a return type and I am not sure whether `'Invalid DeviceSpec call.'` is appropriate.

For example, if we have a method like the following, 
`'invalid int call` sounds not appropriate.

```
int add(a, b) {  // return int type
   if (some error situation)
      throw new Error('invalid int call')
}
```

struss: um.... I will fix it.

hyunsik-yoon (approved):

YongseopKim (approved):

PR #1011, opened 2022-07-11
[OneExplorer] Refactor to get artifact list

dayo09:
This commit refactors OneExplorer, ConfigObject and Helpers
to get Artifact list.

ONE-vscode-DCO-1.0-Signed-off-by: dayo09 <dayoung.lee@samsung.com>

PR #1012, opened 2022-07-11
[Utils] Revisit RealPath

dayo09:
This commit rewrites RealPath's equal commands.

---

From #1011

### DIFF
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -57,7 +57,7 @@ export class RealPath {
     return fs.existsSync(absPath) ? new RealPath(absPath) : null;
   }
 
-  public static exist(rawPath: string|undefined): boolean {
+  public static exists(rawPath: string|undefined): boolean {
     if (!rawPath) {
       return false;
     }
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -36,14 +36,34 @@ export class RealPath {
     this.absPath = absPath;
   }
 
-  public static isEqual(path0: RealPath, path1: RealPath): boolean {
-    return path0.absPath === path1.absPath;
+  public static areEqual(path0: string, path1: string): boolean {
+    const realPath0 = this.createRealPath(path0);
+    const realPath1 = this.createRealPath(path1);
+
+    if (!realPath0 || !realPath1) {
+      return false;
+    }
+
+    return realPath0.equal(realPath1);
+  }
+
+  public equal(lhs: RealPath): boolean {
+    return this.absPath === lhs.absPath;
   }
 
   public static createRealPath(rawPath: string): RealPath|null {
     const absPath = path.resolve(path.normalize(rawPath));
 
     return fs.existsSync(absPath) ? new RealPath(absPath) : null;
+  }
+
+  public static exist(rawPath: string|undefined): boolean {
+    if (!rawPath) {
+      return false;
+    }
+    const absPath = path.resolve(path.normalize(rawPath));
+
+    return fs.existsSync(absPath) ? true : false;
   }
 }
 
--- a/src/Tests/Utils/Helpers.test.ts
+++ b/src/Tests/Utils/Helpers.test.ts
@@ -51,7 +51,7 @@ suite('Utils', function() {
       });
     });
 
-    suite('#isEqual()', function() {
+    suite('#equal()', function() {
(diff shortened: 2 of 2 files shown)


src/Utils/Helpers.ts
```diff
@@ -36,15 +36,35 @@ export class RealPath {
     this.absPath = absPath;
   }
 
-  public static isEqual(path0: RealPath, path1: RealPath): boolean {
-    return path0.absPath === path1.absPath;
+  public static areEqual(path0: string, path1: string): boolean {
```
dayo09, line 39: This function is handy to process `equal` to string paths.

src/Utils/Helpers.ts
```diff
@@ -36,15 +36,35 @@ export class RealPath {
...
+    return this.absPath === lhs.absPath;
   }
 
   public static createRealPath(rawPath: string): RealPath|null {
     const absPath = path.resolve(path.normalize(rawPath));
 
     return fs.existsSync(absPath) ? new RealPath(absPath) : null;
   }
+
+  public static exist(rawPath: string|undefined): boolean {
```
dayo09, line 60: This command makes users easy to check if the path exists.

src/Utils/Helpers.ts
seanshpark: I usually seen `is~~~` as `path0 is equal to path1 ?`
I've never seen `are~~~` and at first seeing this, I interpreted this as `path0: string[] are equal to path1: string[] ?`

src/Utils/Helpers.ts
yunjayh: (optional) For me, `pathlib.Path.exists()` is friendly. How about `exists`?

src/Utils/Helpers.ts
yunjayh: +1 to `isEqual`.

src/Utils/Helpers.ts
```diff
@@ -36,15 +36,35 @@ export class RealPath {
...
+    if (!rawPath) {
+      return false;
+    }
```
yunjayh, line 63: IMHO for the case of undefined rawPath, how about returning the `this.abspath`'s existence? When I read this code, I was expected to use `exist` method with a `RealPath` instance, for example 
```ts
filepath = RealPath('~/my_file.cfg');
filepath.exist()
```

src/Utils/Helpers.ts
dayo09: I tried to search on a google:

Unity assertion, .NET assertion, C# Selenium, etc,... Uses areEqual. (Google search result : 174,000,000+)
iBatis, Apple uses isEqual, (Apple's for an unary compare operator.) (Google search result: 827,000+)

src/Utils/Helpers.ts
dayo09: It's not a member function, but a static function.

Usage: 
```
RealPath.exist('~/my_file.cfg');
```

yunjayh: 😮 Now I catch it..!

yunjayh: Now I catch it..! 😮

src/Utils/Helpers.ts
dayo09: > I've never seen are~~~ and at first seeing this, I interpreted this as path0: string[] are equal to path1: string[] ?

As far as I know, code must be read as it is. In this example,

```areEqual(path0, path1)```

So people will read as: "Are equal: path0 and path1". It sounds gramatically correct.

Whereas this sounds awkward: "Is equal: path0 and path1". 

If it's an unary operator, it isEqual will sound more naturally:
```path0.isEqual(path1)```
"Path0 isEqual to path1"

seanshpark: This is your change so you can go as you wish.
But finding examples from internet with `areEqual`, with .NET, I'm not sure this is appropriate.
I've found some examples with TypeScript which I agree upon.
- https://dev.to/macsikora/typescript-is-more-than-you-think-2nbf
- https://www.anycodings.com/questions/how-to-compare-arrays-in-javascript

Anyway, it's up to you.

yunjayh (approved):
LGTM

dayo09:
@Samsung/one-vscode PTAL :-D

YongseopKim (approved):
LGTM

PR #931, opened 2022-07-04
[OneExplorer] Refactor OneExplorer

llFreetimell:
As `OneExplorer` class is used for anywhere except constructor,
extracting constructor of `OneExplorer` separate function seems better.
Meanwhile, this commit removes unused codes in `OneExplorer`.

ONE-vscode-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #1013, opened 2022-07-12
[OneExplorer] Handle no workspace opened case

yunjayh:
This commit modifies OneExplorer to handle no workspace opened cases.
To introduce welcome view, showing pop-up message to user should be deprecated.

---

draft: #953

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -158,7 +158,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
 
   private tree: Node|undefined;
 
-  constructor(private workspaceRoot: vscode.Uri) {
+  constructor(private workspaceRoot: vscode.Uri|undefined) {
     const fileWatchersEvents =
         [this.fileWatcher.onDidCreate, this.fileWatcher.onDidChange, this.fileWatcher.onDidDelete];
 
@@ -524,8 +524,7 @@ export function initOneExplorer(context: vscode.ExtensionContext) {
     }
   }
 
-  // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
-  const oneTreeDataProvider = new OneTreeDataProvider(workspaceRoot!);
+  const oneTreeDataProvider = new OneTreeDataProvider(workspaceRoot);
 
   let treeView: vscode.TreeView<OneNode|undefined>|undefined = vscode.window.createTreeView(
       'OneExplorerView',
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -20,6 +20,7 @@ import {TextEncoder} from 'util';
 import * as vscode from 'vscode';
 
 import {CfgEditorPanel} from '../CfgEditor/CfgEditorPanel';
+import {Balloon} from '../Utils/Balloon';
 import {obtainWorkspaceRoot, RealPath} from '../Utils/Helpers';
 import {Logger} from '../Utils/Logger';
 
@@ -518,9 +519,14 @@ export function initOneExplorer(context: vscode.ExtensionContext) {
     workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
   } catch (e: unknown) {
     if (e instanceof Error) {
-      Logger.error('OneExplorer', e.message);
+      if (e.message === 'Need workspace') {
+        Logger.info('OneExplorer', e.message);
+      } else {
+        Logger.error('OneExplorer', e.message);
+        Balloon.error('Something goes wrong while setting workspace.', true);
+      }
     } else {
-      throw e;
+      Logger.error('OneExplorer', 'Unknown error has been thrown.');
     }
   }
 
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -344,7 +344,6 @@ input_path=${modelName}.${extName}
 
   getChildren(element?: OneNode): OneNode[]|Thenable<OneNode[]> {
     if (!this.workspaceRoot) {
-      vscode.window.showInformationMessage('Cannot find workspace root');
       return Promise.resolve([]);
     }
 
@@ -514,7 +513,17 @@ input_path=${modelName}.${extName}
 
 export function initOneExplorer(context: vscode.ExtensionContext) {
   // TODO Support multi-root workspace
-  let workspaceRoot: vscode.Uri = vscode.Uri.file(obtainWorkspaceRoot());
+  let workspaceRoot: vscode.Uri|undefined = undefined;
+  try {
+    workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      Logger.error('OneExplorer', e.message);
+    } else {
+      throw e;
+    }
+  }
+
   // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
   const oneTreeDataProvider = new OneTreeDataProvider(workspaceRoot!);
 


seanshpark:
typo `hanlde`

yunjayh:
Commit message fixed

src/OneExplorer/OneExplorer.ts
```diff
@@ -501,12 +500,20 @@ input_path=${modelName}.${extName}
...
-    // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
-    const oneTreeDataProvider = new OneTreeDataProvider(this.workspaceRoot!);
+    try {
+      this.workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+    } catch (e: unknown) {
+      let errmsg = 'Workspace has not been defined yet.';
+      if (e instanceof Error) {
+        errmsg = e.message;
+      }
+      Logger.info('OneExplorer', errmsg);
```
dayo09, line 514:
```suggestion
      Logger.error('OneExplorer', errmsg);
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -501,12 +500,20 @@ input_path=${modelName}.${extName}
...
   public treeView: vscode.TreeView<OneNode|undefined>|undefined;
 
   constructor(context: vscode.ExtensionContext) {
-    // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
-    const oneTreeDataProvider = new OneTreeDataProvider(this.workspaceRoot!);
+    try {
+      this.workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+    } catch (e: unknown) {
+      let errmsg = 'Workspace has not been defined yet.';
+      if (e instanceof Error) {
```
dayo09, line 511: Q. What is normally given as e?
Under what situation does errmsg get printed..?

yunjayh: I referenced few articles, and [here](https://velog.io/@shroad1802/%ED%83%80%EC%9E%85%EC%8A%A4%ED%81%AC%EB%A6%BD%ED%8A%B8%EC%99%80-catch%EC%A0%88-%EC%97%90%EB%9F%AC-%EC%B2%98%EB%A6%AC) is one of them. Since typescript can throw any types as error, it's usual to use such error handling method.

jyoungyun: Using `try/catch` in the constructor is not recommended because it cannot be called again. However, in our use case, we do not need to invoke this constructor again, and when user opens the workspace, it restarts and invokes this function again. So I agree this PR. :)

(optional) How about showing the error message to user? It only prints the error on output window.

jyoungyun (approved):
LGTM

yunjayh:
Let me rewrite this code after #931. I didn't recognize there is a refactoring PR for `OneExplorer.ts`.

src/OneExplorer/OneExplorer.ts
yunjayh: Now I understood. Then, how can we log the error message? How do you think about this approach?

```ts
catch (e: unknown){
  if (e instance of Error) {
    Logger.error('OneExplorer', e.message);
  } else {
    throw e
  }
}
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -500,7 +499,17 @@ input_path=${modelName}.${extName}
...
+  let workspaceRoot: vscode.Uri|undefined = undefined;
+  try {
+    workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      Logger.error('OneExplorer', e.message);
+    } else {
+      throw e;
+    }
+  }
```
yunjayh, line 511: I changed the way to handle the error. How about this way? /cc @dayo09

src/OneExplorer/OneExplorer.ts
```diff
@@ -500,7 +499,17 @@ input_path=${modelName}.${extName}
...
+
   // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
   const oneTreeDataProvider = new OneTreeDataProvider(workspaceRoot!);
```
hyunsik-yoon, line 514: I wonder what happen here when `(e instanceof Error)`. I guess `workspace` is `undefined` and is it OK to create `TreeDataProvider(undefined)` ?

yunjayh: I expected it will be okay since `workspaceRoot!` will block passing `undefined` to the constructor, but it wasn't. I'll fix and add a patch ASAP

src/OneExplorer/OneExplorer.ts
```diff
@@ -500,9 +499,18 @@ input_path=${modelName}.${extName}
...
-  // NOTE: Fix `obtainWorksapceRoot` if non-null assertion is false
-  const oneTreeDataProvider = new OneTreeDataProvider(workspaceRoot!);
+  let workspaceRoot: vscode.Uri|undefined = undefined;
+  try {
+    workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      Logger.error('OneExplorer', e.message);
+    } else {
+      throw e;
```
jyoungyun, line 509: (Q) What case does this if/else syntax distinguish?

If you throw an error, ToolchainView does not work. Is it your intention?

src/OneExplorer/OneExplorer.ts
hyunsik-yoon: @dayo, Could you check if `new OneTreeDataProvider(undefined)` is OK or not. Thanks.

dayo09: @hyunsik-yoon 

What happens is : It'll show an information message "Cannot ifnd workspace root" and shows nothing. 

```ts
  getChildren(element?: OneNode): OneNode[]|Thenable<OneNode[]> {
    if (!this.workspaceRoot) {
      vscode.window.showInformationMessage('Cannot find workspace root'); // Shows an information message
      return Promise.resolve([]);
    }
```

src/OneExplorer/OneExplorer.ts
dayo09: As @jyoungyun says above, I think checking whether it's an instance of Error doens't carry any information here.. I think you can decide here is the treatment for whatever cases. - "whether to throw or log"..

yunjayh:
```suggestion
  try {
    workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
  } catch (e: unknown) {
    if (e instanceof Error && e.message == 'Need workspace') {
      Logger.error('OneExplorer', e.message);
    } else {
      throw e;
    }
  }
```

@dayo09 @jyoungyun This change can check whether the error is what we expected or not. If the error is expected no-workspace error, it's okay to keep going. 
IMHO break because of unexpected error is a feature need to have. How do you think of this approach?

dayo09: I agree with the intention. I think it will be clearer if an Error is specifically defined from `obtainWorkspaceRoot`.

----
+ About throwing an error

I just checked what happens when OneExplorer throws: The whole extension throws. 

0714-throw

So an expected error from OneExplorer shuts down the whole extension, including ToolchainView and TargetDeviceView....
Maybe it's a right thing to do...? I don't know.

hyunsik-yoon (dismissed):

dayo09 (dismissed):

src/OneExplorer/OneExplorer.ts
yunjayh:
```ts
  try {
    workspaceRoot = vscode.Uri.file(obtainWorkspaceRoot());
  } catch (e: unknown) {
    if (e instanceof Error) {
      if (e.message == 'Need workspace') {
        Logger.info('OneExplorer', e.message);
      } else {
        Logger.error('OneExplorer', e.message);
        Balloon.error('Something goes wrong while setting workspace.', true);
        // throw e;
      }
    } else {
      Logger.error('OneExplorer', 'Unknown error has been thrown.');
      // throw e;
    }
  }
```

This version handles unexpected error, and print the separated logs for each types. I don't know either wheter throwing is right thing to do..

dayo09: Agree. Anyways think we can talk about it later on.

dayo09 (approved):
LGTM, Thanks

jyoungyun (approved):
LGTM

yunjayh:
@Samsung/one-vscode_committers ping :smile:

hyunsik-yoon (approved):

PR #1014, opened 2022-07-12
[Backend] Add registerExecutor API on backend register

struss:
This commit will add `registerExcutor` API on `backendRegistrationApi`.

`registerExecutor` API will use to register executor and use this on Device API.

### DIFF
--- a/src/Backend/Backend.ts
+++ b/src/Backend/Backend.ts
@@ -35,7 +35,7 @@ let globalExecutorArray: Executor[] = [];
 
 function backendRegistrationApi() {
   const logTag = 'backendRegistrationApi';
-  return {
+  let registrationAPI = {
     registerBackend(backend: Backend) {
       const backendName = backend.name();
       assert(backendName.length > 0);
@@ -55,6 +55,8 @@ function backendRegistrationApi() {
       Logger.info(logTag, 'Executor', executor.name(), 'was registered into ONE-vscode.');
     }
   };
+
+  return registrationAPI;
 }
 
 export {globalBackendMap, globalExecutorArray, backendRegistrationApi};
--- a/src/Backend/Backend.ts
+++ b/src/Backend/Backend.ts
@@ -18,6 +18,7 @@ const assert = require('assert');
 import {Backend} from './API';
 import {gToolchainEnvMap, ToolchainEnv} from '../Toolchain/ToolchainEnv';
 import {Logger} from '../Utils/Logger';
+import {Executor} from './Executor';
 
 /**
  * Interface of backend map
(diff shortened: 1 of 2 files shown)


src/Backend/Backend.ts
```diff
@@ -29,10 +30,12 @@ interface BackendMap {
 
 // List of backend extensions registered
 let globalBackendMap: BackendMap = {};
+// List of Executor extensions registered
+let globalExecutorArray: Executor[] = [];
 
 function backendRegistrationApi() {
   const logTag = 'backendRegistrationApi';
-  let registrationAPI = {
+  return {
```
struss, line 38: This change occur by `SonarLint`, typescript:S1488 rule.
```
Local variables should not be declared and then immediately returned or thrown (typescript:S1488)
```

struss: I think this change will not related with this PR. I will rebase this.

YongseopKim (approved):
LGTM

src/Tests/Backend/Backend.test.ts
```diff
@@ -35,7 +35,14 @@ class BackendMockup implements Backend {
...
+
+const executorName = 'Mockup';
+class ExecutorMockup extends ExecutorBase {
+  name(): string {
+    return executorName;
+  }
+}
 
 suite('Backend', function() {
   suite('backendRegistrationApi', function() {
```
yunjayh, line 48: Check after backend registration? (Did I understood the code correctly?)

```ts
assert.strictEqual(globalExecutorArray.length, 1);
```

src/Tests/Backend/Backend.test.ts
```diff
@@ -64,5 +84,6 @@ suite('Backend', function() {
     if (gToolchainEnvMap[backendName] !== undefined) {
       delete gToolchainEnvMap[backendName];
     }
+    globalExecutorArray.length = 0;
```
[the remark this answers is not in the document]
struss: Um... I will update this.

src/Tests/Backend/Backend.test.ts
struss: Oh I will add this. Thanks!

YongseopKim (approved):

src/Backend/Backend.ts
```diff
@@ -41,11 +44,19 @@ function backendRegistrationApi() {
       if (compiler) {
         gToolchainEnvMap[backend.name()] = new ToolchainEnv(compiler);
       }
+      const executor = backend.executor();
+      if (executor) {
+        globalExecutorArray.push(executor);
+      }
       Logger.info(logTag, 'Backend', backendName, 'was registered into ONE-vscode.');
+    },
+    registerExecutor(executor: Executor) {
```
dayo09, line 53: Q. Why is executor registered aside from Backend..? I thought it would included in backend.

struss: I thought there could be a Executor which is not related to Backend, so I make a room for this.

struss: but, if `Backend` could get multiple `Executor`, I think we don't keep this, as every `Executor` have each related `Backend`.

struss: I think #1076 change will remove this api as multiple executor could registerd on.

PR #1016, opened 2022-07-12
[OneExplorer] Rewrite ConfigObject 

dayo09:
This commit rewrite ConfigObject to..
(1) Return a list of Artifact instead of Uri
(2) Introduce getBase(Drived)ModelsExists function to filter existing artifacts
(3) Introduce isChildOf method and use it
(4) Remove redundant wrapper `parse` method and use `createObject`
(5) Pass string filepath instead of Uri as parameter of functions (parseDerivedModels, parseBaseModels)
(6) Replace map with forEach to remove unused variable

---

#1012 must be merged first.

From #1011

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -223,7 +223,7 @@ export class ConfigObj {
   }
 
   get getDerivedModels() {
-    return this.obj.baseModels;
+    return this.obj.derivedModels;
   }
 
   /**
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -41,7 +41,7 @@ interface Artifact {
   /**
    * A full path in file system
    */
-  path?: string;
+  path: string;
 }
 
 /**
@@ -144,7 +144,7 @@ export class Locator {
 
 // TODO Move to backend side with some modification
 export class LocatorRunner {
-  private artifactLocators: {artifact: Artifact, locator: Locator}[] = [];
+  private artifactLocators: {artifactAttr: {type: string, ext: string}, locator: Locator}[] = [];
 
   /**
    * A helper function to grep a filename ends with 'ext' within the given 'content' string.
@@ -158,7 +158,7 @@ export class LocatorRunner {
     return fileNames;
   };
 
-  public register(artifactLocator: {artifact: Artifact, locator: Locator}) {
+  public register(artifactLocator: {artifactAttr: {type: string, ext: string}, locator: Locator}) {
     this.artifactLocators.push(artifactLocator);
   }
 
@@ -172,13 +172,11 @@ export class LocatorRunner {
     let artifacts: Artifact[] = [];
 
     // Get Artifacts with {type, ext, path}
-    this.artifactLocators.forEach(({artifact, locator}) => {
+    this.artifactLocators.forEach(({artifactAttr, locator}) => {
       let filePaths: string[] = locator.locate(iniObj, dir);
       filePaths.forEach(filePath => {
-        // Clone this.artifact
-        let newArtifact: Artifact = Object.assign({}, artifact);
-        newArtifact.path = filePath;
-        artifacts.push(newArtifact);
+        let artifact: Artifact = {type: artifactAttr.type, ext: artifactAttr.ext, path: filePath};
+        artifacts.push(artifact);
       });
     });
 
@@ -232,22 +230,22 @@ export class ConfigObj {
    * Returns only the baseModels which exists in file system
    */
   get getBaseModelsExists() {
-    return this.obj.baseModels.filter(artifact => RealPath.exists(artifact.path!));
+    return this.obj.baseModels.filter(artifact => RealPath.exists(artifact.path));
   }
 
   /**
    * Returns only the derivedModels which exists in file system
    */
   get getDerivedModelsExists() {
-    return this.obj.derivedModels.filter(artifact => RealPath.exists(artifact.path!));
+    return this.obj.derivedModels.filter(artifact => RealPath.exists(artifact.path));
   }
 
   /**
    * Return true if the `baseModelPath` is included in `baseModels`
    */
   public isChildOf(baseModelPath: string): boolean {
-    const found = this.obj.baseModels.map(artifact => artifact.path!)
-                      .find(path => RealPath.areEqual(baseModelPath, path!));
+    const found = this.obj.baseModels.map(artifact => artifact.path)
+                      .find(path => RealPath.areEqual(baseModelPath, path));
 
     return found ? true : false;
   }
@@ -315,17 +313,17 @@ export class ConfigObj {
     let locatorRunner = new LocatorRunner();
 
     locatorRunner.register({
-      artifact: {type: 'model', ext: '.tflite'},
+      artifactAttr: {type: 'model', ext: '.tflite'},
       locator: new Locator((value: string) => LocatorRunner.searchWithExt('.tflite', value))
     });
 
     locatorRunner.register({
-      artifact: {type: 'model', ext: '.pb'},
+      artifactAttr: {type: 'model', ext: '.pb'},
       locator: new Locator((value: string) => LocatorRunner.searchWithExt('.pb', value))
     });
 
     locatorRunner.register({
-      artifact: {type: 'model', ext: '.onnx'},
+      artifactAttr: {type: 'model', ext: '.onnx'},
       locator: new Locator((value: string) => LocatorRunner.searchWithExt('.onnx', value))
     });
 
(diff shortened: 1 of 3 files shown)


llFreetimell:
Could you apply following too? :)

#### Before

```ts
    this.artifactLocators.forEach(({artifact, locator}) => {
      let filePaths: string[] = locator.locate(iniObj, dir);
      filePaths.map(filePath => {
        // Clone this.artifact
        let newArtifact: Artifact = Object.assign({}, artifact);
        newArtifact.path = filePath;
        artifacts.push(newArtifact);
```

#### After

```ts
    this.artifactLocators.forEach(({artifact, locator}) => {
      let filePaths: string[] = locator.locate(iniObj, dir);
      filePaths.forEach((filePath) => {
        // Clone this.artifact
        let newArtifact: Artifact = Object.assign({}, artifact);
        newArtifact.path = filePath;
        artifacts.push(newArtifact);
```

llFreetimell:
Yes, just for removing static warning :D
`map` returns new array but it is not used anywhere T.T

dayo09:
@llFreetimell I know. :-D Thanks for catching that out.

src/OneExplorer/ConfigObject.ts
```diff
@@ -215,37 +216,49 @@ export class ConfigObj {
...
   /**
-   * A simple parse function with directly `ConfigObj.obj` fields after parsing
-   * without keeping an ConfigObj instance.
-   *
-   * @param uri
-   * @returns `{baseModels, derivedModels}` or
-   *          `null` if failed to open/parse into ini object
+   * Returns only the baseModels which exists in file system
+   */
+  get getBaseModelsExists() {
```
dayo09, line 234: (2) Introduce getBase(Drived)ModelsExists function to filter existing artifacts

src/OneExplorer/ConfigObject.ts
```diff
@@ -321,26 +334,26 @@ export class ConfigObj {
...
   };
 
   /**
    * Find derived models written in the ini object and return the absolute path.
    *
    * @param uri cfg uri is required to calculate absolute path
    */
-  private static parseDerivedModels = (uri: vscode.Uri, iniObj: object): vscode.Uri[] => {
-    const dir = path.dirname(uri.fsPath);
+  private static parseDerivedModels = (filePath: string, iniObj: object): Artifact[] => {
```
dayo09, line 355: (5) Pass string filepath instead of Uri as parameter of functions (parseDerivedModels, parseBaseModels)

There is no specific reason for the function to get it as Uri.
In a while, this change removes redundant translation from string to uri.

src/OneExplorer/ConfigObject.ts
```diff
@@ -172,7 +174,7 @@ export class LocatorRunner {
     // Get Artifacts with {type, ext, path}
     this.artifactLocators.forEach(({artifact, locator}) => {
       let filePaths: string[] = locator.locate(iniObj, dir);
-      filePaths.map(filePath => {
+      filePaths.forEach(filePath => {
```
dayo09, line 177: (6) Replace map with forEach to remove unused variable

@llFreetimell 

> Yes, just for removing static warning :D
> map returns new array but it is not used anywhere T.T

dayo09:
@Samsung/one-vscode PTAL :-D

src/OneExplorer/ConfigObject.ts
```diff
@@ -215,37 +216,49 @@ export class ConfigObj {
...
-    this.obj = {
-      baseModels: ConfigObj.parseBaseModels(uri, rawObj),
-      derivedModels: ConfigObj.parseDerivedModels(uri, rawObj)
-    };
+  get getBaseModels() {
+    return this.obj.baseModels;
+  }
+
+  get getDerivedModels() {
+    return this.obj.baseModels;
```
jyoungyun, line 228:
```suggestion
    return this.obj.derivedModels;
```
 Maybe?

src/OneExplorer/ConfigObject.ts
```diff
@@ -215,37 +216,49 @@ export class ConfigObj {
...
 
-    if (!cfgObj) {
-      return null;
-    }
+  /**
+   * Return true if the `baseModelPath` is included in `baseModels`
+   */
+  public isChildOf(baseModelPath: string): boolean {
+    const found = this.obj.baseModels.map(artifact => artifact.path!)
+                      .find(path => RealPath.areEqual(baseModelPath, path!));
```
jyoungyun, line 250:
```suggestion
                      .find(path => path && RealPath.areEqual(baseModelPath, path));
```

(optional)
If the path is always defined, ignore my comment.

src/OneExplorer/ConfigObject.ts
```diff
@@ -321,26 +334,26 @@ export class ConfigObj {
     if (artifacts.length > 1) {
       // TODO Notify the error with a better UX
       // EX. put question mark next to the config icon
-      Logger.warn('OneExplorer', `There are multiple input models in the configuration(${uri}).`);
+      Logger.warn(
+          'OneExplorer', `There are multiple input models in the configuration(${filePath}).`);
```
jyoungyun, line 338: (optional)
What about defining the log tag of `OneExplorer` keyword?

src/OneExplorer/OneExplorer.ts
```diff
@@ -470,31 +470,32 @@ input_path=${modelName}.${extName}
...
-                           // Display only the existing node
-                           const realPath = RealPath.createRealPath(derivedModel.fsPath);
-                           if (realPath) {
-                             pairNode.childNodes.push(new Node(
-                                 NodeType.derivedModel, [], vscode.Uri.file(realPath.absPath)));
-                           }
-                         });
+      derivedModels ?.forEach(derivedModel => {
+                       // Display only the existing node
+                       const realPath = RealPath.createRealPath(derivedModel.path!);
```
jyoungyun, line 491: (Q) Is this path always defined?

src/OneExplorer/ConfigObject.ts
dayo09: Yes, the path is always defined by the point.

src/OneExplorer/ConfigObject.ts
dayo09: What do you mean..? Can you explain more?

src/OneExplorer/OneExplorer.ts
dayo09: Yes, the path is always defined by the point. (ditto)

src/OneExplorer/ConfigObject.ts
jyoungyun: I mean you can make a logtag for `OneExplorer`.

```
const logtag = 'OneExplorer';
Logger.warn(logtag, ...);
```

dayo09: Oh, sure... Can I apply the change later then?

src/OneExplorer/ConfigObject.ts
jyoungyun: Then, how about using `path: string;` instead of `path?: string;`?

from [src/OneExplorer/ConfigObject.ts](https://github.com/Samsung/ONE-vscode/pull/1016/files#diff-59c113e60e8574721c44c2b6f9f619191c57a5b338e920de950af5ca9db2f728L42)

src/OneExplorer/ConfigObject.ts
jyoungyun: Yes. It's up to you to apply this suggestion. :)

src/OneExplorer/ConfigObject.ts
dayo09: It's not possible for the moment, as the interface 'Artifact' is used when it asks 'ArtifactLocator' to find the path...

So 'ArtifactLocator' is given an Artifact{type, ext} and it returns Artifact{type, ext, path}.

I think that it's not a good design to use the interface for both query and return.. I think I need to rewrite to distinguish them. 

I will remove artifact from the 'artifact locator' query, and do as what you said: use `path: string;` instead of `path?: string;`

src/OneExplorer/ConfigObject.ts
```diff
@@ -39,7 +41,7 @@ interface Artifact {
   /**
    * A full path in file system
    */
-  path?: string
+  path: string;
```
dayo09, line 44: @jyoungyun Applied :-D

dayo09: Before, `Artifact` was used for query too, with an undefined `path`.
Here, I have replaced that with `ArtifactAttr: {type, ext}` and used `Artifact` with path only.

src/OneExplorer/ConfigObject.ts
dayo09: I committed the change. 😄  https://github.com/Samsung/ONE-vscode/pull/1016#discussion_r919723520

jyoungyun (approved):
LGTM Thank you :)

llFreetimell (approved):

PR #1017, opened 2022-07-12
[CircleGraph] Use consistent return statement

llFreetimell:
`openView` function was returning two types,
one is `true` and the other is `void`.
This commit fixes this to use consistent return statement.

### DIFF
--- a/src/CircleGraph/CircleViewer.ts
+++ b/src/CircleGraph/CircleViewer.ts
@@ -83,7 +83,7 @@ class CircleViewerDocument implements vscode.CustomDocument {
         if (view.owner(panel)) {
           view.disposeGraphCtrl();
           this._circleViewer.splice(index, 1);
-          return true;  // break forEach
+          return;  // break forEach
         }
       });
     });
--- a/src/CircleGraph/CircleViewer.ts
+++ b/src/CircleGraph/CircleViewer.ts
@@ -83,7 +83,6 @@ class CircleViewerDocument implements vscode.CustomDocument {
         if (view.owner(panel)) {
           view.disposeGraphCtrl();
           this._circleViewer.splice(index, 1);
-          return;  // break forEach
         }
       });
     });


src/CircleGraph/CircleViewer.ts
```diff
@@ -83,7 +83,7 @@ class CircleViewerDocument implements vscode.CustomDocument {
         if (view.owner(panel)) {
           view.disposeGraphCtrl();
           this._circleViewer.splice(index, 1);
-          return true;  // break forEach
+          return;  // break forEach
```
seanshpark, line 86: return true here is to break forEach() loop.
please do not change this.

llFreetimell: Actually, both `return` and `return true;` has exactly same effect.
But, if you wanted to break `forEach`, both `return;` and `return true;` has no effect for `forEach`.
Please refer to following example.

```ts
const arr = [1, 2, 3, 4];
arr.forEach((val) => {
  console.log(val + " hello");
  if (val === 2) {
    console.log(val + " is caught");
    return true;
  }
  console.log(val + " bye");
});

/*
1 hello
1 bye
2 hello
2 is caught
3 hello
3 bye
4 hello
4 bye
*/
```

---

```ts
const arr = [1, 2, 3, 4];
arr.forEach((val) => {
  console.log(val + " hello");
  if (val === 2) {
    console.log(val + " is caught");
    return;
  }
  console.log(val + " bye");
});

/*
1 hello
1 bye
2 hello
2 is caught
3 hello
3 bye
4 hello
4 bye
*/
```

seanshpark: Hm...

seanshpark: change to `return` is useless for my intention of `return true` line;
(1) I'll revise using `forEach` to something else to break the loop or
(2) You can remove `return true;` line for now.

llFreetimell: If that line will be revised in the near future, (1) would be better :)
But if we cannot sure when (1) would be done, (2) would be better..!
(Just for me)

seanshpark: I cannot promis when `near future` would be so you can go with (2) if you like.

llFreetimell: I am good with (2) :)
Then, may I merge this PR and #1023 ?

seanshpark: > Then, may I merge this PR and https://github.com/Samsung/ONE-vscode/pull/1023 ?

Yes :)

llFreetimell: To do so, I need your approval..! :sweat_smile:

llFreetimell: LGTM for :heavy_check_mark: :)

seanshpark:
> LGTM for ✔️ :) 

Sorry for misunderstanding this... -_-;;;
What I wrote was: `(2) You can remove return true; line for now.`
that is, `return` itself is not needed so we can remove this line.

```diff
---          return true;  // break forEach
```

src/CircleGraph/CircleViewer.ts
seanshpark: > LGTM for ✔️ :) 

Sorry for misunderstanding this... -_-;;;
What I wrote was: `(2) You can remove return true; line for now.`
that is, `return` itself is not needed so we can remove this line.

```diff
-          return true;  // break forEach
```

seanshpark (approved):
LGTM thank you!

PR #1018, opened 2022-07-12
[Tests] Use forEach instead of map

llFreetimell:
`map(...)` function returns new array but it is not used.
This commit will fix to use `forEach` instead of `map`.


YongseopKim (approved):
LGTM 👍

dayo09 (approved):

PR #1019, opened 2022-07-12
[Execute] Use consistent return statement

llFreetimell:
`selectInputModel` and `pickBackend` function return two types,
one is `void` and the other is `undefined`.
This commit fixes to return `void` consistently.


YongseopKim (approved):
LGTM 👍

PR #1023, opened 2022-07-12
[PartEditor] Use consistent return statement

llFreetimell:
Some functions were returning two types,
one is `true` and the other is `void`.
This commit fixes this to use consistent return statement.

### DIFF
--- a/src/PartEditor/PartEditor.ts
+++ b/src/PartEditor/PartEditor.ts
@@ -396,7 +396,7 @@ export class PartEditorProvider implements vscode.CustomTextEditorProvider, Part
     this._partEditors.forEach((editor, index) => {
       if (e === editor) {
         this._partEditors.splice(index, 1);
-        return true;
+        return;
       }
     });
   }
--- a/src/PartEditor/PartEditor.ts
+++ b/src/PartEditor/PartEditor.ts
@@ -396,7 +396,6 @@ export class PartEditorProvider implements vscode.CustomTextEditorProvider, Part
     this._partEditors.forEach((editor, index) => {
       if (e === editor) {
         this._partEditors.splice(index, 1);
-        return;
       }
     });
   }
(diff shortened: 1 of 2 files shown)


src/PartEditor/PartEditor.ts
```diff
@@ -396,7 +396,7 @@ export class PartEditorProvider implements vscode.CustomTextEditorProvider, Part
     this._partEditors.forEach((editor, index) => {
       if (e === editor) {
         this._partEditors.splice(index, 1);
-        return true;
+        return;
```
seanshpark, line 399: `return true` here is to break `forEach()` loop.
please do not change this.

src/PartEditor/PartGraphSelector.ts
```diff
@@ -135,7 +135,7 @@ export class PartGraphSelPanel extends CircleGraphCtrl implements CircleGraphEve
     PartGraphSelPanel.panels.forEach((selpan) => {
       if (docPath === selpan._documentPath && id === selpan._ownerId) {
         result = selpan;
-        return true;  // break forEach
+        return;  // break forEach
```
seanshpark, line 138: ditto

src/PartEditor/PartGraphSelector.ts
```diff
@@ -224,7 +224,7 @@ export class PartGraphSelPanel extends CircleGraphCtrl implements CircleGraphEve
     PartGraphSelPanel.panels.forEach((selPan, index) => {
       if (this._documentPath === selPan._documentPath && this._ownerId === selPan._ownerId) {
         PartGraphSelPanel.panels.splice(index, 1);
-        return true;  // break forEach
+        return;  // break forEach
```
seanshpark, line 227: ditto

seanshpark (approved):
LGTM thank you!

PR #1031, opened 2022-07-13
[OneExplorer] Do not use useless return

llFreetimell:
`vscode.workspace.fs.delete` returns `Thenable<void>`, so `return` is useless.
This commit removes this.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -261,10 +261,10 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
         .then(ans => {
           if (ans === approval) {
             Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
-            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true});
+            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true})
+                .then(() => this.refresh());
           }
-        })
-        .then(() => this.refresh());
+        });
   }
 
   /**
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -261,8 +261,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
         .then(ans => {
           if (ans === approval) {
             Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
-            return vscode.workspace.fs.delete(
-                oneNode.node.uri, {recursive: recursive, useTrash: true});
+            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true});
           }
         })
         .then(() => this.refresh());


jyoungyun (approved):
LGTM

dayo09 (approved):

dayo09:
Q. Is it not recommended to use useless return? Because, for the code consistencies, most filesystem jobs returns promise. And I thought it's consistent to return the void promise, it's handy to add something else in the future. And some language recommends using return to make sure the function's exit point

llFreetimell:
> most filesystem jobs returns promise. And I thought it's consistent to return the void promise

Ah, I am sorry for confusion.
Original warning was inconsistent return statement.

```ts
    vscode.window.showInformationMessage(title, {detail: detail, modal: true}, approval)
        .then(ans => {
          if (ans === approval) {
            Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
            return vscode.workspace.fs.delete(
                oneNode.node.uri, {recursive: recursive, useTrash: true});
            // return Promise<void>
          }
          // return void
        })
        .then(() => this.refresh());
```

As you can see above, the lambda function returns both `void` and `Promise<void>`.
I was to fix this, but I think your question makes sense.
What should I change `// return void` to? :D

dayo09:
@llFreetimell Oh I thought it was returning from the most outer function. I read the change again and found out this change will make a bug. Let me explain in the code.

src/OneExplorer/OneExplorer.ts
```diff
@@ -261,8 +261,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
         .then(ans => {
           if (ans === approval) {
             Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
-            return vscode.workspace.fs.delete(
-                oneNode.node.uri, {recursive: recursive, useTrash: true});
+            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true});
```
dayo09, line 264: Please do not remove `return` here as it must return a promise to let the next then phrase know that the promise was done.

dayo09 (changes requested):

src/OneExplorer/OneExplorer.ts
```diff
@@ -261,8 +261,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
           }
```
dayo09, line 265: If you want to make the return type consistent, then it's better to return a resolved promise of void type outside the if statement.

src/OneExplorer/OneExplorer.ts
dayo09, line 264: OR, 

```suggestion
            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true}).then(()=>this.refresh());
```

This looks better then adding a resolved Promise<void>

dayo09 (approved):

llFreetimell:
@jyoungyun 

Since this PR is simple and original author validated this PR, I will merge this!

PR #1033, opened 2022-07-14
[CircleGraph] Remove no-effect return statement

llFreetimell:
This commit removes no-effect return statement.


llFreetimell:
This PR is similar with #1017 and #1023.
But I am not sure this PR keeps original intention :(
If there are something wrong, please let me know

seanshpark (approved):
LGTM

seanshpark:
As `return true` does not break the loop, removing them is OK. (I will not believe all internet articles.... -_-;;;). I'll change to loop using `for` later.
Thank you!

PR #1034, opened 2022-07-14
[CircleGraph] Remove no-effect return statement

llFreetimell:
This commit removes no-effect return statement.


llFreetimell:
Sorry but our analyzer does not find every warnings at once.... -_-;

PR #1035, opened 2022-07-14
Exclude external codes from static analysis

llFreetimell:
This commit excludes netron source codes from static analysis.


seanshpark (approved):
LGTM

PR #1037, opened 2022-07-14
[Spec] revisit sdbSpec

struss:
This commit revisit `sdbSpec` to use script file.
As `spawnsync()` with `shell: false` could not execute piped command, using `&&` or `|`.

### DIFF
--- a/res/script/sdbSpecList.sh
+++ b/res/script/sdbSpecList.sh
@@ -1,7 +1,0 @@
-#!/bin/sh
-
-#
-# This shell script print tizen device serial name on each line.
-#
-
-sdb devices | grep -v devices | grep device | awk '{print $1}'
--- a/res/script/sdbSpecList.sh
+++ b/res/script/sdbSpecList.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+#
+# This shell script print tizen device serial name on each line.
+#
+
+sdb devices | grep -v devices | grep device | awk '{print $1}'
--- a/script/sdbSpecList.sh
+++ b/script/sdbSpecList.sh
@@ -3,18 +3,20 @@
 #
 # This shell script prints tizen device serial name on each line.
 #
-# on `sdb devices`, it will show result below.
+# On `sdb devices`, it will show result below.
 #
 # List of devices attached 
 # 192.168.223.1:26101 	device    	0
+# emulator-1            emulator    0
 # 192.168.224.1:26101 	device    	0
 #
-# so, only imformation we need is list of -s in first column from second row to end.
-# just pass `sdb device` result except first line and grep only device and pring first column
-# this will show result below
+# First colmn is serialNumber which is unique key for device.
+# Second colmn is state.
+# Third colmn is targetName.
+# (Reference: https://docs.tizen.org/application/tizen-studio/common-tools/smart-development-bridge/#managing-targets)
 #
-# 192.168.223.1:26101
-# 192.168.224.1:26101
+# `sdb devices` will not give data that exactly we want, So we need to modify result.
+# We only need serialNumber, So use `awk` to pick serialNumber information.
 #
 
-sdb devices | awk 'NR > 1 {print}' | grep device | awk '{print $1}'
+sdb devices | awk 'NR > 1 {if ($2 == "device") print $1}'
--- a/script/sdbSpecList.sh
+++ b/script/sdbSpecList.sh
@@ -6,9 +6,9 @@
 # On `sdb devices`, it will show result below.
 #
 # List of devices attached 
-# 192.168.223.1:26101 	device    	0
-# emulator-1            emulator    0
-# 192.168.224.1:26101 	device    	0
+# 1.2.3.4:26101 	device    	0
+# emulator-1            emulator        0
+# 1.2.3.4:26101 	device    	0
 #
 # First colmn is serialNumber which is unique key for device.
 # Second colmn is state.
--- a/script/sdbSpecList.sh
+++ b/script/sdbSpecList.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+#
+# This shell script prints tizen device serial name on each line.
+#
+# on `sdb devices`, it will show result below.
+#
+# List of devices attached 
+# 192.168.223.1:26101 	device    	0
+# 192.168.224.1:26101 	device    	0
+#
+# so, only imformation we need is list of -s in first column from second row to end.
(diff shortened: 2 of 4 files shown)


struss:
um...little bit change to find `resDir`.

struss:
if we use script in `res/script`, then we could use `shell: false` process execute. but I little bit weird to use like this.

res/script/sdbSpecList.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+#
+# This shell script print tizen device serial name on each line.
+#
+
+sdb devices | grep -v devices | grep device | awk '{print $1}'
```
struss, line 7: is it okay for save this script on this position?

hyunsik-yoon: Currently `res` contains some test materials.
I guess that it would be possible to exclude `res` from files list to be packaged.

IMHO, how about `{workspace_folder}/script`?
Also we have to make sure that `{workspace_folder}/script` should be packaged.

res/script/sdbSpecList.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+#
+# This shell script print tizen device serial name on each line.
```
hyunsik-yoon, line 4:
```suggestion
# This shell script prints tizen device serial name on each line.
```

src/Backend/Spec.ts
```diff
@@ -70,8 +74,7 @@ class BridgeSpec {
 }
 
 // TODO add more BridgeSpec like docker or ADB......
-const sdbSpec = new BridgeSpec(
-    'sdb', 'sdb devices | grep -v devices | grep device | awk \'{print $1}\'', 'sdb shell');
+const sdbSpec = new BridgeSpec('sdb', vscode.Uri.joinPath(ext?.extensionUri, 'res', 'script', 'sdbSpecList.sh').fsPath, 'sdb shell');
```
hyunsik-yoon, line 77: If we change the `extensionId` in future, it would better to make this line fail by using `!`.

```suggestion
const sdbSpec = new BridgeSpec('sdb', vscode.Uri.joinPath(ext!.extensionUri, 'res', 'script', 'sdbSpecList.sh').fsPath, 'sdb shell');
```

src/Tests/Backend/Spec.test.ts
```diff
@@ -71,11 +72,12 @@ suite('Spec', function() {
     });
   });
   suite('sdbSpec', function() {
+    const extensionId = 'Samsung.one-vscode';
+    const ext = vscode.extensions.getExtension(extensionId) as vscode.Extension<any>;
+    const deviceListCom = vscode.Uri.joinPath(ext?.extensionUri, 'res', 'script', 'sdbSpecList.sh').fsPath;
```
hyunsik-yoon, line 77: ditto

res/script/sdbSpecList.sh
hyunsik-yoon: Can I ask a favor of you?
For readers of this code to understand this script, could you add comment showing what the output of `$ sdb devices` looks like ?

script/sdbSpecList.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# This shell script prints tizen device serial name on each line.
+#
+# on `sdb devices`, it will show result below.
+#
+# List of devices attached 
+# 192.168.223.1:26101 	device    	0
+# 192.168.224.1:26101 	device    	0
+#
+# so, only imformation we need is list of -s in first column from second row to end.
```
hyunsik-yoon, line 12: Could you explain what `-s` is?

script/sdbSpecList.sh
```diff
@@ -0,0 +1,20 @@
...
+# just pass `sdb device` result except first line and grep only device and pring first column
```
hyunsik-yoon, line 13: - I am not sure what `pring` is. `print`??
- Please use correct English grammar...
  - First word in statement starts with capital letter.
  - Please put `.` at the end of statement. (L12?). Since L13 does not ends with `.`, its hard to tell if L13 and L14 is one statement or two statements)
  - Please replace first `and` to `,`

script/sdbSpecList.sh
struss: um... I think I make some mistake on this comment. I will fix it. Sorry for bothering.

hyunsik-yoon (dismissed):
Thanks!

dayo09 (dismissed):

script/sdbSpecList.sh
```diff
@@ -0,0 +1,22 @@
...
+
+#
+# This shell script prints tizen device serial name on each line.
+#
+# On `sdb devices`, it will show result below.
+#
+# List of devices attached 
+# 192.168.223.1:26101 	device    	0
+# emulator-1            emulator    0
+# 192.168.224.1:26101 	device    	0
```
llFreetimell, line 11: Maybe little bit safer..?

```suggestion
# 1.2.3.4:26101 	device    	0
# emulator-1            emulator        0
# 1.2.3.4:26101 	device    	0
```

jyoungyun (approved):
LGTM

script/sdbSpecList.sh
struss: Oh I will fix it. Thanks!

script/sdbSpecList.sh
```diff
@@ -6,9 +6,9 @@
 # On `sdb devices`, it will show result below.
 #
 # List of devices attached 
-# 192.168.223.1:26101 	device    	0
-# emulator-1            emulator    0
-# 192.168.224.1:26101 	device    	0
+# 1.2.3.4:26101 	device    	0
```
hyunsik-yoon, line 9: I think previous IP address looks better for this example because `1.2.3.4` does  not look like IP addess. You also mentioned that this is serial number.. I am afraid that some readers might think that this is some kind of serial number format of sdb, not IP address.

hyunsik-yoon: Or you may want to mention that `the first column is some unique string such as serial or IP address`.

hyunsik-yoon (approved):

jyoungyun (dismissed):
LGTM

hyunsik-yoon (approved):

llFreetimell (approved):

jyoungyun (approved):
LGTM

dayo09 (approved):

struss:
@llFreetimell please merge this.

PR #1043, opened 2022-07-15
Provide welcome guide via ONE view container

yunjayh:
This commit provides welcome guide via ONE view container.
OneExplorerView shows the brief description about one-explorer-view,
and guide to open a workspace which includes model files and `cfg`s.
If no backend toolchain is registered, toolchainView's welcome page
leads the user to register or install toolchain.

---

1. The link to our document will be updated after introducing better README. (since json doesn't support comment, I couldn't add todo)
2. Toolchain timing issue (https://github.com/Samsung/ONE-vscode/issues/932#issuecomment-1181320311) will be resolved in near future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,16 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+		"viewsWelcome": [
+      {
+        "view": "OneExplorerView",
+        "contents": "Open folder which includes models and configuration files.\n[Open Folder](Command:workbench.action.files.openFolder)\nFollowing types of models will be displayed.\n- *.tflite, *pb and *.onnx\n- Additional file types which are described by backend.\n\nTo learn more about ONE-vscode extension usage [read our docs](https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md)."
+      },
+      {
+        "view": "ToolchainView",
+        "contents": "Register already installed toolchain or install a new toolchain.\n[Register Toolchain](Command:one.toolchain.refresh)\n\n[Install Toolchain](Command:one.toolchain.install)"
+      }
+    ],
     "viewsContainers": {
       "activitybar": [
         {
--- a/package.json
+++ b/package.json
@@ -19,11 +19,7 @@
 		"viewsWelcome": [
       {
         "view": "OneExplorerView",
-        "contents": "Open folder which includes models and configuration files.\n[Open Folder](Command:workbench.action.files.openFolder)\nFollowing types of models will be displayed.\n- *.tflite, *pb and *.onnx\n- Additional file types which are described by backend.\n\nTo learn more about ONE-vscode extension usage [read our docs](https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md)."
-      },
-      {
-        "view": "ToolchainView",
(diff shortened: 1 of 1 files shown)


yunjayh:
image

dayo09:
### GIF
0715-welcome


Works well, thanks!

dayo09 (approved):

package.json
```diff
@@ -16,6 +16,16 @@
   ],
   "main": "./out/extension.js",
   "contributes": {
+		"viewsWelcome": [
+      {
+        "view": "OneExplorerView",
+        "contents": "Open folder which includes models and configuration files.\n[Open Folder](Command:workbench.action.files.openFolder)\nFollowing types of models will be displayed.\n- *.tflite, *pb and *.onnx\n- Additional file types which are described by backend.\n\nTo learn more about ONE-vscode extension usage [read our docs](https://github.com/Samsung/ONE-vscode/blob/main/docs/HowToUse.md)."
```
hyunsik-yoon, line 22: `Additional file types which are described by backend.` 
- I am not 100% sure what this sentence means.

Can we write in more specific?

package.json
```diff
@@ -16,6 +16,16 @@
...
+      },
+      {
+        "view": "ToolchainView",
+        "contents": "Register already installed toolchain or install a new toolchain.\n[Register Toolchain](Command:one.toolchain.refresh)\n\n[Install Toolchain](Command:one.toolchain.install)"
```
hyunsik-yoon, line 26: I am not sure if we use this term: 'Register toolchain'.

I think you're working on timing issue problem. You may need to change this sentence after fixing timing issue.
Instead of adding ToolchainView welcome vew now and changing again later, how about adding ToolchainView welcome vew _after_ fixing the timing issue?

yunjayh: 2 sounds nice to me. I'll update this pr.

hyunsik-yoon:
One question occurred to me.
What happen if user already opened a folder but the folder does not have any cfg or models?
In that case, I wonder if this welcome view is shown or not.

IMHO, there are some cases:
1. user didn't open any folder -> show this welcome view
2. user opened a folder but the folder does not have model or cfg. 
    A. show some notification message, e.g., "The folder you trying to open does not have model or cfg files.. blah..."
    B and show this welcome view.

2-A is not a scope of this PR but it is one feature we could implement later.

hyunsik-yoon (approved):

seanshpark:
(Just a comment)
3. when the `.cfg` is not our format...

yunjayh:
> What happen if user already opened a folder but the folder does not have any cfg or models? In that case, I wonder if this welcome view is shown or not.

In this case, this welcome view is opened. (vscode welcome view is shown when the tree has no children) I'm not sure how to make different welcome view for each cases, but I agree with your 2B idea. :)



> 3. when the `.cfg` is not our format...

Hmm.. AFAIK it is shown in `one-explorer-view` and opened with cfg editor. However, since it cannot be parsed as ini format, it's same with the empty cfg file. (@llFreetimell might be able to give better answer)

yunjayh:
@dayo09 PTAL :smile:

dayo09 (approved):
LGTM, Thanks!

PR #1045, opened 2022-07-15
[DeviceAPI] Add ExecutionManager

struss:
This commit will add `ExecuteManager` to handle `Device` and `Executor`.

### DIFF
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -40,7 +40,7 @@ class Device {
     this.availableExecutor = new Set<Executor>();
   }
 
-  refresh(executorList: Executor[]): void {
+  checkAvailExecutor(executorList: Executor[]): void {
     for (const executor of executorList) {
       if (executor.require().satisfied(this.spec)) {
         this.availableExecutor.add(executor);
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Executor } from '../Backend/Executor';
+import {Executor} from '../Backend/Executor';
 import {DeviceSpec} from '../Backend/Spec';
 
 /**
@@ -39,8 +39,8 @@ class Device {
     this.spec = spec;
     this.availableExecutor = new Set<Executor>();
   }
-  
-  refresh(executorList: Executor[]): void{
+
+  refresh(executorList: Executor[]): void {
     for (const executor of executorList) {
       if (executor.require().satisfied(this.spec)) {
         this.availableExecutor.add(executor);
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -33,17 +33,18 @@ import {DeviceSpec} from '../Backend/Spec';
 class Device {
   name: string;
   spec: DeviceSpec;
-  availableExecutor: Set<Executor>;
+  availableExecutors: Set<Executor>;
   constructor(name: string, spec: DeviceSpec) {
     this.name = name;
     this.spec = spec;
-    this.availableExecutor = new Set<Executor>();
+    this.availableExecutors = new Set<Executor>();
   }
 
-  checkAvailExecutor(executorList: Executor[]): void {
+  // Register only available executor.
+  registerExecutor(executorList: Executor[]): void {
     for (const executor of executorList) {
       if (executor.require().satisfied(this.spec)) {
-        this.availableExecutor.add(executor);
+        this.availableExecutors.add(executor);
       }
     }
   }
--- a/src/Execute/Device.ts
+++ b/src/Execute/Device.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { Executor } from '../Backend/Executor';
 import {DeviceSpec} from '../Backend/Spec';
 
 /**
@@ -32,10 +33,20 @@ import {DeviceSpec} from '../Backend/Spec';
 class Device {
   name: string;
   spec: DeviceSpec;
+  availableExecutor: Set<Executor>;
   constructor(name: string, spec: DeviceSpec) {
     this.name = name;
     this.spec = spec;
+    this.availableExecutor = new Set<Executor>();
   }
-};
+  
+  refresh(executorList: Executor[]): void{
+    for (const executor of executorList) {
+      if (executor.require().satisfied(this.spec)) {
+        this.availableExecutor.add(executor);
+      }
+    }
+  }
+}
 
 export {Device};
--- a/src/Execute/ExecutionManager.ts
+++ b/src/Execute/ExecutionManager.ts
@@ -22,7 +22,7 @@ class ExecutionManager {
   constructor(devices: Device[]) {
     this.allDevices = devices;
     for (let device of this.allDevices) {
-      device.refresh(globalExecutorArray);
+      device.checkAvailExecutor(globalExecutorArray);
     }
   }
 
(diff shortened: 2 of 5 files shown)


dayo09:
Shouldn't PR Title be..?
```[DeviceAPI] Add ExecuteManager``` -> ```[DeviceAPI] Add ExecutionManager```

File: src/Execute/ExecutionManager.ts [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+ */
+import {globalExecutorArray} from '../Backend/Backend';
+import {Executor} from '../Backend/Executor';
+import {Device} from './Device';
+
+interface DeviceExecutorSet {
+  [key: string]: Set<Executor>;
+}
+
+class ExecutionManager {
```
dayo09, line 24: Q. How about 'ExecutorManager' here?

src/Execute/ExecutionManager.ts
```diff
@@ -0,0 +1,60 @@
...
+  allDevices: Device[];
+  availableDevices: DeviceExecutorSet;
```
dayo09, line 26: So it seems Executor is  'Available Device'

If so, what about defining the term 'executor' and rename `availableDevices` to `executors`? For the term's consistency

src/Execute/ExecutionManager.ts
struss: hmmm It will handle both `Executor` and `Device`, which both related to on `Execution`. so I name like this.

src/Execute/ExecutionManager.ts
struss: um..... re-consider about your comment, that object have relation about device and executor. so, I think change this to `availableExecutor` would be more nice, I think?

src/Execute/ExecutionManager.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(devices: Device[]) {
+    this.allDevices = devices;
+    this.availableDevices = {};
+    for (let device of this.allDevices) {
+      this.availableDevices[device.name] = new Set<Executor>();
```
dayo09, line 32: Q. how about 'Device' class to include 'Exectuor set'..?

src/Execute/ExecutionManager.ts
dayo09, line 26: The property name 'availableDevices' is expected to be Device[].

How about making it as Device[] and [put the DeviceExecutorSet under the Device class..?](https://github.com/Samsung/ONE-vscode/pull/1045/files#r921866023..?) (I am not sure if that is possible under the current structure as I don't know about this well, so I just suggest)

src/Execute/ExecutionManager.ts
struss: um.... looks nice. I will fix it.

src/Execute/ExecutionManager.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  findDevice(name: string): Device|undefined {
+    for (const device of this.allDevices) {
+      if (device.name === name) {
+        return device;
+      }
+    }
+    return undefined;
```
dayo09, line 43: You can do this like

return this.allDevices.find(device=>{
return device.name === name");

It's equivalent.

dayo09, line 43: You can do this like

```suggestion
return this.allDevices.find(device=>{ return device.name === name; });
```

It's equivalent.

src/Execute/Device.ts
```diff
@@ -32,10 +33,20 @@ import {DeviceSpec} from '../Backend/Spec';
 class Device {
   name: string;
   spec: DeviceSpec;
+  availableExecutor: Set<Executor>;
```
dayo09, line 36: If there is no 'unavailable executor', why not naming it as 'executors'?

src/Execute/ExecutionManager.ts
dayo09: Now that 'Device' class has 'Executors' as a member, it can be a DeviceManager. :-D

src/Execute/ExecutionManager.ts
```diff
@@ -0,0 +1,41 @@
...
+    return this.allDevices.find(device => {
+      return device.name === name;
+    });
+  }
+}
+
+// This ExecutionManagerMap will be used on Provider.
+interface ExecutionManagerMap {
+  [key: string]: ExecutionManager;
+}
```
dayo09, line 39: 'used on Provider' means that it will be used somewhere outside? If so, what about moving the interface there? Moreover, as the interface only has a map of ExecutionManager, it'll be more explicit if you defined the map there.

src/Execute/Device.ts
```diff
@@ -32,10 +33,20 @@ import {DeviceSpec} from '../Backend/Spec';
...
   constructor(name: string, spec: DeviceSpec) {
     this.name = name;
     this.spec = spec;
+    this.availableExecutor = new Set<Executor>();
   }
-};
+
+  checkAvailExecutor(executorList: Executor[]): void {
```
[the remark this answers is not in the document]
dayo09: I think if you are gonna use the member variable name as `executors`, `registerExecutors` will be find and you can add some comment like
```
// Add only available executors
```

src/Execute/ExecutionManager.ts
struss: I will fix it. Thanks.

src/Execute/Device.ts
struss: > If there is no 'unavailable executor'

That's right. but I want to mark those `Executor' could run on this `Device`, so I name like that.

src/Execute/Device.ts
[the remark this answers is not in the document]
hyunsik-yoon: `setExecutors` also seems commonly-used name for such method.

src/Execute/Device.ts
struss: um...I think keep this on would be nice. if `Executor` which not `satisfied()` Device's Spec means `unavailable executor`. so, I think available will be a minimum expression for those `executors`.

src/Execute/ExecutionManager.ts
hyunsik-yoon: Could you write a comment explaining what kind of string will be allowed as `key`?

hyunsik-yoon:
Only with this code, it was somewhat difficult for me to understand how ExecutionManager and ExecutionManagerMap will be used.
So I tried to find draft but I couldn't. Mentioning the draft number could help reviewer to understand the context around this code.

src/Execute/ExecutionManager.ts
struss: > Could you write a comment explaining what kind of string will be allowed as `key`?

Oh I will comment this on next PR update.

struss:
I will update this and ping to all.

src/Execute/ExecutionManager.ts
struss: > Could you write a comment explaining what kind of string will be allowed as `key`?

key will be a location of manager. for example) host on local or IP for some other host.

struss:
> Only with this code, it was somewhat difficult for me to understand how ExecutionManager and ExecutionManagerMap will be used. So I tried to find draft but I couldn't. Mentioning the draft number could help reviewer to understand the context around this code.

Sorry For late. I think I forget to create Draft before this. I update PR #1063 for how to use this on `DeviceViewProvider` with `ExecutionManager` and others.

src/Execute/ExecutionManager.ts
struss: um.... I think this implementation only handle about Device on this manager, so I accept `DeviceManager`. I will fix this.

hyunsik-yoon (approved):

struss:
@dayo09 PTAL

dayo09:
LGTM, Thanks!

Issue #1049, opened 2022-07-17
[Backend] Exchange the file names: API.ts and Backend.ts


dayo09:
## What?

In Backend/Backend.ts,
```
function backendRegistrationApi() {
```

In Backend/API.ts,

```
export interface Backend {
  // backend's name. this doesn't mean the name of the toolchain
  name(): string;

  // compiler specs by being filled by impl
  compiler(): Compiler|undefined;

  // executor specs by being filled by impl
  executor(): Executor|undefined;
}
```

It sounds wierd that the file `Backend` has only an API, while the file `API` has only Backend definition.

Let's find out the behind story and exchange the file names.

hyunsik-yoon:
Currently we provodes two types of API:
- API.ts defines API of which name is `Backend`, which is implemented by backend extension and called by one-vscode.
- Backend.ts defined _registration API_, which is defined in one-vscode and called by backend extension.

Some ideas I could imagine are..

- API.ts -> Backend.ts
  - interface API  // or more specific name
- Backend.ts -> BackendRegistration.ts
  - backendRegistrationApi // or just an API ?

or 

- API.ts -> BackendAPI.ts ?
  - interface API -> interface BackendAPI
- Backend.ts -> RegistrationAPI.ts
  - backendRegistrationApi

Or could we combine those two API into one file, API.ts?

dayo09:
@hyunsik-yoon If we go with that approach, then the `Compiler.ts` and `Executor.ts` are also considered to be APIs. And I also plan to add more file like them.

I thought that most files in `Backend.ts` are those which must be implemented by backends. From the point of view, we can just leave them as is. 😃

dayo09:
Done

PR #1050, opened 2022-07-18
[Backend] Exchange the file names: API, Backend

dayo09:
This commit exchagnes the file names of
Backend/API.ts and Backend/Backend.ts

----

From #1049

### DIFF
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -14,28 +14,49 @@
  * limitations under the License.
  */
 
-import {Compiler} from './Compiler';
+const assert = require('assert');
+import {Backend} from './Backend';
+import {gToolchainEnvMap, ToolchainEnv} from '../Toolchain/ToolchainEnv';
+import {Logger} from '../Utils/Logger';
 import {Executor} from './Executor';
 
 /**
- * This file defines common API exposed by backend extension.
- *
- * Each backend extensions should implement Backend interface.
- * Through registration API of ONE-vscode extension, those backend extensions
- * needs to register implementation of Backend interface.
- *
- * TODO ONE-vscode and backend extensions have copies of this file. Check if this is really OK.
+ * Interface of backend map
+ * - Use Object class to use the only string key
  */
-
-// ** The scope of Backend is defined by each backend supporter **
-// A kind of proxy. Backend doesn't know where it-self is (local? remote? it doesn't know.)
-export interface Backend {
-  // backend's name. this doesn't mean the name of the toolchain
-  name(): string;
-
-  // compiler specs by being filled by impl
-  compiler(): Compiler|undefined;
-
-  // executor specs by being filled by impl
-  executor(): Executor|undefined;
+interface BackendMap {
+  [key: string]: Backend;
 }
+
+// List of backend extensions registered
+let globalBackendMap: BackendMap = {};
+// List of Executor extensions registered
+let globalExecutorArray: Executor[] = [];
+
+function backendRegistrationApi() {
+  const logTag = 'backendRegistrationApi';
+  let registrationAPI = {
+    registerBackend(backend: Backend) {
+      const backendName = backend.name();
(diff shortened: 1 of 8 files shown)


dayo09:
NOTE: The current commit runs the toolchain well with the current backend extension.

hyunsik-yoon:
`interface Backend` in `backend.ts` seems good to me.
but the name `API.ts` (having `registerBackendAPI`) seems too broad. How about something like `Registration.ts`?

dayo09:
@hyunsik-yoon I think the name is good as is. We may add other kind of APIs in the future. We may leave it as APIs and use it as API set of our ONE-vscode extension.

hyunsik-yoon:
Related to communication between backend and one-vscode,
I think that we currently have two types of API.

- `interface Backend `
  - implemented by backend extension and called by one-vscode
- `backendRegistrationAPI`
  - implemented by one-vscode and called by backend extension 

IMHO, if you want to use file name `API.ts`, I think it would be better to put `interface Backend` into `API.ts` too since they all API.

If we want to keep two separate files,
it would be better to have more specific and consistent naming, e.g., both file name would ends with `API`. Or both file wouldn't ends with `API`.

dayo09:
@hyunsik-yoon 

If we go with that approach, then the `Compiler.ts` and `Executor.ts` are also considered to be APIs. And I also plan to add more file like them.

I thought that most files in `Backend.ts` are those which must be implemented by backends. From the point of view, we can just leave them as is. 😃

(copied from https://github.com/Samsung/ONE-vscode/issues/1049#issuecomment-1186762154)

Or we can put them as is and `export` them from the API.ts file.

dayo09:
@hyunsik-yoon Or, how about [this](https://github.sec.samsung.net/RS7-RuntimeNTools/ONE-vscode-TRIV/pull/92) approach?

hyunsik-yoon:
The discussion is very confusing to me since there are two files and two times (current and future) involved.
I'd like to suggest not to use pronoun if available and use more specif terms. :-) 

Let me clarify your answer... again.

If this PR is merged, we will have the following structure:

- API.ts
  - backendRegistrationAPI
- Backend.ts
  - inteface Backend
- Compiler.ts
- ....

You mentioned,

> I think the name is good as is. We may introduce other kind(non-registration) of APIs in the future. We may leave it as APIs and use it as API set of our ONE-vscode extension.

Here, I guess you meant API.ts by _the name_.

>  I also plan to add more file like them.

I have one question here. If you want ot introduce, e.g., `File` class, I wonder where `File` would exist in the above structure. Would it be inside API.ts ? or File.ts?

dayo09:
> Here, I guess you meant API.ts by the name.

Right :-D

> I have one question here. If you want ot introduce, e.g., File class, I wonder where File would exist in the above structure. Would it be inside API.ts ? or File.ts?

I will make a `File.ts`(for example) and make a proper api function in `API.ts`. (as you said, kinda registration functions) And if I proceed to make `BackendManager` class, I think I can wrap those API registrations somehow with BackendManager class functions.

jyoungyun:
I like your changes. :) If you ready to merge this PR, please let me know it.

hyunsik-yoon:
@dayo09 
I understand. Maybe we can check this name again when you finish your planned work in future but I am OK with this change. :-)

hyunsik-yoon (approved):

hyunsik-yoon:
@dayo09 

> @hyunsik-yoon Furthermore, how about [this](https://github.sec.samsung.net/RS7-RuntimeNTools/ONE-vscode-TRIV/pull/92) approach?

Great. It's amazing that `package.json` has such feature. If this approach work, it would be much better to synchronize two repos. 👍

dayo09:
@jyoungyun It's ready. PTAL :-D!

jyoungyun (approved):
LGTM

PR #1057, opened 2022-07-19
[Tests] Add unit tests for Nonce

seanshpark:
This will add unit tests for Nonce in Utils.

### DIFF
--- a/src/Tests/Utils/external/Nonce.test.ts
+++ b/src/Tests/Utils/external/Nonce.test.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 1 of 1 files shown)


llFreetimell (approved):
LGTM

hyunsik-yoon (approved):

hyunsik-yoon:
@llFreetimell I wonder if we need to test files in `external`.

seanshpark:
will wait for confirm if no need to test `external`; but will need to excluded `external`s

llFreetimell:
I confirmed that test is not needed for 100% external codes.
Instead, as @seanshpark said, `external`s should be excluded from coverage report :)
I will exclude soon!

seanshpark:
closing as not needed :)

PR #1065, opened 2022-07-19
[Mondrian] Update status line info format

ai-moiseev:
This commit modifies status line output format with the following upgrades:

* Trim very long origin names to the maximum of 32 characters
* Display block lifetime using `alive_from -> alive_till (total size)` format.

### DIFF
--- a/media/Mondrian/mondrianViewer.js
+++ b/media/Mondrian/mondrianViewer.js
@@ -290,9 +290,12 @@ function updateViewport(data, viewer) {
     box.style.right = ((viewport.cycles - alloc.alive_till) / viewport.cycles * 100) + '%';
     box.style.backgroundColor = boxColors[i % boxColors.length];
     box.addEventListener('mouseover', (event) => {
-      statusLineContainer.innerHTML =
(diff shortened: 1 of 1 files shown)


hyunsik-yoon (approved):

hyunsik-yoon:
I like this PR.
One little concern is,
what if lengths of most nodes are very long and the first 32 chars of nodes are almost same.
I saw such situation when I worked on circle models from some Tensorflow language models.

llFreetimell (approved):

llFreetimell:
I will wait the response for https://github.com/Samsung/ONE-vscode/pull/1065#issuecomment-1189706321 :)

ai-moiseev:
@hyunsik-yoon @llFreetimell The idea is to introduce a new UI element, a side panel accessible on node click to display all of the node metadata in a similar way that Netron does:

image

I can implement this UI and merge it first so this PR will not break current usage for longer origins. On-click side panel has several advantages over an on-hover status bar - it allows to select and copy values, has enough space for origin of any length (the need for value trimming in the status bar came after implementation of backend origin info exporter, where nodes contained origins with 1000+ symbol length).

hyunsik-yoon:
@ai-moiseev Thanks for the explanation. I am looking forward to seeing the new UI!

PR #1068, opened 2022-07-20
[CircleGraph] Revise CircleViewerDocument

seanshpark:
This will revise CircleViewerDocument as export class and
openView method to return view object for the document.


seanshpark:
from #966

PR #1069, opened 2022-07-20
[OneExplorer] Use default value for function parameter

llFreetimell:
`canHide` is `boolean` and thus `false` has same effect with `canHide?`.
Let's use default value for function parameter to be more evident.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -220,18 +220,25 @@ class BaseModelNode extends Node {
   readonly type = NodeType.baseModel;
 
   // Do not open file as default
-  openViewType: string|undefined = undefined;
+  static defaultOpenViewType = undefined;
   // Display 'symbol-variable' icon to represent model file as default
-  icon = new vscode.ThemeIcon('symbol-variable');
+  static defaultIcon = new vscode.ThemeIcon('symbol-variable');
   // Show file always as default
-  canHide = false;
+  static defaultCanHide = false;
 
-  constructor(uri: vscode.Uri, openViewType?: string, icon?: vscode.ThemeIcon, canHide?: boolean) {
+  openViewType: string|undefined = BaseModelNode.defaultOpenViewType;
+  icon = BaseModelNode.defaultIcon;
+  canHide = BaseModelNode.defaultCanHide;
+
+  constructor(
+      uri: vscode.Uri, openViewType: string|undefined = BaseModelNode.defaultOpenViewType,
+      icon: vscode.ThemeIcon = BaseModelNode.defaultIcon,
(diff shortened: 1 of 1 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -309,7 +309,8 @@ class ProductNode extends Node {
   // Show file always as default
   canHide = false;
 
-  constructor(uri: vscode.Uri, openViewType?: string, icon?: vscode.ThemeIcon, canHide?: boolean) {
+  constructor(
+      uri: vscode.Uri, openViewType?: string, icon?: vscode.ThemeIcon, canHide: boolean = false) {
```
dayo09, line 313: Actually I have revisited the code lately. It's the code which was.

Because the above `canHide = false;`(line 310) already sets the default value, I removed the default parameter here. Otherwise, there will be two default values.

dayo09: Imagine that you give another value than `false` within this constructor as a default parameter. Then, the class's default value will be gone. It's quiet confusing and it needs devs to make in sync, which requires effort to maintain..

llFreetimell: If so, how about setting default value only in constructor like following?

```ts
class ProductNode extends Node {
  readonly type = NodeType.product;

  // Do not open file as default
  openViewType: string|undefined;
  // Display file icon as default
  icon: any;
  // Show file always as default
  canHide: boolean;

  constructor(uri: vscode.Uri, openViewType?: string, icon: vscode.ThemeIcon = vscode.ThemeIcon.File, canHide: boolean = false) {
    super(uri);

    this.openViewType = openViewType;
    this.icon = icon;
    this.canHide = canHide;
  }

  buildChildren = (): void => {
    // Do nothing
  };
}
```

dayo09: I think that the comments are way easier to read if the default values are near.

```
  // Do not open file as default
  openViewType: string|undefined;
  // Display file icon as default
  icon: any;
  // Show file always as default
  canHide: boolean;
```

For example, if I read the code for the first time, I'd be frustrated as the default values are not their while the comments explains about something. Then I'd try finding the values and they will need to follow down to the constructor.

llFreetimell: I agree with you that original codes is good but I must find a solution for https://rules.sonarsource.com/typescript/RSPEC-4798 :(
Could you give me some clue for this..?

dayo09: In that case, I suggest declaring default values

llFreetimell: Does it mean that you withdraw your objection...?

dayo09 (approved):
Thanks!

PR #1072, opened 2022-07-20
[res] Sample circle model for unit testing

seanshpark:
This will introduce sample circle model for CircleGraph unit testing.


seanshpark:
from #966

hyunsik-yoon (approved):

PR #1075, opened 2022-07-20
[CircleGraph/test] Initial test for CircleGraphPanel

seanshpark:
This will introduce initial unit test for CircleGraphPanel class.

### DIFF
--- a/src/Tests/CircleGraph/CircleGraph.test.ts
+++ b/src/Tests/CircleGraph/CircleGraph.test.ts
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {assert} from 'chai';
+import * as vscode from 'vscode';
+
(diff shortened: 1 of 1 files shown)


seanshpark:
from #966

llFreetimell:
Thanks for creating tests for `CircleGraph` ! :D
But I have a question, I cannot find any `assert` in the `CircleGraph.test.ts`.
Was it intentional?

seanshpark:
> I cannot find any assert in the CircleGraph.test.ts.
> Was it intentional?

Mostly there is no return value to test. some exist with condition.
Methods to test are mostly simple logic.
If no assert() is a problem from SE feedback (1) add dummy `assert(true)` (2) refactor later to get some return values; I'm not sure how this is possible for now.

llFreetimell:
> If no assert() is a problem from SE feedback

Unfortunately, when I checked this with static analysis tool, it reported that there is no assertion in test code :(
But refactoring may bother you and may take some time.
So how about just adding following simple assertions for now? :)

```ts
  suite('#createOrShow', function() {
    test('createOrShow model with no error ', function() {
      // NOTE createOrShow(extensionUri, undefied) will show file open dialog
      // which test doen't execute as expected
      const panel = CircleGraphPanel.createOrShow(extensionUri, modelToLoad);
      assert.notStrictEqual(panel, undefined);
      panel?.dispose();
      const panel2 = CircleGraphPanel.createOrShow(extensionUri, modelToLoad);
      assert.notStrictEqual(panel2, undefined);
    });
  });
  suite('#isReady', function() {
    test('isReady check', function() {
      const panel = CircleGraphPanel.createOrShow(extensionUri, modelToLoad);
      assert.notStrictEqual(panel, undefined);
      assert.strictEqual(panel?.isReady(), false);
      panel?.dispose();
    });
  });
  ...
  suite('#setPartition', function() {
    test('setPartition check', function() {
      const panel = CircleGraphPanel.createOrShow(extensionUri, modelToLoad);
      assert.notStrictEqual(panel, undefined);
      panel?.setPartition({a: 'B'});
      panel?.dispose();
      assert.ok(true);
    });
  });
```

llFreetimell (approved):
LGTM!

hyunsik-yoon (approved):

PR #1077, opened 2022-07-20
[CircleGraph] Use consistent return statement

llFreetimell:
lambda function in `createOrShow` returns both `void` and `CircleGraphPanel`.
This commit fixes to return `undefined` instead of `void`.


seanshpark (approved):
LGTM thank you!

PR #1078, opened 2022-07-21
[CircleGraph/test] Initial test for CircleGraphCtrl

seanshpark:
This will introduce initial unit test for CircleGraphCtrl class.

### DIFF
--- a/src/Tests/CircleGraph/CircleGraphCtrl.test.ts
+++ b/src/Tests/CircleGraph/CircleGraphCtrl.test.ts
@@ -1,0 +1,198 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


seanshpark:
from #966

hyunsik-yoon (approved):

llFreetimell:
Nowadays, we are too hard to create so many negative TCs.
So I was thinking following things
- Integrating some positive TCs which are just copy & pasted test codes.
- Excluding source codes from TC analysis which are too hard to validate by TCs because of UI, message communication, etc.

I am checking second option for now.
If second option is accepted, this PR can be closed.
Otherwise, I will ask you to integrate positive TCs into one.

Please wait a moment :)

llFreetimell:
Since #1082 is merged, could you determine whether this PR will be closed or we may go as it is? :)

I am not against with merging this PR. Actually many TC is always right :D
But just worried about "Can negative TC be created...?"

seanshpark:
> could you determine whether this PR will be closed or we may go as it is? :)

I'll check with test coverage :)

PR #1094, opened 2022-07-25
[Backend] revisit `Backend` API to have multiple Executor

struss:
This commit will update `Backend` API to have multiple `Executor`s.
This commit will also update some test for this.

### DIFF
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -44,15 +44,11 @@ function backendRegistrationApi() {
       if (compiler) {
         gToolchainEnvMap[backend.name()] = new ToolchainEnv(compiler);
       }
-      const executor = backend.executor();
-      if (executor) {
-        globalExecutorArray.push(executor);
+      const executors = backend.executors();
+      if (executors) {
+        globalExecutorArray.push(...executors);
       }
       Logger.info(logTag, 'Backend', backendName, 'was registered into ONE-vscode.');
-    },
-    registerExecutor(executor: Executor) {
-      globalExecutorArray.push(executor);
-      Logger.info(logTag, 'Executor', executor.name(), 'was registered into ONE-vscode.');
     }
   };
 
--- a/src/Backend/Backend.ts
+++ b/src/Backend/Backend.ts
@@ -37,5 +37,5 @@ export interface Backend {
   compiler(): Compiler|undefined;
 
   // executor specs by being filled by impl
-  executor(): Executor|undefined;
+  executors(): Executor[]|undefined;
 }
--- a/src/Tests/Backend/Backend.test.ts
+++ b/src/Tests/Backend/Backend.test.ts
@@ -25,6 +25,10 @@ import {gToolchainEnvMap} from '../../Toolchain/ToolchainEnv';
 // TODO: Move it to Mockup
 const backendName = 'Mockup';
 class BackendMockup implements Backend {
+  executor: ExecutorMockup;
+  constructor(executor: ExecutorMockup) {
+    this.executor = executor;
+  }
   name(): string {
     return backendName;
   }
@@ -32,8 +36,8 @@ class BackendMockup implements Backend {
     return new CompilerBase();
   }
 
-  executor(): Executor|undefined {
-    return new ExecutorBase();
+  executors(): Executor[]|undefined {
+    return [this.executor];
   }
 }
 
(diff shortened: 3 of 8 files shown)


src/Backend/API.ts
```diff
@@ -44,15 +44,11 @@ function backendRegistrationApi() {
       if (compiler) {
         gToolchainEnvMap[backend.name()] = new ToolchainEnv(compiler);
       }
-      const executor = backend.executor();
-      if (executor) {
-        globalExecutorArray.push(executor);
+      const executors = backend.executors();
+      if (executors) {
+        globalExecutorArray.push(...executors);
```
struss, line 49: This array could be removed. but current PR not removed. this will remove next PR about `DeviceView`.

yunjayh: Q) Is it ready to modify our backend extension repo? 
IMHO Revisiting inference and execute quickinputs and runner doesn't matter for the other developers, but failing to activate extensions could be critical. 
How about remaining this part as is, and just adding `executors` to `Backend.ts`? Since renaming to `executors` causes error while activating the extensions. It might be short time, but someone who rebased main branch will be confused from the error.

yunjayh: Hmm.. I thought these lines could be removed, but you commented `this will remove next PR about DeviceView`. I think I couldn't clear understand what that means. Can you give me a little more details?

src/Backend/Backend.ts
```diff
@@ -37,5 +37,5 @@ export interface Backend {
   compiler(): Compiler|undefined;
 
   // executor specs by being filled by impl
-  executor(): Executor|undefined;
+  executors(): Executor[]|undefined;
```
dayo09, line 40: Should it include `undefined`? As it's an array, we can replace the use case of `undefined` with an empty array.

src/Tests/Execute/InferenceRunner.test.ts
```diff
@@ -88,16 +93,18 @@ suite('Execute', function() {
...
         assert.strictEqual(runner.modelPath, modelPath);
         assert.strictEqual(runner.inputSpec, inputSpec);
       });
     });
 
     suite('#getInferenceCmd()', function() {
       test('gets inference command', function() {
-        let runner = new InferenceRunner(backend, modelPath, inputSpec);
+        let runner = new InferenceRunner(
+            backend, (backend.executors() as Executor[])[0], modelPath, inputSpec);
```
dayo09, line 107: Related to https://github.com/Samsung/ONE-vscode/pull/1094/files#r928556316,

(backend.executors() as Executor[])[0] looks a bit unnatural.

dayo09: It looks unnatural to test only the first executor.
How about running the inference command for all the registered executors from the code, for example?

src/Tests/Backend/Backend.test.ts
```diff
@@ -25,15 +25,19 @@ import {gToolchainEnvMap} from '../../Toolchain/ToolchainEnv';
...
   name(): string {
     return backendName;
   }
   compiler(): Compiler|undefined {
     return new CompilerBase();
   }
 
-  executor(): Executor|undefined {
-    return new ExecutorBase();
+  executors(): Executor[]|undefined {
```
dayo09, line 39: ditto

src/Backend/API.ts
dayo09: rr> `How about remaining this part as is, and just adding executors to Backend.ts?` 

I agree here.

More specifically,
You can (1) first add 'executors' and, (2) replace the internal logic of `executor` with that of `executors`, add 'TODO Remove executor' comment, (3) then finally remove executor.

struss: > Hmm.. I thought these lines could be removed, but you commented `this will remove next PR about DeviceView`. I think I couldn't clear understand what that means. Can you give me a little more details?

Oh, I write this to remind me to check this. but I think this make you more confused. I will edit this comment as

```
This array could be removed as we can access all `Executor`s using `globalBackendMap`
```

Sorry for make you confusing!

struss: > More specifically,
> You can (1) first add 'executors' and, (2) replace the internal logic of `executor` with that of `executors`, add 'TODO Remove executor' comment, (3) then finally remove executor.

If so, I think I will make a new PR and remain this as DRAFT.

dayo09:
@struss Has it been all merged..?

PR #1101, opened 2022-07-26
Refresh toolchain View after backend registraion

yunjayh:
This commit refreshes toolchainView after backend registraion to prevent
showing empty toolchainView tree.

---

This will resolve `ONE-vscode` extension and backend extension timing issue. (from https://github.com/Samsung/ONE-vscode/issues/932#issuecomment-1182658014)

### DIFF
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -50,6 +50,9 @@ function backendRegistrationApi() {
         globalExecutorArray.push(executor);
       }
       Logger.info(logTag, 'Backend', backendName, 'was registered into ONE-vscode.');
+      // NOTE: This might not 100% guaratee the activating extension has been done.
+      //   - link: https://github.com/Samsung/ONE-vscode/pull/1101#issuecomment-1195099002
+      // TODO: Consider better way to refresh toolchainView after backend's registration.
       vscode.commands.executeCommand('one.toolchain.refresh');
     },
     registerExecutor(executor: Executor) {
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import * as vscode from 'vscode';
 const assert = require('assert');
 import {Backend} from './Backend';
 import {gToolchainEnvMap, ToolchainEnv} from '../Toolchain/ToolchainEnv';
@@ -49,6 +50,7 @@ function backendRegistrationApi() {
         globalExecutorArray.push(executor);
       }
       Logger.info(logTag, 'Backend', backendName, 'was registered into ONE-vscode.');
+      vscode.commands.executeCommand('one.toolchain.refresh');
     },
     registerExecutor(executor: Executor) {
       globalExecutorArray.push(executor);


dayo09 (dismissed):

jyoungyun (dismissed):
LGTM

If this patch is applied when the backend is already registered normally, the default toolchain notification will be called twice. I am not sure if there is a better way, so I confirm this PR.

hyunsik-yoon:
Anyways, this code would work most time.

However, `registrationAPI` is called inside backend's `activate()` function.
So, technically, it would be possible that backend's `activate()` function is _not_ 100% guaranteed to finish when command `refresh` is being executed. I wonder if such situation is OK.

How about adding a comment (`TODO`) to avoid such possibility?

yunjayh:
> So, technically, it would be possible that backend's `activate()` function is _not_ 100% guaranteed to finish when command `refresh` is being executed. I wonder if such situation is OK.

IMHO if backend's `activate()` function doesn't finished, it is equivalent to failed to install the backend extension. So, for that case, the backend will not be registered to `ONE-vscode`, which means refresh isn't needed anymore. If I misunderstood something, please let me know.

hyunsik-yoon:
@yunjah, What I meat was as follows:

time t: Inside `activate()` (L39) of  extension.ts in backend extension eventually calls `registerBackend()` of one-vscode.
time t+1: inside `registerBackend()`, refresh command is sent. (L53 of this PR)
time t+2: two possibilities
- possibility 1: refresh command is executed (`activate()` (L39) of  extension.ts in backend extension at time t is not finished yet)
- possibility 2:  `activate()` (L39) of  extension.ts in backend extension finishes

If possibility 1 happens, that might mean backend extension is not yet fully activated (because `activate()` of backend extension does not finish yet). So I wonder if possibility 1 could happen and it could be OK.

[Could you try this at TS playground](https://www.typescriptlang.org/play?#code/FAQwzgngdgxgBAMwK6wC4EsD2U4DcwyYAmApgPokAeJMSqJAwpgLbMhREAUzAXHGKgBO6KAHMAlHADewOHLiEoYTABsSAOhWZRnAAYASKcwC+fBCPRgAFiSK7xAbmDHgwZGiw5s5fIVJlBElFLekEAIRAYAGsSDk5JGXk8AmJyKho6RhY2OIByVExVGCsQEThAhEDrXPFnV3cYDGw4ACNImI4ySIxcEHp46Vl5RWU1TW1OACI26NiiOG70Xvp+VBBBVDAeScchuW8yX1SAoJCScPa5+KckkdUNLR1dGY75xeWSODn7JxdgF7mXUaSz6JHiwCAA)?

In my case, it prints the following: 
```
[LOG]: "backend activate starts:" 
[LOG]: "toolchain refresh finished" 
[LOG]: "backend activate end"
```

yunjayh:
I changed the codes as following (including adding log in toolchain refresh function) and packaged, installed and executed. 

```ts
      ...
      vscode.commands.executeCommand('one.toolchain.refresh').then(()=>{Logger.info(logTag, 'Backend', 'TEMP')});
      Logger.info(logTag, 'Backend', backendName, 'was registered into ONE-vscode.');
      ...
```

I found the log comes out like this. 

```
[7/27/2022, 2:05:17 PM][ToolchainProvider][info] Refresh
[7/27/2022, 2:05:17 PM][backendRegistrationApi][info] Backend tv2 was registered into ONE-vscode.
[7/27/2022, 2:05:17 PM][backendRegistrationApi][info] Backend TEMP
```

I think it executes as you concerned. 

However, adding backend to global backend list is executed before calling refresh command asynchronously. It might guarantee the

hyunsik-yoon:
Still I don't have confidence with this code. To me it looks similar with the following code:

```ts
function register(foo: FOO) {
    foo.run()
}

class FOO {
    constructor() {
        register(this)
        console.log("FOO Construction is about to finish")
    }

    run() { console.log("runng FOO!!") }
}

new FOO()
```

where result is 
```
"runng FOO!!" 
"FOO Construction is about to finish" 
```

Here, constucting FOO is not finished but the method of FOO is called. It works. However, I am not sure this code will work in all versions of node.js. This PR might seems a bit similar with the above code...

Anyway, the show must go on. :-) 
Could you add some comment about such situation for readers like me in future?
Thanks.

yunjayh:
Your sample code makes me fully understand your concern clearly. I'll add proper comment. Thank you a lot!!

hyunsik-yoon (approved):
Thanks. :-)

jyoungyun (approved):
LGTM

dayo09 (approved):

PR #1098, opened 2022-07-25
[DRAFT][CfgEditor] Support codicon events

llFreetimell:
This draft is for supporting codicon hover help msg.
Furthermore, this draft includes improvements of current editor UI.

ONE-vscode-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

codicon

PR #1105, opened 2022-07-26
[CfgEditor] Add codicon events

llFreetimell:
This commit adds following events for codicon
- `codicon-search` : Open file dialog to select file or directory.
- `codicon-question` : Show help message when mouseover.
And this commit also removes some unnecessary codicons.

---

Draft : #1098 

---

#### With this PR

codicon

### DIFF
--- a/src/CfgEditor/CfgEditorPanel.ts
+++ b/src/CfgEditor/CfgEditorPanel.ts
@@ -106,6 +106,22 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
             vscode.workspace.applyEdit(edit);
           }
           break;
+        case 'getPathByDialog':
+          const dialogOptions = {
+            canSelectMany: false,
+            canSelectFolders: e.isFolder,
+            openLabel: 'Open',
+            filters: {'target files': e.ext, 'all files': ['*']}
+          };
+          let newPath = e.oldPath;
+          vscode.window.showOpenDialog(dialogOptions).then(fileUri => {
+            if (fileUri && fileUri[0]) {
+              newPath = fileUri[0].fsPath.toString();
+              webview.postMessage(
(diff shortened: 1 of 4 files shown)


yunjayh:
Q) When the file open dialog has been popped up, the file filter is applied (it is `target files`). What's that mean?

dayo09 (dismissed):

llFreetimell:
> Q) When the file open dialog has been popped up, the file filter is applied (it is target files). What's that mean?

If the dialog is for `one-quantize`, `circle` file would be filtered.
If the dialog is for `one-import-tflite`, `tflite` file would be filtered.
`target files` means those `circle` or `tflite` :)
I couldn't think about great word... :(

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -106,6 +106,22 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
             vscode.workspace.applyEdit(edit);
           }
           break;
+        case 'getPathByDialog':
+          const dialogOptions = {
+            canSelectMany: false,
+            canSelectFolders: e.isFolder,
+            openLabel: 'Open',
+            filters: {'target files': e.ext, 'all files': ['*']}
```
yunjayh, line 114: (Optional) How about this kind of approach?

```suggestion
            filters: {`target files(*.${e.ext})`: e.ext, 'all files': ['*']}
```

image

I followed how windows file open dialog shows.

yunjayh (dismissed):

src/CfgEditor/CfgEditorPanel.ts
llFreetimell: It looks good but key value should be constant I think :(
I will handle it later...!

yunjayh (approved):

dayo09 (approved):

PR #1106, opened 2022-07-26
[CfgEditor] Remove NYI button

llFreetimell:
There was a button for switching editor for cfg file.
But as the button is hard to be implemented now,
this commit removes the button not to cause unnecessary confusionfor users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,20 +103,6 @@
         "icon": "$(add)"
       },
       {
-        "command": "one.editor.openCfg",
-        "title": "Open cfg in GUI",
-        "category": "ONE",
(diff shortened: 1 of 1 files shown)


dayo09 (dismissed):

dayo09:
> But as the button is no more needed, this commit removes it.

I think we still need it. We need to 'open with text editor' by right clicking it for the action. If we  offer the switching button, UI'll be better.

llFreetimell:
I revised commit message, thanks :)

yunjayh (approved):
LGTM great works before release. :smile:

dayo09 (approved):

PR #1112, opened 2022-07-28
[CfgEditor] Change event type for text field

llFreetimell:
Until now, text field was added by `change` event type.
As a result, the value of the text field was applied after focused out.
This commit fixes to be applied right away with `input` event type.


jyoungyun (approved):
LGTM

hyunsik-yoon (approved):

Issue #988, opened 2022-07-07
Security-sensitive OS command


llFreetimell:
### What

Following codes are using security-sensitive OS commands.

https://github.com/Samsung/ONE-vscode/blob/aee748145b541dbc83141db05df718bd4c65ddb8/src/Execute/executeQuickInput.ts#L132

https://github.com/Samsung/ONE-vscode/blob/aee748145b541dbc83141db05df718bd4c65ddb8/src/Project/ToolRunner.ts#L178

Let's try to resolve this..!

### Reference

- https://rules.sonarsource.com/typescript/RSPEC-4721/

hyunsik-yoon:
I tried the following:

```
$ cat ./askpass.sh 
#!/bin/bash
echo your_password

$ export SUDO_ASKPASS=$(pwd)/askpass.sh
$ sudo -k  # reset previous sudo cache

$ sudo ls  # this asks password
[sudo] password for eric:   # let's press ctrl-C

$ sudo -A ls  # with -A, sudo does not ask password since it receive password from running SUDO_ASKPASS
askpass.sh
```

So, what we can try could be

1. get password from GUI
2. create a temp script 
3. run `spawnSync('sudo -A ...')` with `SUDO_ASKPASS` envvar
4. right after running, make sure (and make sure again!!) to remove the temp script!!

Period from 2 to 4 should be very short. I guess we could run 3 in `async` function...?

```ts
async run() {
   spawnSync(...)
}

async runMain() {
   createScript(...)
   run()  // run this asynchronously
   makeSureRemoveScript(...)
}

```

ref: `-A` option in `man sudo` and https://stackoverflow.com/questions/12608293/how-to-setup-a-sudo-askpass-environment-variable

jyoungyun:
I am worried that there should be a file with password if it's temporary....
Anyway I tried to apply this changes:

```
        const tempfile="temp.sh";
        cp.spawnSync(`echo "#!/bin/bash\necho ${process.env.userp}" > ${tempfile}`, {cwd: path});
        Logger.info(this.tag, `SUDO_ASKPASS=${path}/${tempfile} sudo -A ` + tool);
        this.child = cp.spawn(`SUDO_ASKPASS=${path}/${tempfile} sudo -A ` + tool, toolargs, {cwd: path, shell: true});
        cp.spawnSync(`rm ${tempfile}`);
```
Result:
```
[...]

PR #1123, opened 2022-08-02
[DRAFT] Remove `spawn()` with `shell:True`

hyunsik-yoon:
This tries to replace `spawn()` with `shell:True` with two `spawn()` calls.

for https://github.com/Samsung/ONE-vscode/issues/988

### DIFF
--- a/src/Job/ToolRunner.ts
+++ b/src/Job/ToolRunner.ts
@@ -17,6 +17,7 @@
 import * as cp from 'child_process';
 import * as fs from 'fs';
 import {Logger} from '../Utils/Logger';
+import {pipedSpawn} from '../Utils/PipedSpawn';
 import {ToolArgs} from './ToolArgs';
 
 const path = require('path');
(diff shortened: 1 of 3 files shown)


hyunsik-yoon:
I got strange error in log window.
```
[8/2/2022, 11:53:41 AM][JobRunner][info] Run tool: aptitude args: ["install","-o Aptitude::ProblemResolver::SolutionCost='100*canceled-actions,200*removals'","***","-q","-y"] cwd: /home/eric root: true
[8/2/2022, 11:53:41 AM][ToolRunner][info] Running: install
[sudo] password for eric: Error opening terminal: unknown.
```

I guess that this error is related to aptitude terminal setting...? 
(FYI, I uninstalled and installed aptitude yesterday.)

hyunsik-yoon:
Note: With current main branch, the above problem does not happen.

hyunsik-yoon:
Some articles mentioned that some env  variable(e.g., `TERM`) might be necessary. (in case of `htop` or `nano`)
But the following (`env -i` means no env var) worked fine. So this would not be env var issue.
```env -i echo some_pwd | sudo -S aptitude purge some_toolchain_name -q -y```

hyunsik-yoon:
The above error was from my silly mistake. I will hide related comment.

PR #1143, opened 2022-08-03
Remove `spawn(.., shell:true)`

hyunsik-yoon:
This removes `spawn(.., shell:true)` by introducing
a new function `pipedSpawn(..)`.

For #988

### DIFF
--- a/src/Job/ToolRunner.ts
+++ b/src/Job/ToolRunner.ts
@@ -17,6 +17,7 @@
 import * as cp from 'child_process';
 import * as fs from 'fs';
 import {Logger} from '../Utils/Logger';
+import {pipedSpawn} from '../Utils/PipedSpawn';
 import {ToolArgs} from './ToolArgs';
 
 const path = require('path');
@@ -52,7 +53,7 @@ export class ToolRunner {
   tag = this.constructor.name;  // logging tag
 
   // This variable is undefined while a prcess is not running
-  private child: cp.ChildProcessWithoutNullStreams|undefined = undefined;
+  private child: cp.ChildProcess|undefined = undefined;
 
   // When the spawned process was killed by kill() method, set this true
   // This value must be set to false when starting a process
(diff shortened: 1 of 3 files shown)


src/Job/ToolRunner.ts
```diff
@@ -52,27 +53,49 @@ export class ToolRunner {
   tag = this.constructor.name;  // logging tag
 
   // This variable is undefined while a prcess is not running
-  private child: cp.ChildProcessWithoutNullStreams|undefined = undefined;
+  private child: cp.ChildProcess|undefined = undefined;
```
jyoungyun, line 56: (Q) Why do you change to `ChildProcess`?

src/Job/ToolRunner.ts
```diff
@@ -52,27 +53,49 @@ export class ToolRunner {
...
+
+          `echo correct_pw | sudo -S ...` --> `sudo -k` --> `echo wrong_pw | sudo -S ...`
+
+        When should we call `sudo -k`?
+        1. We observed that `sudo -k` right after spawning `echo correct_pw | sudo -S ...` does not
+          seem to work, making `echo wrong_pw | sudo -S ...` succeed.
+        2. Calling `sudo -k` after first sudo process finishes seems to make
+          `echo wrong_pw | sudo -S ...` fail.
+        */
+        cp.spawnSync('sudo', ['-k']);
```
jyoungyun, line 96:
```suggestion
        cp.spawnSync('sudo', ['-k']);
        process.env.userp = '';
```

from https://github.com/Samsung/ONE-vscode/blob/main/src/Job/JobRunner.ts#L96

src/Utils/PipedSpawn.ts
```diff
@@ -0,0 +1,61 @@
...
+ * grep = pipedSpawn('ls', ['-hl', '/'], {cwd: '/'}, 'grep', ['usr'], {});
+ * grep.stdout.on('data', (data)=>{ console.log(data.toString(); })
+ * grep.stderr.on('data', (data)=>{ console.log(data.toString(); })
+ * grep.on('exit', (exitcode: number | null, signal: NodeJS.Signals | null)=>{
+ *   if (exitcode === null) { // handle signal
+ *   } else if (exitcode === 0) { console.log("success!");
+ *   } else { console.log("exit code is not 0"); })
+ */
+export function pipedSpawn(
+    cmd1: string, cmd1List: string[], cmd1Option: SpawnOptions, cmd2: string, cmd2List: string[],
```
dayo09, line 37: how about taking arguments as : `string[] and SpawnOptions`..? 
```suggestion
    cmd1: { cmdstrs: string[], options: SpawnOpotions}, cmd2: { cmdstrs: string[], options: SpawnOpotions}...
```

Because 'cmd1' is the command string which comes first and cmd1List is others, you can simplify the args.

src/Job/ToolRunner.ts
hyunsik-yoon: I looked into the code and found how I can use previous `ChildProcessWithoutNullStreams`.
Let me use `ChildProcessWithoutNullStreams` again. Thanks.

src/Utils/PipedSpawn.ts
hyunsik-yoon: Thanks for the suggestion.
Actually, this follows the parameter of `spawn(cmd: string, args: string[], option: SpawnOption)`. So I'd like to stick to this convention. :-)
Plus, `pipedSpawn()` will call two `spawn()` internally. If we use `cmd1: { cmdstrs: string[]...` we need to construct param (cmd string and arg of string[]) for `spawn` again.

jyoungyun (approved):
LGTM Thank you!!!

src/Utils/PipedSpawn.ts
dayo09: Sure, I understand your intention :-D

dayo09 (approved):

PR #1144, opened 2022-08-03
[Job] Use logger instead of Balloon

jyoungyun:
This commit uses logger instead of Balloon.
The information messages depend on the requested job.
So it is better to handle the message where it is requested.


llFreetimell (approved):
LGTM

hyunsik-yoon (approved):

PR #1146, opened 2022-08-03
Initialize Test Case Administrator with fixing some TCs

llFreetimell:
This commit initialize Test Case Administrator with fixing following things.
- Remove duplicated TC names even they are in the same suite
- Add missing assert


hyunsik-yoon (approved):

dayo09 (approved):

jyoungyun (approved):
LGTM

PR #1154, opened 2022-08-04
[Utils] Improve line and function coverage

llFreetimell:
This commit improves line and function coverage by following things
- Remove unused function
- Add more testcases

### DIFF
--- a/src/Tests/Utils/Helpers.test.ts
+++ b/src/Tests/Utils/Helpers.test.ts
@@ -64,17 +64,35 @@ suite('Utils', function() {
     });
 
     suite('#areEqual()', function() {
-      test('compare lexically the same paths', function() {
+      test('compare same paths', function() {
         assert.isTrue(RealPath.areEqual('/', '/'));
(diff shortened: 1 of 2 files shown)


src/Tests/Utils/Helpers.test.ts
```diff
@@ -64,17 +64,35 @@ suite('Utils', function() {
...
+
+      test('NEG: compare not creatable paths - 5', function() {
+        assert.isFalse(RealPath.areEqual('/dummy', '/./../dummy/./'));
       });
     });
 
     suite('#exists()', function() {
       test('check if the path exists', function() {
+        assert.isFalse(RealPath.exists(undefined));
         assert.isFalse(RealPath.exists('/dummy/not/exist'));
```
hyunsik-yoon, line 96: It seems like these two are negative test.

llFreetimell: I think 'negative' means that something hard to expect or something different with what we usually think :)
So I thought that this is positive test case because they are definitely expected to be false.

hyunsik-yoon (approved):

PR #1162, opened 2022-08-05
[CircleGraph] Remove testcases

llFreetimell:
As all the classes in `CircleGraph` are excluded for coverage measurement,
testcases are no more used.
So this commit removes them.


seanshpark (approved):
LGTM thank you!

PR #1169, opened 2022-08-05
Hide one.toolchain.{install,refresh} from cmd palette

hyunsik-yoon:
This hides the following from command palette (`ctrl-shift-p`) just like vwe already did for `one.toolchain.uninstall`.

- `one.toolchain.install`
- `one.toolchain.refresh`

### DIFF
--- a/package.json
+++ b/package.json
@@ -307,7 +307,15 @@
           "when": "false"
         },
         {
+          "command": "one.toolchain.install",
+          "when": "false"
+        },
+        {
           "command": "one.toolchain.uninstall",
+          "when": "false"
+        },
+        {
+          "command": "one.toolchain.refresh",
           "when": "false"
         },
         {
--- a/package.json
+++ b/package.json
@@ -308,7 +308,7 @@
         },
         {
           "command": "one.toolchain.install",
-          "when": "false"
+          "when": "true"
         },
         {
           "command": "one.toolchain.uninstall",


yunjayh (dismissed):

package.json
```diff
@@ -306,10 +306,18 @@
           "command": "one.explorer.showExtra",
           "when": "false"
         },
+        {
+          "command": "one.toolchain.install",
+          "when": "false"
+        },
```
jyoungyun, line 312: Um, I want to show this command in command palette. Because it is very useful to install new toolchain. :)

package.json
```diff
@@ -306,10 +306,18 @@
...
         {
           "command": "one.toolchain.uninstall",
           "when": "false"
         },
+        {
+          "command": "one.toolchain.refresh",
+          "when": "false"
+        },
```
jyoungyun, line 320: 👍

yunjayh (approved):

hyunsik-yoon:
I intentionally  left the following inside `"commandPalette" : [ ... ]` 
because I believe that explicitly showing is better than implicit one. 😸 

```
        {
          "command": "one.toolchain.install",
          "when": "true"
        },
```

package.json
hyunsik-yoon: I see. Let me change `faler` to `true`.

jyoungyun (approved):
LGTM Thank you :)

Issue #1135, opened 2022-08-03
[Profile] Support JsonTracer one-click opening


dayo09:
## What?

Let's enable opening 'JsonTracer' by clicking trace.json file.
Currently, 'JsonTracer' opens only a panel and users should manually 'load' file with file selection window. 

## Why?

To support 'open with view' for chrome trace files in OneExplorer.

dayo09:
For this, we need to introduce ChromeTracerViewer custom text editor provider... It takes a bit of works

dayo09:
#### Study

1. CustomEditorProvider vs CustomTextEditorProvider : which one to use?
  - CustomEditorProvider for binary files, CustomTextEditorProvider for text based files (json)
  - Let's use CustomTextEditorProvider

PR #1172, opened 2022-08-05
[Jsontracer] Enable JsonTracer as custom editor

llFreetimell:
This commit enables JsonTracer as custom editor,
and deprecate commands.

---

Parent Issue : #1135


package.json
```diff
@@ -87,6 +87,16 @@
...
+      },
+      {
+        "viewType": "one.editor.jsonTracer",
+        "displayName": "chromeTrace",
+        "selector": [
+          {
+            "filenamePattern": "*.trace.json"
+          }
+        ],
+        "priority": "default"
```
hyunsik-yoon, line 99: It would be better to open this json to be opened not in vscode explorer but in ONE explorer.

```suggestion
        "priority": "option"
```

package.json
```diff
@@ -87,6 +87,16 @@
           }
         ],
         "priority": "default"
+      },
+      {
+        "viewType": "one.editor.jsonTracer",
+        "displayName": "chromeTrace",
+        "selector": [
+          {
+            "filenamePattern": "*.trace.json"
```
hyunsik-yoon, line 96: One concern is the name could be anything (our backend profiling too have file name option from user)
Some possible solution could be
- Make constraint with file name: "File name must ends with `trace.json`. Maybe we need such constraint in `Profile` tab of cfg editor.
- Or can ONE explorer open json if json has, e.g., `"traceEvents"` key? (it seems chrome trace file has `"traceEvents"` key) /cc @dayo09

package.json
llFreetimell: Except `.cfg`, other file extensions are registered as `default`.
Should we change them all?

package.json
llFreetimell: I thought that `*.timeline.json` was used for profiling tool, but not sure :(

dayo09: > Or can ONE explorer open json if json has, e.g., "traceEvents" key? (it seems chrome trace file has "traceEvents" key) /cc @dayo09

Yes, it's possible. 

> I thought that *.timeline.json was used for profiling tool, but not sure :(

It's just a conventions, not a rule.

src/Jsontracer/JsonTracerViewerPanel.ts
```diff
@@ -0,0 +1,126 @@
...
+    registrations.forEach(disposable => context.subscriptions.push(disposable));
+  };
+
+  constructor(private readonly context: vscode.ExtensionContext) {}
+
+  public async resolveCustomTextEditor(
+      document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel,
+      _token: vscode.CancellationToken): Promise<void> {
+    await this.initWebview(document, webviewPanel.webview);
+    this.initWebviewPanel(document, webviewPanel);
```
dayo09, line 74: Q. 'await' not needed? It seems the function sets some contexts.

src/Jsontracer/JsonTracerViewerPanel.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (x) {
+          x.dispose();
+        }
+      }
+      vscode.commands.executeCommand('setContext', JsonTracerViewerPanel.viewType, false);
+    });
+  }
+
+  private updateWebview(document: vscode.TextDocument, webview: vscode.Webview): void {
+    webview.postMessage({type: 'load', content: JSON.parse(document.getText()).traceEvents});
```
dayo09, line 124: (optional, maybe TODO later)

`content: JSON.parse(document.getText()).traceEvents`
I think this will trigger some javscript errors, how about handling the case?

package.json
```diff
@@ -87,6 +87,16 @@
           }
         ],
         "priority": "default"
+      },
+      {
+        "viewType": "one.editor.jsonTracer",
+        "displayName": "chromeTrace",
```
dayo09, line 93:
```suggestion
        "displayName": "Chrome Tracer",
```

How about changing the displayName is above?

It's a name that will be used by vscode UI commands, for example,
Screenshot from 2022-08-05 13-44-40

src/Jsontracer/JsonTracerViewerPanel.ts
```diff
@@ -0,0 +1,126 @@
...
+import {getUri} from '../Utils/external/Uri';
+
+
+/* istanbul ignore next */
+export class JsonTracerViewerPanel implements vscode.CustomTextEditorProvider {
+  private _disposables: vscode.Disposable[] = [];
+
+  public static readonly viewType = 'one.editor.jsonTracer';
+
+  public static register(context: vscode.ExtensionContext): void {
```
dayo09, line 54: Thanks for going along with the issue 😄 
- #1176

src/Jsontracer/JsonTracerViewerPanel.ts
```diff
@@ -0,0 +1,126 @@
...
+https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/custom-editor-sample/src/catScratchEditor.ts
+*/
+
+import * as vscode from 'vscode';
+import {getNonce} from '../Utils/external/Nonce';
+import {getUri} from '../Utils/external/Uri';
+
+
+/* istanbul ignore next */
+export class JsonTracerViewerPanel implements vscode.CustomTextEditorProvider {
```
dayo09, line 49: (just a suggestion, not about this PR)

It seems the name 'JsonTracer' is a bit confusing because we have another 'JsonTracer' module : mondrianViewer. Mondrian files are another type of our 'json' file which 'traces' mondrian (its file extension names are 'tracealloc.json').

Therefore, I suggest changing the module name as 'ChromeTracer'. We actually uses the 'chrome tracer' open source solution here, so the naming sounds more direct.

package.json
hyunsik-yoon: Sure. That would be awesome. I guess that could be done in next PR. :-)

package.json
hyunsik-yoon: I think that we all understand this issue now.
My opinion is... let's leave this line in this PR as it is and consider this issue in another PR. :-)

src/Jsontracer/JsonTracerViewerPanel.ts
llFreetimell: It's because of `initWebview` :)
In the `initWebview` function, following `await` is used.

```ts
let html = Buffer.from(await vscode.workspace.fs.readFile(htmlUri)).toString();
```

package.json
llFreetimell: Make sense :D
But as this is "Json Tracer", not "Chrome", I will change it to `"Json Tracer" :)

src/Jsontracer/JsonTracerViewerPanel.ts
llFreetimell: I agree with the background of this suggestion and also I agree with the confusion :)

Only one thing I worry about is... "Can we really use the name of `Chrome` or `ChromeTracer` for our tool?" in legally.
Since `Chrome` and `ChromeTracer` is a kind of trademark of Google's, we may invade their right of the naming :(

So how about naming with another idea?
`OneTracer`, `CycleViewer`, `TimelineViewer`, etc.... nothing good idea comes now T.T

package.json
llFreetimell: I will follow the result of #1183 :D
But I agree with capital letters!

package.json
hyunsik-yoon: @dayo I left a comment in the issue you mentioned. :-)

hyunsik-yoon (approved):

src/Jsontracer/JsonTracerViewerPanel.ts
dayo09: I don't mean initWebview functions, but initWebviewPanel...
I just found out that it's syncronous, so never mind 😄

dayo09: It doesn't look critical, and I believe that you already examined that, so never mind :-D

dayo09 (approved):
LGTM, thanks :-D

dayo09:
Q. I fetched the PR and run the extension, and I cannot open *.timeline.json file via default explorer. I cannot find it on `option` for open file. 😢

hyunsik-yoon (approved):

media/Jsontracer/processData.js
```diff
@@ -50,35 +50,7 @@ import renderDashboard from './dashboard.js';
...
-}
-
-function processFile(file) {
-  const reader = new FileReader();
-  reader.onload = () => {
-    const data = JSON.parse(reader.result);
-    processData(data.traceEvents);
-
-    // set data to DOM
-    setData.dataset['displayTimeUnit'] = data.displayTimeUnit;
```
yunjayh, line 76: Since this part has been deprecated, the features which are using `displayTimeUnit` makes bugs :sob:

For example, when user hit a block in jsontracer, `ts` and `dur` item are shown as `NaN undefined`. That's caused by following lines.
https://github.com/Samsung/ONE-vscode/blob/9324f977bae56cea9ea3e5f6404c89d43bc8b0d1/media/Jsontracer/detail.js#L76-L80
`const displayTimeUnit` became undefined, since it has not been set yet. T^T 

I'll post a new issue tomorrow :)

PR #1176, opened 2022-08-05
[Toolchain] Remove unset userpassword in child process's exit function

jyoungyun:
This commit removes unset userpassword code in child process's exit
function. This password can be used again even after one job is
finished. For example, uninstall and install are executed in one
workflow. So it should not delete the user password just because
one job is finished.

### DIFF
--- a/src/Job/ToolRunner.ts
+++ b/src/Job/ToolRunner.ts
@@ -94,7 +94,6 @@ export class ToolRunner {
           `echo wrong_pw | sudo -S ...` fail.
         */
         cp.spawnSync('sudo', ['-k']);
-        process.env.userp = '';
       }
 
       // From https://nodejs.org/api/child_process.html#event-exit


hyunsik-yoon (approved):

llFreetimell (approved):

hyunsik-yoon:
> uninstall and install are executed in one workflow. So it should not delete the user password just because one job is finished.

I am OK with this PR.
Also we need to make sure that we invalidate this env var after uninstall and then install. 
Currently it seems like we invalidate this env var whenever one job is finished... but `uninstall and install ` needs two jobs right? So I guess env var is invalidate after the first job....?

jyoungyun:
/cc @hyunsik-yoon 

> Currently it seems like we invalidate this env var whenever one job is finished... but uninstall and install needs two jobs right? So I guess env var is invalidate after the first job....?


The JobRunner may have multiple jobs to be executed. The JobRunner executes one job by executing ToolRunner. Thus, the ToolRunner can be called multiple times in one JobRunner flow. The existing code invalidates the user password after one job is finished. But, if there are many jobs to run on JobRunner, the user password must be maintained until all jobs are finished. So I suggested this PR and the user password will be invalidated in [here](https://github.com/Samsung/ONE-vscode/blob/main/src/Job/JobRunner.ts#L101).

PR #1182, opened 2022-08-05
[View] Add InstallQuickInput test cases

jyoungyun:
This commit adds InstallQuickInput test cases.

### DIFF
--- a/src/View/InstallQuickInput.ts
+++ b/src/View/InstallQuickInput.ts
@@ -30,7 +30,6 @@ async function shouldResume() {
       });
 }
 
-/* istanbul ignore next */
 class InnerButton implements vscode.QuickInputButton {
   constructor(public iconPath: vscode.ThemeIcon, public tooltip: string) {}
 }
@@ -42,7 +41,7 @@ enum InstallQuickInputStep {
   pickVersion = 3
 }
 
-interface State {
+interface InstallQuickInputState {
   selectedItem: vscode.QuickPickItem|InnerButton;
   current: InstallQuickInputStep;
 }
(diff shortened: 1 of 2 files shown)


jyoungyun:
**After**

install_quickinput_tem

dayo09 (approved):

src/View/InstallQuickInput.ts
```diff
@@ -168,11 +166,12 @@ class InstallQuickInput {
     });
     this.toolchainEnv = this.getToolchainEnvFromGlobal(state.selectedItem.label);
 
-    this.changeCurrentStepAfter(this.pickBackend.name, state as State);
+    this.changeCurrentStepAfter(this.pickBackend.name, state as InstallQuickInputState);
   }
 
-  async pickType(input: MultiStepInput, state: Partial<State>) {
-    this.changeCurrentStepBefore(this.pickType.name, state as State);
+  /* istanbul ignore next */
```
llFreetimell, line 172: I think some of codes can be tested, can I ask you why this function is excluded? :)

src/View/InstallQuickInput.ts
```diff
@@ -201,11 +200,12 @@ class InstallQuickInput {
...
     }
 
-    this.changeCurrentStepAfter(this.pickType.name, state as State);
+    this.changeCurrentStepAfter(this.pickType.name, state as InstallQuickInputState);
   }
 
-  async pickVersion(input: MultiStepInput, state: Partial<State>) {
-    this.changeCurrentStepBefore(this.pickVersion.name, state as State);
+  /* istanbul ignore next */
+  public async pickVersion(input: MultiStepInput, state: Partial<InstallQuickInputState>) {
```
llFreetimell, line 207: I think some of codes can be tested, can I ask you why this function is excluded? :)

src/View/InstallQuickInput.ts
```diff
@@ -264,10 +264,11 @@ class InstallQuickInput {
     this.version = state.selectedItem.label;
     this.toolchain = toolchains[versions.indexOf(this.version)];
 
-    this.changeCurrentStepAfter(this.pickVersion.name, state as State);
+    this.changeCurrentStepAfter(this.pickVersion.name, state as InstallQuickInputState);
   }
 
-  async updateBackend(input: MultiStepInput, state: Partial<State>) {
+  /* istanbul ignore next */
```
llFreetimell, line 270: I cannot find any UI related codes... is there something I missed..?

src/View/InstallQuickInput.ts
```diff
@@ -59,62 +58,60 @@ class InstallQuickInput {
...
     return this.error;
   }
 
-  getQuickPickItems(items: string[]): vscode.QuickPickItem[] {
+  public getQuickPickItems(items: string[]): vscode.QuickPickItem[] {
     return items.map(label => ({label}));
   }
 
-  getAllToolchainEnvNames(): string[] {
+  public getAllToolchainEnvNames(): string[] {
```
dayo09, line 97: FYI, if they are private functions, then it's not necessary to be cover them separately in unit test as long as they can be covered by other public functions. 

I think some of those functions have been used a private because they don't need to be used outside. In terms of encapsulation, I think you can keep some of them as private rather then exposing them..

src/View/InstallQuickInput.ts
jyoungyun: Because this function is operated by user input.
`input.showQuickPick` function in L.191 can not be tested. :(

llFreetimell: I misunderstood that `MultiStepInput` is kind of vector or array something like that :(
Now I understand why these functions should be excluded :)
Thanks!

src/View/InstallQuickInput.ts
jyoungyun: Thank you for information. :) This function is originally a public method. I just added the keyword explicitly. :)

llFreetimell (approved):
LGTM

PR #1190, opened 2022-08-08
[OneExplorer] Show chrome trace files

dayo09:
This commit shows chrome trace files on ONE Explorer.

----

Need rebase after #1172

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -198,6 +198,12 @@ export class ConfigObj {
 
     let locatorRunner = new LocatorRunner();
 
+    /**
+     * ABOUT ORDERING
+     *
+     * The registration order determines the order in the tree view
+     */
+
     locatorRunner.register({
       artifactAttr: {
         ext: '.circle',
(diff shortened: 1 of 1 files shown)


dayo09:
### GIF
0808-trace-json

struss (approved):
LGTM

src/OneExplorer/ConfigObject.ts
```diff
@@ -225,6 +231,20 @@ export class ConfigObj {
...
+    locatorRunner.register({
+      artifactAttr: {
+        ext: '.json',
+        icon: new vscode.ThemeIcon('graph'),
+        openViewType: 'one.editor.jsonTracer',
+        canHide: true
+      },
+      locator: new Locator(
+          (value: string) => {
+            return LocatorRunner.searchWithCommandOption(value, '--save-chrome-trace', '.json');
```
hyunsik-yoon, line 243: `--save-chrome-trace` is backend-specific option and I guess this is OK for quick fix because we already have many backend specific terms in this file (`tracealloc.json`, `tvn` (on L217, L230 etc), `tv2m`,... ) 😹 
Maybe after release, we should eventually remove the these backend-specific terms from this file.

hyunsik-yoon (approved):

src/OneExplorer/ConfigObject.ts
dayo09: Sure ;-D Actually I was working on that and tried backend-refactoring before. I will raise up its priority after this.

PR #1192, opened 2022-08-08
[Jsontracer] Fix to display json content initially

llFreetimell:
Until now, json content is not loaded initially, especially in remote-ssh environment.
It was because displaying json request is done even before all the elements are not ready.
This commit will fix it to request displaying json content after all the elements are ready.


hyunsik-yoon (approved):
Now it works well on my SSH extension env. Thanks.

dayo09 (approved):

PR #1197, opened 2022-08-08
[Tests/Utils] Add a negative TC

llFreetimell:
This commit adds a negative test case.

### DIFF
--- a/src/Tests/Utils/PipedSpawn.test.ts
+++ b/src/Tests/Utils/PipedSpawn.test.ts
@@ -39,7 +39,7 @@ suite('Utils', function() {
 
     test('NEG: first cmd fails', function() {
       try {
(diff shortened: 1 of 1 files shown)


src/Tests/Utils/PipedSpawn.test.ts
```diff
@@ -37,6 +37,14 @@ suite('Utils', function() {
       });
     });
 
+    test('NEG: first cmd fails', function() {
+      try {
+        let cat = pipedSpawn('cat', ['invalid_file'], {}, 'grep', ['not_exist'], {});
```
hyunsik-yoon, line 42: How about the following? (to show that only first line can fail and throw exception)
```suggestion
        let cat = pipedSpawn('cat', ['invalid_file'], {}, 'wc', ['-l'], {});
```

hyunsik-yoon (approved):
Thanks. :-)

PR #1207, opened 2022-08-10
[OneExplorer] Fix the bug of delete command in ssh-remote env

dayo09:
This commit fixes the bug of deleting file not working in ssh-remote env.
It fixes:
- remove undefined operator '? :', it's only for cpp
- do not use 'useTrash' option if it's ssh-remote env. With the option, the entire deletion command fails.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -610,7 +610,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
     let recursive: boolean;
     let title = `Are you sure you want to delete '${oneNode.node.name}'`;
     if (isDirectory) {
-      title += `and its contents?`;
+      title += ` and its contents?`;
       recursive = true;
     } else {
       title += `?`;
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -606,19 +606,39 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
    */
   delete(oneNode: OneNode): void {
     const isDirectory = (oneNode.node.type === NodeType.directory);
-    const title =
-        `Are you sure you want to delete '${path.parse(oneNode.node.path).base}'` + isDirectory ?
-        'and its contents?' :
-        '?';
-    const detail = `You can restore this file from the Trash.`;
-    const approval = 'Move to Trash';
-    const recursive = isDirectory ? true : false;
+
+    let recursive: boolean;
+    let title = `Are you sure you want to delete '${oneNode.node.name}'`;
+    if (isDirectory) {
+      title += `and its contents?`;
+      recursive = true;
+    } else {
+      title += `?`;
+      recursive = false;
+    }
+
+    let detail: string|undefined;
+    let approval: string;
+    let useTrash: boolean;
+
+    if (vscode.env.remoteName === 'ssh-remote') {
+      // NOTE(dayo)
+      // By experience, the file is not deleted with 'useTrash:true' option.
+      approval = 'Delete';
+      detail = 'The file will be deleted permanently.';
+      useTrash = false;
+    } else {
+      approval = 'Move to Trash';
+      detail = `You can restore this file from the Trash.`;
+      useTrash = true;
+    }
+
 
     vscode.window.showInformationMessage(title, {detail: detail, modal: true}, approval)
         .then(ans => {
           if (ans === approval) {
             Logger.info('OneExplorer', `Delete '${oneNode.node.name}'.`);
-            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: true})
+            vscode.workspace.fs.delete(oneNode.node.uri, {recursive: recursive, useTrash: useTrash})
                 .then(() => this.refresh());
           }
         });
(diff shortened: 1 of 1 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -606,19 +606,39 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
-        'and its contents?' :
-        '?';
-    const detail = `You can restore this file from the Trash.`;
-    const approval = 'Move to Trash';
-    const recursive = isDirectory ? true : false;
+
+    let recursive: boolean;
+    let title = `Are you sure you want to delete '${oneNode.node.name}'`;
+    if (isDirectory) {
+      title += `and its contents?`;
```
hyunsik-yoon, line 613:
```suggestion
      title += ` and its contents?`;
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -606,19 +606,39 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
+    } else {
+      title += `?`;
+      recursive = false;
+    }
+
+    let detail: string|undefined;
+    let approval: string;
+    let useTrash: boolean;
+
+    if (vscode.env.remoteName === 'ssh-remote') {
```
hyunsik-yoon, line 624: image

hyunsik-yoon:
This version works like a charm in my env. :-)
We could consider deleting files with multiple selection later. :-)

hyunsik-yoon (dismissed):
Thanks.

src/OneExplorer/OneExplorer.ts
dayo09: FYI,

image

from https://code.visualstudio.com/api/references/vscode-api

jyoungyun:
Well, I don't understand why that happens, but my environment does not delete files on system.
delete_1

hyunsik-yoon:
@jyoungyun Could you describe your env?

In my case,
- vscode on Windows 10
- workspace on Ubuntu 18.04
- SSH-extension

src/OneExplorer/OneExplorer.ts
```diff
@@ -606,19 +606,39 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
...
```
jyoungyun, line 624:
```suggestion
    if (vscode.env.remoteName) {
```

The `remoteName` is defined by extensions. There are many cases that can be included in this value. If it is remote, it can't use the trash, so it would be better to process it like this.

dayo09: I agree :-D
But I was not sure if `vscode.env.remoteName` would be undefined when it's run on local.

jyoungyun:
> @jyoungyun Could you describe your env?

The `remoteName` in my environment is `attached-container`. 

This is my test environment:

- vscode on Windows 10
- ssh-extension (to Linux Ubuntu 18.04)
- docker extension (in Linux ssh extension)
- Test on Ubuntu 18.04 & Ubuntu 20.04

src/OneExplorer/OneExplorer.ts
dayo09: Actually I cannot fully understand what the line means. Maybe I should find by `extensionKind`?

image

extensionKind in our package.json is 'workspace'.
image

jyoungyun: I also think it would be better to use `extensionKind` option. After I defined `extensionKind` as `ui` in our package.json file, I created a vsix package and installed it within the docker extension. Then, the `extensionKind` value was changed to `workspace`. But we need to investigate it more. :)

dayo09: Oh I just figured out what it means.

It concerns the case that our extension is installed in both local and remote. 
If so, the value will be always 'ssh-remote' even if we open the local workspace.
As our extension defines 'extensionKind' as 'workspace', it will run an extension only depending on where the workspace is (remote or local)

hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

Please confirm this code to @dayo09 one more time before it is merged.

src/OneExplorer/OneExplorer.ts
dayo09: It seems

(1) Case 1. Local host only
remoteName : undefined

(2) Case 2. Remote only
remoteName : ssh-remote or else, defined
extensionKind: workspace

(3) Case 3. Local extension and remote extension both installed
Because our exensionKind in 'workspace', it will run dependes on which directory we opens on vscode

When I opened the workspace in remote, extensionKind: workspace
When I opened the workspace in local window pc, extensionKind: ui

jyoungyun: What is the value in this case?

> (1) Case 1. Local host only
> remoteName : undefined

dayo09: @jyoungyun

Well I just uninstalled the remote installed one-vscode and tested in my window pc,
it's UI, as same as `Case 3-Local`.

I think that `~ the value is defined in all extension hosts (local and remote) in case a remote extension host exists.` is not about Case 3.

dayo09: ------------
The conversation is moved to #1209

dayo09: Well now it looks like 'extensionKind: ui' means it's running on local PC, and 'extensionKind: workspace' means it's running on remote PC..

PR #1208, opened 2022-08-10
[OneExplorer] Fix the bug of not showing trace result immediately

dayo09:
This commit fixes the bug of not showing the trace result automatically,
but only after refreshing the view, if the profiling took long.

---

Thanks for the report @jyoungyun 😄

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -279,6 +279,13 @@ export class ConfigObj {
       })
     });
 
+    /**
+     * When you add a new product type, please append the ext type to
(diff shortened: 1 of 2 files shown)


dayo09:
#### Tested

I tested with `fsrcnn` model which @jyoungyun has reported with, and it showed up instantly after the long profiling times :-D

hyunsik-yoon (approved):

hyunsik-yoon:
In my env, now json generated from 'profile' appears in the ONE Explorer well.  (previously it didn't show up.)

jyoungyun (approved):
LGTM

PR #1210, opened 2022-08-10
[CfgEditor] Bugfix for multiple cfg editor panels are created

llFreetimell:
When multiple cfg editor are created, file content was not modified correctly.
It was because `CfgEditorPanel` is only created once,
while multiple panels try to modify cfg file content.
This commit will fix it by introducing `map` structure by using document URI as key.


jyoungyun (approved):
LGTM Thank you

hyunsik-yoon (approved):

PR #1212, opened 2022-08-11
[OneExplorer] Fix to use then phrase

llFreetimell:
`refresh()` was sometimes called after `rename(...)` is not yet finished.
This commit fixes it by using then phrase.


yunjayh (approved):
LGTM Thank you!

struss (approved):
LGTM

Issue #1209, opened 2022-08-10
How to know if the extension state is remote


dayo09:
## What?

Let's understand and find out the most accurate way to know the extension state

From https://github.com/Samsung/ONE-vscode/pull/1207#discussion_r942275882_

@jyoungyun 
> The `remoteName` is defined by extensions. There are many cases that can be included in this value. If it is remote, it can't use the trash, so it would be better to process it like this.

@dayo09
> I agree :-D But I was not sure if `vscode.env.remoteName` would be undefined when it's run on local.

@dayo09
> Actually I cannot fully understand what the line means. Maybe I should find by `extensionKind`?
> 
> ![image](https://user-images.githubusercontent.com/17171963/183877590-a3bca211-ef2c-4af2-9be6-a34d724c7128.png)
> 
> extensionKind in our package.json is 'workspace'. ![image](https://user-images.githubusercontent.com/17171963/183878543-3cf78836-e805-44d2-a881-8f08c54c608f.png)

@jyoungyun 
> I also think it would be better to use `extensionKind` option. After I defined `extensionKind` as `ui` in our package.json file, I created a vsix package and installed it within the docker extension. Then, the `extensionKind` value was changed to `workspace`. But we need to investigate it more. :)

@dayo09
> Oh I just figured out what it means.
> 
> It concerns the case that our extension is installed in both local and remote. If so, the value will be always 'ssh-remote' even if we open the local workspace. As our extension defines 'extensionKind' as 'workspace', it will run an extension only depending on where the workspace is (remote or local)

@dayo09
> It seems
> 
> (1) Case 1. Local host only remoteName : undefined
> 
> (2) Case 2. Remote only remoteName : ssh-remote (attached-container, for docker) extensionKind: Workspace
> 
> (3) Case 3. Local extension and remote extension both installed Because our exensionKind in 'workspace', it will run dependes on which directory we opens on vscode
> 
> * extensionKind: workspace (when I opened the workspace in remote)
>   
>   * remoteName: ssh-remote
>   * extensionKind: Workspace
> * Local: when I opened the workspace in the local window pc
>   
>   * remoteName: undefined (My expectation was that remoteName is ssh-remote here too, by [[OneExplorer] Fix the bug of delete command in ssh-remote env #1207 (comment)](https://github.com/Samsung/ONE-vscode/pull/1207#discussion_r942283543) 😮‍💨 )
>   * extensionKind: ui

@jyoungyun
> What is the value in this case?
> 
[...]

PR #1220, opened 2022-08-19
[OneExplorer] Use extensionKind to know whether extensions' remote

dayo09:
This commit uses extensionKind instead of remoteName
to know if the extension is remote or not.

----

For #1209
/cc @jyoungyun @hyunsik-yoon

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -22,7 +22,7 @@ import * as vscode from 'vscode';
 
 import {CfgEditorPanel} from '../CfgEditor/CfgEditorPanel';
 import {Balloon} from '../Utils/Balloon';
-import {obtainWorkspaceRoot, RealPath} from '../Utils/Helpers';
+import {obtainWorkspaceRoot} from '../Utils/Helpers';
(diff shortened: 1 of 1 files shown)


jyoungyun (approved):
Looks good to me :) Thank you.

src/OneExplorer/OneExplorer.ts
```diff
@@ -621,7 +648,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
     let approval: string;
     let useTrash: boolean;
 
-    if (vscode.env.remoteName) {
+    if (this.isRemote) {
```
llFreetimell, line 651: Maybe...?

```suggestion
    if (this.isRemote()) {
```

dayo09: It's a getter, so it doesn't need parenthesis :-D

https://www.typescripttutorial.net/typescript-tutorial/typescript-getters-setters/

hyunsik-yoon:
I checked this in my SSH-extension env and also in a single Linux box env.
In both case, `_extensionKind` has value described by @dayo09.

llFreetimell (approved):
LGTM

Issue #1223, opened 2022-08-22
[OneExplorer] Improve latency


dayo09:
## What?

Let's improve latency of building the tree in the view.

## How?

- [x] Save cfg list and rebuild the list on refresh
- [ ] Do not use synchronous functions
- [ ] More?

dayo09:
### TEST: Save cfg list and rebuild the list on refresh
#### Current main branch, latency on refresh
![0823-cacheOFF](https://user-images.githubusercontent.com/17171963/186050061-6702fe70-2d7c-4d96-9bcd-bb3a7d930e2c.gif)

#### When the change is applied
![0823-cacheON](https://user-images.githubusercontent.com/17171963/186050093-7df330a9-cce2-42a9-bc13-08296b8e75c5.gif)


By observing the progress bar's movement, it seems about 20% of improvement was made..

dayo09:
### TEST: Save cfglist, cfgObjMap, baseModelMap

### Current main branch, latency on refresh (1~2sec)

![0823-cacheOFF](https://user-images.githubusercontent.com/17171963/186050061-6702fe70-2d7c-4d96-9bcd-bb3a7d930e2c.gif)


#### When #1227 is applied, (almost immediate)
![0823-cache-GOOD](https://user-images.githubusercontent.com/17171963/186136683-574f4aed-5ddc-4fa6-9c2b-ea8909318eeb.gif)

dayo09:
The object is reached by the first optimization, so let's close this.

PR #1226, opened 2022-08-23
[OneExplorer] Cache cfgList

dayo09:
This commit caches cfg list and refresh on file system change.

----

> 
> ### TEST: Save cfg list and rebuild the list on refresh
> #### Current main branch, latency on refresh
> 0823-cacheOFF
> 
> #### When the change is applied
> 0823-cacheON
> 
> 
> By observing the progress bar's movement, it seems about 20% of improvement was made..

From https://github.com/Samsung/ONE-vscode/issues/1223#issuecomment-1223417728


dayo09:
Included in #1227

PR #1227, opened 2022-08-23
[OneExplorer] Introduce OneStorage to improve latency

dayo09:
This commit introduces OneStorage module which caches the storage context.
It improves ONE Explorer's latency EFFECTIVELY.

----

For #1223

#### Current main branch, latency on refresh (1~2sec)

0823-cacheOFF

#### With this PR (almost immediate)
0823-cache-GOOD

Also, when we first open the ONE Explorer, it applies the same.

### DIFF
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -163,7 +163,7 @@ export class OneStorage {
     return OneStorage._obj;
   }
 
-  static reset(): void {
+  public static reset(): void {
     OneStorage._obj = undefined;
   }
 }
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -63,7 +63,7 @@ export class OneStorage {
   private _baseModelToCfgsMap: StringMap;
 
   /**
-   * Get the list of .cfg files wiithin the workspace
+   * Get the list of .cfg files within the workspace
    * @param root  the file or directory,
    *              which MUST exist in the file system
    */
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -45,8 +45,8 @@ interface ConfigObjMap {
  *
  * The singleton is created when the first get() is called.
  * The object remains until OneStorage.reset() is called.
- * OneStorage.reset() is called within OneTreeDataProvider.refresh(), which is called on every file system
- * change within the repository.
+ * OneStorage.reset() is called by OneTreeDataProvider.refresh(), which is called on every file
+ * system change within the repository.
  */
 export class OneStorage {
   /**
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -45,7 +45,7 @@ interface ConfigObjMap {
  *
  * The singleton is created when the first get() is called.
  * The object remains until OneStorage.reset() is called.
- * OneStorage.reset() is called OneTreeDataProvider.refresh(), which is called on every file system
+ * OneStorage.reset() is called within OneTreeDataProvider.refresh(), which is called on every file system
  * change within the repository.
  */
 export class OneStorage {
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -69,10 +69,8 @@ export class OneStorage {
    */
   private _getCfgList(root: string = obtainWorkspaceRoot()): string[] {
     /**
-     * Returns every file inside directory
+     * Returns an array of all the file names inside the root directory
      * @todo Check soft link
-     * @param root
-     * @returns
      */
     const readdirSyncRecursive = (root: string): string[] => {
       if (fs.statSync(root).isFile()) {
@@ -89,13 +87,11 @@ export class OneStorage {
     };
 
     try {
-      fs.statSync(root);
+      return readdirSyncRecursive(root).filter(val => val.endsWith('.cfg'));
     } catch {
       Logger.error('OneExplorer', 'getCfgList', 'called on not existing directory or file.');
       return [];
     }
-
-    return readdirSyncRecursive(root).filter(val => val.endsWith('.cfg'));
   }
 
   private _getCfgToCfgObjMap(cfgList: string[]): ConfigObjMap {
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,167 @@
...
+ * To build each 'Node' of OneTreeDataProvider,
+ * it is neccessary to access the file system, read the files and build objects(ConfigObj, ...).
+ * By keeping some file system information as data structure (list, map),
+ * some duplicated works can be reduced.
+ * 
+ * LIFE CYCLE 
+ * 
+ * The singleton is created when the first get() is called.
+ * The object remains until OneStorage.reset() is called.
+ * OneStorage.reset() is called OneTreeDataProvider.refresh(), which is called on every file system change within the repository.
```
dayo09, line 47: For reviewers,

Please read this part.

This is a singleton object to store some file system(==storage) information.
It builds the context only once and exploit the data when OneTreeDataProvider builds the `Node`s.

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+ * By keeping some file system information as data structure (list, map),
+ * some duplicated works can be reduced.
+ *
+ * LIFE CYCLE
+ *
+ * The singleton is created when the first get() is called.
+ * The object remains until OneStorage.reset() is called.
+ * OneStorage.reset() is called OneTreeDataProvider.refresh(), which is called on every file system
+ * change within the repository.
+ */
```
dayo09, line 50: For reviewers,

Please read this part.

This is a singleton object to store some file system(==storage) information.
It builds the context only once and exploit the data when OneTreeDataProvider builds the Nodes.

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+ * To build each 'Node' of OneTreeDataProvider,
+ * it is neccessary to access the file system, read the files and build objects(ConfigObj, ...).
+ * By keeping some file system information as data structure (list, map),
+ * some duplicated works can be reduced.
+ *
+ * LIFE CYCLE
+ *
+ * The singleton is created when the first get() is called.
+ * The object remains until OneStorage.reset() is called.
+ * OneStorage.reset() is called OneTreeDataProvider.refresh(), which is called on every file system
```
llFreetimell, line 48:
```suggestion
 * OneStorage.reset() is called within OneTreeDataProvider.refresh(), which is called on every file system
```

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+   * A map of ConfigObj (key: cfg path)
+   */
+  private _cfgToCfgObjMap: ConfigObjMap;
+  /**
+   * A map of BaseModel path to Cfg path
+   */
+  private _baseModelToCfgsMap: StringMap;
+
+  /**
+   * Get the list of .cfg files wiithin the workspace
```
llFreetimell, line 66:
```suggestion
   * Get the list of .cfg files within the workspace
```

dayo09: Thanks for finding out some typos :-D

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+ * some duplicated works can be reduced.
+ *
+ * LIFE CYCLE
+ *
+ * The singleton is created when the first get() is called.
+ * The object remains until OneStorage.reset() is called.
+ * OneStorage.reset() is called within OneTreeDataProvider.refresh(), which is called on every file system
+ * change within the repository.
+ */
+export class OneStorage {
```
llFreetimell, line 51: This class seems available for unit test but not sure... is it available or not? @.@

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Get a singleton object
+   */
+  private static get(): OneStorage {
+    if (!OneStorage._obj) {
+      OneStorage._obj = new OneStorage;
+    }
+    return OneStorage._obj;
+  }
+
+  static reset(): void {
```
llFreetimell, line 166: Just for consistency? :D

```suggestion
  public static reset(): void {
```

src/OneExplorer/OneStorage.ts
dayo09: Seemingly, it's possible to add some unit tests.
I will add them later on :-D

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Get the list of .cfg files within the workspace
+   * @param root  the file or directory,
+   *              which MUST exist in the file system
+   */
+  private _getCfgList(root: string = obtainWorkspaceRoot()): string[] {
+    /**
+     * Returns every file inside directory
+     * @todo Check soft link
+     * @param root
+     * @returns
```
llFreetimell, line 75:
```suggestion
     * @returns array of files in 'root' directory
```

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    try {
+      fs.statSync(root);
+    } catch {
+      Logger.error('OneExplorer', 'getCfgList', 'called on not existing directory or file.');
+      return [];
+    }
+
+    return readdirSyncRecursive(root).filter(val => val.endsWith('.cfg'));
```
llFreetimell, line 98: (Optional) After I saw `fs.statSync` is in try-catch, I thought following suggestion because `readdirSyncRecursive` also includes `fs.statSync` :)
But not that critical I think :D

```suggestion
    try {
      fs.statSync(root);
      return readdirSyncRecursive(root).filter(val => val.endsWith('.cfg'));
    } catch {
      Logger.error('OneExplorer', 'getCfgList', 'called on not existing directory or file.');
      return [];
    }
```

src/OneExplorer/OneStorage.ts
dayo09: Lemme remove it. It's duplicated of the line 72 :-D

src/OneExplorer/OneStorage.ts
dayo09: I agree :-D Then it seems the line `fs.statSync(root);` can be removed, too. Thanks :-D

llFreetimell (approved):
LGTM

src/OneExplorer/OneStorage.ts
```diff
@@ -0,0 +1,165 @@
...
+   * Get the list of .cfg files within the workspace
+   * @param root  the file or directory,
+   *              which MUST exist in the file system
+   */
+  private _getCfgList(root: string = obtainWorkspaceRoot()): string[] {
+    /**
+     * Returns an array of all the file names inside the root directory
+     * @todo Check soft link
+     */
+    const readdirSyncRecursive = (root: string): string[] => {
```
yunjayh, line 75: NIT, but is there any reason for `readdirSyncRecursive`? May be in separate PR..

```suggestion
    const readDirSyncRecursive = (root: string): string[] => {
```

yunjayh (approved):
LGTM

src/OneExplorer/OneStorage.ts
dayo09: I think I named if after `fs.readdirSync` function :-D

dayo09: Let's leave it for consistency. I will check it again when I replace `fs.readdirSync` with `vscode.workspace.fs.readDirectory`, too.

Issue #1236, opened 2022-08-24
[CfgEditor] Ask to save before run


dayo09:
## What?

Let's ask user whether to save the file before running cfg file.

dayo09:
### Reference
#### Vscode settings: Save before debug/task/run
![0824-save-options](https://user-images.githubusercontent.com/17171963/186411511-9ed87e81-ac5e-4ac3-857b-58a29d394aac.png)

#### Vscode save dialog

![0824-save-prompt](https://user-images.githubusercontent.com/17171963/186411616-7c763fe9-f405-41fd-a4a6-098eedc832ff.png)

dayo09:
- [ ] TODO: Emit a log if there are 2+ unsaved documents for a single file.

PR #1238, opened 2022-08-24
[ToolchainProvider] Ask user to save documents before run

dayo09:
This commit is to ask user to save dirty target document
before running the document as onecc config.

---

For #1236

### DIFF
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -21,8 +21,6 @@ import * as vscode from 'vscode';
 import {Balloon} from './Balloon';
 import {Logger} from './Logger';
 
-var ini = require('ini');
-
 const logTag = 'Helpers';
 
 /**
@@ -100,30 +98,40 @@ export interface FileSelector {
 }
 
 /**
- * Show an information message to ask users whether to save the unsaved file
- * @param filepath a full path to the file
- * @return rejected promises if users don't save the file.
+ * Show an information message to ask users whether to save the unsaved file.
+ *
+ * @param filepath a full path to the file.
+ * @return a Promise returning true, if documents are saved.
+ *         a Promise returning false, if error occurs or user choose not
+ *          to save documents.
  */
-export async function askToSaveFile(filepath: string): Promise<void> {
+export async function saveDirtyDocuments(filepath: string): Promise<boolean> {
   const unsavedDocuments = vscode.workspace.textDocuments.filter(td => {
     if ((td.fileName === filepath) && td.isDirty) {
       return td;
     }
   });
 
-  await Promise.all(unsavedDocuments.map(td => {
-    const title = `Do you want to save the changes you made to ${path.parse(td.fileName).name}?`;
-    const detail = undefined;
-    const ansSave = 'Save';
-    return new Promise((resolve, reject) => {
-      vscode.window.showInformationMessage(title, {detail: detail, modal: true}, ansSave)
-          .then(ans => {
-            if (ans === ansSave) {
-              return resolve(td.save());
-            } else {
-              return reject();
-            }
-          });
+  if (unsavedDocuments.length === 0) {
+    return true;
+  }
+
+  const title = `Do you want to save the changes you made to ${path.parse(filepath).name}?`;
+  const detail = undefined;
+  const ansSave = 'Save';
+  const ans =
+      await vscode.window.showInformationMessage(title, {detail: detail, modal: true}, ansSave);
+
+  if (ans === ansSave) {
+    return Promise.all(unsavedDocuments.map(doc => doc.save())).then(res => {
+      if (res.includes(false)) {
+        Logger.error('Failed to save document');
+        return false;
+      } else {
+        return true;
+      }
     });
-  }));
+  } else {
+    return false;
+  }
 }
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -98,3 +98,32 @@ export function obtainWorkspaceRoot(): string {
 export interface FileSelector {
   onFileSelected(uri: vscode.Uri|undefined): void;
 }
+
+/**
+ * Show an information message to ask users whether to save the unsaved file
+ * @param filepath a full path to the file
+ * @return rejected promises if users don't save the file.
+ */
+export async function askToSaveFile(filepath: string): Promise<void> {
+  const unsavedDocuments = vscode.workspace.textDocuments.filter(td => {
+    if ((td.fileName === filepath) && td.isDirty) {
+      return td;
+    }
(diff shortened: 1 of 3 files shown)


hyunsik-yoon:
I am kinda curious about the role and responsibility of toolchain provider.
Checking unsaved doc and asking user to save the doc might be out of the responsibility of toolchain.

dayo09:
@hyunsik-yoon Yes it looks awkward for me, too. I will find way to put the function in somewhere else. It can be moved into a command block.

hyunsik-yoon:
Some pattern like MVC could be a candidate to think.  Model could be toolchain in our case, View could be vscode and some module to show specific UI and Contorller could be something that calls toolchain and ask vscode to show some screen. :-)

dayo09:
@hyunsik-yoon I learned about MVC pattern and found out that the point you fingered on is not only about this PR but also about the all the other functions which asks things to users directly. 

So here is what I suggest: what if we make a 'controller'(of MVC) or a 'mediator'(of mediator pattern, [related reading](https://refactoring.guru/design-patterns/mediator)) for our project to clarify the responsibility of each modules?

hyunsik-yoon:
Sound like a good plan. Maybe a small time for a discussion or a seminar could help others to understand your direction. :-)

dayo09:
.@hyunsik-yoon Sure :-D And for this PR, I just followed jy's way as much as I can. (put the function as private method and linked it to the public function which also has a dialog function)

src/Utils/Helpers.ts
```diff
@@ -98,3 +98,32 @@ export function obtainWorkspaceRoot(): string {
...
   onFileSelected(uri: vscode.Uri|undefined): void;
 }
+
+/**
+ * Show an information message to ask users whether to save the unsaved file
+ * @param filepath a full path to the file
+ * @return rejected promises if users don't save the file.
+ */
+export async function askToSaveFile(filepath: string): Promise<void> {
+  const unsavedDocuments = vscode.workspace.textDocuments.filter(td => {
```
hyunsik-yoon, line 108: Here, a plural name, `unsavedDocuments`, is used. I am wondering if the length of unsavedDocuments[] is expected to be 1 (or 0 if filepath is not the one in workspace documents) and could you explain the reason of using plural name?

dayo09: Well, I initially worried about having multiple unsaved document editor opened. However I didn't checked if the case really is possible. I will check

dayo09: I inferred so because..
- There is `readonly supportsMultipleEditorsPerDocument?` option for custom editors. 
- By experience, I sometimes see that multiple editors are opened for a single file.

hyunsik-yoon: From https://code.visualstudio.com/api/extension-guides/custom-editors,

```
If the user opens the same resource in another editor group using our custom 
editor—for example by splitting the first editor—the extension's job is simplified. 
In this case, VS Code just calls resolveCustomEditor with the same CustomDocument 
we created when the first editor was opened.
```

/cc @llFreetimell  We may need to consider using `supportsMultipleEditorsPerDocument: True` in cfg editor.

src/Utils/Helpers.ts
```diff
@@ -98,3 +98,32 @@ export function obtainWorkspaceRoot(): string {
...
+      return td;
+    }
+  });
+
+  await Promise.all(unsavedDocuments.map(td => {
+    const title = `Do you want to save the changes you made to ${path.parse(td.fileName).name}?`;
+    const detail = undefined;
+    const ansSave = 'Save';
+    return new Promise((resolve, reject) => {
+      vscode.window.showInformationMessage(title, {detail: detail, modal: true}, ansSave)
```
hyunsik-yoon, line 119: If `unsavedDocuments`'s length is 2, and they are two instance of _one_ text document,
I am afraid that a user would see the same message two times. This might make people feel a bit annoyed.

dayo09: But I think that's what happen we use vscode, for example when I tried to close the workspace, it asks me multiple times about one single document whether to save changes or not.

dayo09: I knew that when we use 'Cfg Edit'r, only one opened 'TextDocument' exists and the changes are synchonized between cfg editor and text editor. However, it's a part of feature, so I believe that it didn't work like that before. Based on that, I thought that some editors don't synchronize its editing material in a view with other text documents - that's why I decided to put 'unsavedDocuments' in an array.

hyunsik-yoon: I understand what's been explained about `unsavedDocuments`. What I am worry is usability.
If two editor object (let's call this o1 and o2) opens the same file, e.g., `foo.txt` and both are modified at different lines,
`unsavedDocuments` will contains two foo.txt and
user will be asked 
"Do you want to save the changes you made to foo.txt"? -->msg 1 
after clicking 'yes'
user will be asked again
"Do you want to save the changes you made to foo.txt"? --> msg 2 (msg string is same with msg 1)
and the second question might confuse the user because user already click yes.

Or another user, who wants to save not o1 but o2, would have a question, like, "which msg is for o2? msg1 or msg2?"

dayo09: Sure, then let me edit the information message to show only once :-D

src/Utils/Helpers.ts
```diff
@@ -98,3 +96,42 @@ export function obtainWorkspaceRoot(): string {
...
+  }
+
+  const title = `Do you want to save the changes you made to ${path.parse(filepath).name}?`;
+  const detail = undefined;
+  const ansSave = 'Save';
+  const ans =
+      await vscode.window.showInformationMessage(title, {detail: detail, modal: true}, ansSave);
+
+  if (ans === ansSave) {
+    return Promise.all(unsavedDocuments.map(doc => doc.save())).then(res => {
```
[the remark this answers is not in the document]
dayo09: @hyunsik-yoon If it's the same 'text editor view', then you don't need to be worried. I tested a bit and looked into the codes, it looks that it maintains only one cache file to save the changes. Thus, in your example, (1) and (2) will be applied immediately to both the editors. It works the same to CfgEditor, when a person put any change, then TextEditor and CfgEditor viewers synchronize the currently change before we save it.

But I understand your point, if a user answered to save multiple documents and we save it without considering the order or priority, it's out of the user's expectation. 

However, we currently cannot find the real example of the 'any editor opens the viewer and keeps its change independently from the other text editor' case. 
How about I add some exception instead, for the multiple unsaved Documents case?

hyunsik-yoon (approved):

jyoungyun (approved):
LGTM

Issue #1241, opened 2022-08-25
[OneExplorer] Support viewer for tflite/onnx


dayo09:
## What?
Let's support viewer for tflite/onnx.
Logically, netron support boths, but it seems that the current circle-dedicated viewer 'one.viewer.circle' is not working with tflite files.

The error message shows up when I tried to open a tflite file with circle viewer.
![image](https://user-images.githubusercontent.com/17171963/186613984-416cc965-e5a4-4c77-8b9c-1778b8e18231.png)

/cc @lemmaa @seanshpark

dayo09:
@seanshpark AFAIK, the circle-netron supports tflite/onnx files. However, it seems that our CircleViewer doesn't support them. Do you think that it's possible to support tflite/onnx files, too?

seanshpark:
> Do you think that it's possible to support tflite/onnx files, too?

Current implementation: NO

How netron supports multiple formats?
- it dynamically loads file loader .js files

How current circle viewer control works?
- static loading of circle file loader codes and some fixed codes based on handing circle files
- has some modification for partition editing that may disturb loading tflite/onnx

So, IMHO, how to support tflite/onnx files?
- for simplicity, add another module for tflite and onnx both or two modules for each

dayo09:
> for simplicity, add another module for tflite and onnx both or two modules for each

Oh, therefore we need to port other modules to support tflite/onnx viewers...  :-D 

/cc @Samsung/one-vscode

PR #1242, opened 2022-08-25
[OneExplorer] View tflite/onnx files

dayo09:
This commit is to show tflite and onnx files in ONE Explorer using the CircleViewer.

----

For #1241


dayo09:
CircleViewer doesn't support tflite and onnx files. We need to add dedicated file viewers for those file types.

PR #1247, opened 2022-08-30
DRAFT: [OneExplorer] Build children on demand

dayo09:
This commit builds child nodes on demand.
It includes
- Change OneTreeDataProvider into type of TreeDataProvider<Node> instead of <OneNode>
  - OneNode is a TreeItem, which is a representation node
  - Node is an 'element', which is a core data structure
  - TreeDataProvider is designed to have 'element' to be 'unseen materials' and 'TreeItem' to be 'shown nodes, which is collapsible'.
    To make our OneExplorer to build children only when expanding the 'TreeItem', we need to make it as a proper TreeDataProvider<Node>.
- Introduce getChildren() function which builds children on demand.
- Do not create 'hidden' extra nodes
  - Change didHideExtra as static, accordingly
- Change getTree() not to build the whole tree but to obtain the root
  - Move workspaceRoot-checking code inside this
- Change getChildren() and getTreeItem()
  - Implant toOneNode() code into getTreeItem()
  - Simplify getChildren() which only returns its child.

-----

To resolve the problem of #1246

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -445,11 +445,10 @@ export class OneNode extends vscode.TreeItem {
 }
 
 /* istanbul ignore next */
-export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
-  private _onDidChangeTreeData: vscode.EventEmitter<OneNode|undefined|void> =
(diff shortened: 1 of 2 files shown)


dayo09:
As it contains too much changes, let's split them.

PR #1248, opened 2022-08-31
[OneExplorer] Do not build hidden nodes

dayo09:
This commit is not to build hidden nodes.
This is required to change the OneTreeDataProvider's element type from OneNode to Node.
The provider will expect every nodes to be correspond to OneNodes, which is a visible tree item.

----

From #1247

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -177,6 +177,10 @@ abstract class Node {
 
 class NodeFactory {
   static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    // WHY HIDDEN NODES ARE NOT TO BE CREATED?
+    //
+    // A 'TreeDataProvider<element>' expects every elements (Node) to be correspond to visible
+    // TreeItem, so let's not build hidden nodes.
     if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
     {
       return undefined;
@@ -185,7 +189,6 @@ class NodeFactory {
     const uri = vscode.Uri.file(fpath);
 
     let node: Node;
-
     if (type === NodeType.directory) {
       assert.strictEqual(attr, undefined, 'Directory nodes cannot have attributes');
       node = new DirectoryNode(uri);
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -181,7 +181,7 @@ class NodeFactory {
     //
     // A 'TreeDataProvider<element>' expects every elements (Node) to be correspond to visible
     // TreeItem, so let's not build hidden nodes.
-    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
+    if(attr && attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
     {
       return undefined;
     }
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -176,10 +176,16 @@ abstract class Node {
 }
 
 class NodeFactory {
-  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node {
+  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
+    {
+      return undefined;
+    }
+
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -176,10 +176,16 @@ abstract class Node {
...
+  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
+    {
+      return undefined;
+    }
+
     const uri = vscode.Uri.file(fpath);
 
     let node: Node;
+
```
dayo09, line 188:
```suggestion
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -176,10 +176,16 @@ abstract class Node {
 }
 
 class NodeFactory {
-  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node {
+  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
```
dayo09, line 180:
```suggestion
    // TreeDataProvider expects every elements (Node) to be correspond to visible TreeItem, so let's not build hidden nodes.
    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -176,7 +176,13 @@ abstract class Node {
 }
 
 class NodeFactory {
-  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node {
+  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    // TreeDataProvider expects every elements (Node) to be correspond to visible TreeItem, so let's not build hidden nodes.
+    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
```
dayo09, line 181: For reviewers,

To refer here, `OneTreeDataProvider.didHideExtra` is changed to static.

src/OneExplorer/OneExplorer.ts
```diff
@@ -176,7 +176,16 @@ abstract class Node {
 }
 
 class NodeFactory {
-  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node {
+  static create(type: NodeType, fpath: string, attr?: ArtifactAttr): Node|undefined {
+    // WHY HIDDEN NODES ARE NOT TO BE CREATED?
+    //
+    // A 'TreeDataProvider<element>' expects every elements (Node) to be correspond to visible
+    // TreeItem, so let's not build hidden nodes.
+    if(attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
```
yunjayh, line 184: Q) How does the code be executed if `attr` parameter is not given?

dayo09: It makes an error, thanks for pointing out :-D

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

yunjayh (approved):
LGTM :D

src/OneExplorer/OneExplorer.ts
```diff
@@ -423,7 +438,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<OneNode> {
   private fileWatcher = vscode.workspace.createFileSystemWatcher(`**/*`);
 
   private tree: DirectoryNode|undefined;
-  public didHideExtra: boolean = false;
+  public static didHideExtra: boolean = false;
```
hyunsik-yoon, line 441: optional) I am not sure whether making this to `static` of `OneTreeDataProvider` is correct approach. To me this var in `OneTreeDataProvider` is not class var but object var.. and it seems like a quick fix to use factory class. :-)
Some thought is... I think we eventually need some object that stores persistent state of ONE vscode and that is loaded (from persistent storage) when a ONE user reloads vscode. `didHideExtra` could be possibly one in the object (maybe not in `OneTreeDataProvider`).  :-)

dayo09: Yes, I agree with that. Until we introduce some kind of extension-wise context manager, we may face the similar cases. :-D Likewise, I set 'one.explorer:didHideExtra' context. Let's talk about that later on.

hyunsik-yoon (approved):

PR #1249, opened 2022-08-31
[OneExplorer] Build children on demand 

dayo09:
This commit makes OneExplorer to build children on demand.

It means, OneExplorer do not know if the node has any children until users click the node to expand.
When users click nodes, then it builds the children nodes.
The latency improvement is expected.

    NOTE
    As baseModel nodes and config nodes need to know its children count,
    they build their children one-level earlier.
    But they don't build the tree until the leaf nodes,
    so the 'on-demand' concept still works on them.

----

Waiting for #1248 to be merged
From #1247

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -139,7 +139,7 @@ enum NodeType {
 
 abstract class Node {
   abstract readonly type: NodeType;
-  childNodes: Node[];
+  protected _childNodes: Node[]|undefined;
   uri: vscode.Uri;
 
   abstract icon: vscode.ThemeIcon;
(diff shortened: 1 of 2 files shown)


dayo09:
@Samsung/one-vscode PTAL :-D

llFreetimell (approved):
LGTM

yunjayh (approved):
LGTM

src/OneExplorer/OneExplorer.ts
```diff
@@ -139,19 +139,31 @@ enum NodeType {
...
+   * Build `_childNodes` on demand, which is initially undefined
+   */
+  abstract _buildChildren: () => void;
+
+  getChildren(): Node[] {
+    if (this._childNodes) {
+      return this._childNodes;
+    }
+
+    this._buildChildren();
```
hyunsik-yoon, line 164: While I was reviewing this line, this file was merged. :-)

`getChilldren()` seems immutable (since it is getter)
but it does mutable work by calling `_buildChildren()`.

I believe that `_buildChildren()` should not be inside this method.

Issue #1253, opened 2022-09-02
[eslint] Add recommended rules


dayo09:
## What?

To enables eslint:recommended extends, let's first remove some violations by adding simple recommended rules.

- For #587

dayo09:
### Sub-task

- [ ] #1254
- [ ] Remove unused variables

dayo09:
After #1254, #1255 is merged,


```

/home/dayo/git/ONE-vscode/src/Backend/Compiler.ts
  80:17  error  'toolchainType' is defined but never used  no-unused-vars
  80:40  error  'start' is defined but never used          no-unused-vars
  80:55  error  'count' is defined but never used          no-unused-vars
  84:26  error  'toolchainType' is defined but never used  no-unused-vars

/home/dayo/git/ONE-vscode/src/Backend/Executor.ts
  53:16  error  '_modelPath' is defined but never used  no-unused-vars
  53:36  error  '_options' is defined but never used    no-unused-vars

/home/dayo/git/ONE-vscode/src/Backend/Toolchain.ts
  61:7  error  'cfg' is defined but never used  no-unused-vars

/home/dayo/git/ONE-vscode/src/CfgEditor/CfgEditorPanel.ts
   76:7   error  '_token' is defined but never used            no-unused-vars
  146:11  error  Unexpected lexical declaration in case block  no-case-declarations
  152:11  error  Unexpected lexical declaration in case block  no-case-declarations
  176:39  error  'e' is defined but never used                 no-unused-vars

/home/dayo/git/ONE-vscode/src/CircleGraph/CircleGraph.ts
  87:38  error  'e' is defined but never used  no-unused-vars

/home/dayo/git/ONE-vscode/src/CircleGraph/CircleGraphCtrl.ts
  345:21  error  'err' is defined but never used   no-unused-vars
  366:26  error  Unnecessary escape character: \%  no-useless-escape

/home/dayo/git/ONE-vscode/src/CircleGraph/CircleViewer.ts
  127:24  error  'openContext' is defined but never used  no-unused-vars
  128:7   error  '_token' is defined but never used       no-unused-vars
  142:7   error  '_token' is defined but never used       no-unused-vars

/home/dayo/git/ONE-vscode/src/Execute/DeviceViewProvider.ts
   24:9   error  'Balloon' is defined but never used  no-unused-vars
  153:44  error  'reject' is defined but never used   no-unused-vars

/home/dayo/git/ONE-vscode/src/Job/Job.ts
[...]

PR #1254, opened 2022-09-02
[lint] Remove unnecessary semicolons

dayo09:
This commit removes unnecessary semicolons.

---

For #1253

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,6 +14,7 @@
       "curly": "warn",
       "eqeqeq": "warn",
       "no-throw-literal": "warn",
-      "semi": "off"
+      "semi": "off",
+      "no-extra-semi": "warn"
   }
 }
--- a/media/PartEditor/index.js
+++ b/media/PartEditor/index.js
@@ -202,7 +202,7 @@ editor.Editor = class {
         this.backends.push(new editor.Backend(backend, becolor));
         this.beToCode[backend] = idx;
       }
-    };
+    }
 
     this.updateBeComboColor();
     this.requestOpNames();
@@ -234,7 +234,7 @@ editor.Editor = class {
         // becode will be updated after document partition is received
         this.operators.push(new editor.Operator(name, opcode, 0));
       }
-    };
+    }
 
     this.requestPartition();
   }
@@ -270,7 +270,7 @@ editor.Editor = class {
       let opt = listbox.options[i];
       let idx = opt.value;
       opt.selected = (selection.includes(this.operators[idx].name));
-    };
+    }
   }
 
   getSelectionNames() {
@@ -368,7 +368,7 @@ editor.Editor = class {
           this.setOperatorBeCode(name, 0);
         }
       }
-    };
+    }
   }
 
   /**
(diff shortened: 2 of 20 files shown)


yunjayh (approved):
LGTM

hyunsik-yoon:
In C++ we put semicolon at the end of class declaration. I wonder the reason why we need to remove semicolon after class declaration in TS.

dayo09:
@hyunsik-yoon It's a redundant semicolon for TS language. Therefore, eslint gives an error. 

https://eslint.org/docs/latest/rules/no-extra-semi

I referred eslint rules, you may also be able to find typescript grammars :-D

dayo09:
NOTE: This is done with eslint --fix option, as unncessary semicolon error is almost the only fixable by that. :-D

hyunsik-yoon:
> Therefore, eslint gives an error.
...
> you may also be able to find this in typescript language specification, too 

The following lines in TS are all valid and work: ([link](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8C+1oHpPQHZIgFMBbAS2HhHlwFgAoC3CSwgOioHMAKAcgDNEPAJQZseAiXKVq0AC6J6oSDABCYAE4pUAblE4iZClVoNqzEG068ARhuG6sOfNANTjchXSA))

```ts
class Foo {}  // 1. no semicolon
console.log('foo')  // 2. no semicolon too
class Foo {};  // 3. semicolon
console.log('foo');  // 4. no semicolon too
```

I am personally not against this PR. :-)

I just want to know any proof to approve this PR. 

Maybe providing some reference URL or screen capture of error from eslint could be good evidence to persuade other people who need to review.

dayo09:
@hyunsik-yoon 
It removes not only the semicolons at the end of class definition but the floating (literally unused) semicolons.
Here, it removes semicolons just because they are 'redundant'.

(plus, As far as I know, javascript's function does not need to end with ';' but it's recommended to put semicolons.)

The playground doesn't emit every warnings as eslint does.  Naturally, eslint does more strict checking. For example, it doesn't emit any error on every whitespace but we run 'clang-format' to remove them..

Eslint offers us a 'recommended' checker option. When it's turned on, 'unused semicolon' option is on.
I think it's has the same quality as we remove redundant white spaces.

hyunsik-yoon:
I think that sometimes providing useful links could give more objective view.
So I tried to find why `;` after class in C++ is useful and https://stackoverflow.com/questions/1783465/why-must-i-put-a-semicolon-at-the-end-of-class-declaration-in-c mentioned that `;` is very useful in C++. However, grammar in TS is different from C++ and I now also think that `;` after TS class seems redundant.

hyunsik-yoon (approved):

llFreetimell:
@dayo09 @hyunsik-yoon 

At first, I had same thoughts with @hyunsik-yoon, which is "why we should remove those semicolons? is there any official guide for this?"
In the point of view, https://eslint.org/docs/latest/rules/no-extra-semi seems to be about "style", not "guide".

But I could find https://www.typescriptlang.org/docs/handbook/2/classes.html and the official example is following current PR's policy.
So now I think we can merge this PR with confidence :)

PR #1256, opened 2022-09-06
[lint] Remove no-case-declarations

dayo09:
This commit adds "no-case-declarations" rule to our lint configuration and
fixes the errors.

----

For #1253

> 
> ### Remove no-case-declarations
> 
> https://eslint.org/docs/latest/rules/no-case-declarations
> 
> #### Why is this a problem?
> >  The reason is that the lexical declaration is visible in the entire switch block but it only gets initialized when it is assigned, which will only happen if the case where it is defined is reached.
> 
> #### Sol
> > To ensure that the lexical declaration only applies to the current case clause wrap your clauses in blocks.
> 
> 
> ### 2 Errors
> ```
> /home/dayo/git/ONE-vscode/src/CfgEditor/CfgEditorPanel.ts
>   146:11  warning  Unexpected lexical declaration in case block  no-case-declarations
>   152:11  warning  Unexpected lexical declaration in case block  no-case-declarations
> 
> ✖ 2 problems (0 errors, 2 warnings)
> ```
> 
>

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,6 +15,7 @@
       "eqeqeq": "warn",
       "no-throw-literal": "warn",
       "semi": "off",
-      "no-extra-semi": "warn"
+      "no-extra-semi": "warn",
+      "no-case-declarations": "warn"
   }
 }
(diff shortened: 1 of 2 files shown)


jyoungyun (approved):
LGTM

llFreetimell (approved):
LGTM

PR #1258, opened 2022-09-06
[lint] Remove no-control-regex

dayo09:
This commit enables 'no-control-regex' rule
and removes the errors accordingly.

---

For #1253

> 
> ### Remove no-control-regex
> 
> #### Reference
> > https://stackoverflow.com/questions/14313183/javascript-regex-how-do-i-check-if-the-string-is-ascii-only
> 
> #### Problematic lines
> 
> ```typescript
> // NOTE ASCII characters have codes ranging from u+0000 to u+007f
> function containsNonAscii(str: string): boolean {
>   return !/^[\u0000-\u007f]*$/.test(str);
> }
> ```
> 
> The line is required to check if it's an ascii, but how about checking 'printable ascii' aside from those problematic non-printables?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
       "eqeqeq": "warn",
       "no-throw-literal": "warn",
       "semi": "off",
+      "no-control-regex": "warn",
       "no-extra-semi": "warn",
       "no-case-declarations": "warn"
   }
--- a/src/View/PasswordQuickInput.ts
+++ b/src/View/PasswordQuickInput.ts
@@ -18,6 +18,7 @@ import * as vscode from 'vscode';
 
 // NOTE ASCII characters have codes ranging from u+0000 to u+007f
 function containsNonAscii(str: string): boolean {
+  // eslint-disable-next-line no-control-regex
   return !/^[\u0000-\u007f]*$/.test(str);
 }
 
--- a/src/View/PasswordQuickInput.ts
+++ b/src/View/PasswordQuickInput.ts
@@ -16,7 +16,18 @@
 
 import * as vscode from 'vscode';
 
-// NOTE ASCII characters have codes ranging from u+0000 to u+007f
+// NOTE
+//
+//   u+0000-u+001f: control ascii codes,    ascii (ALLOWED)
+//   u+0020-u+007f: printable ascii codes,  ascii (ALLOWED)
+//   u+0080-u+00ff: extended ascii codes,   extended ascii (NOT ALLOWED)
+//
+// By experience, Linux os takes all extended ascii characters(u+0000-u+00ff) for passwords, even
+// controll asciis. whereas 'eslint' recommends not using regex including control ascii codes.
+// Therefore, we allow users to put control ascii codes.
+//
+// By experience, the passwords including extended ascii characters are not working when loging in
+// to remote vscode. Therefore, here we allow only 'ascii characters'(0x00-0x7f).
 function containsNonAscii(str: string): boolean {
   // eslint-disable-next-line no-control-regex
   return !/^[\u0000-\u007f]*$/.test(str);
--- a/src/View/PasswordQuickInput.ts
+++ b/src/View/PasswordQuickInput.ts
@@ -18,15 +18,15 @@ import * as vscode from 'vscode';
 
 // NOTE
 //
-//   u+0000-u+001f: control ascii codes,    ascii (ALLOWED)
+//   u+0000-u+001f: control ascii codes,    ascii (ALLOWED) *
 //   u+0020-u+007f: printable ascii codes,  ascii (ALLOWED)
-//   u+0080-u+00ff: extended ascii codes,   extended ascii (NOT ALLOWED)
+//   u+0080-u+00ff: extended ascii codes,   extended ascii (NOT ALLOWED) **
 //
-// By experience, Linux os takes all extended ascii characters(u+0000-u+00ff) for passwords, even
-// controll asciis, whereas 'eslint' recommends not to use the regex which includes any control ascii.
-// Therefore, we allow users to put control ascii codes.
+// *By experience, Linux os takes all extended ascii characters(u+0000-u+00ff) for passwords, even
+// controll asciis, whereas 'eslint' recommends not to use the regex which includes any control
+// ascii. Therefore, we allow users to put control ascii codes.
 //
-// By experience, the passwords including extended ascii characters are not working when loging in
+// **By experience, the passwords including extended ascii characters are not working when loging in
 // to remote vscode. Therefore, here we allow only 'ascii characters'(0x00-0x7f).
 function containsNonAscii(str: string): boolean {
   // eslint-disable-next-line no-control-regex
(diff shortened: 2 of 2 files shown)


dayo09:
I think that \n, \r and \t characters are in non-printable ascii and the tests includes them. I guess any password cannot contain them, so we may be able to remove those characters from the tests.

.eslintrc.json
```diff
@@ -24,6 +24,7 @@
       "eqeqeq": "warn",
       "no-throw-literal": "warn",
       "semi": "off",
+      "no-control-regex": "warn",
```
dayo09, line 27: I put this above 'no-extra-semi' to avoid possible merge conflict

src/Tests/View/PasswordQuickInput.test.ts
```diff
@@ -16,24 +16,24 @@
...
-        assert.isNotTrue(containsNonAscii(strWithAlphabets));
+        assert.isTrue(containsOnlyPrintableAscii(strWithAlphabets));
 
         const strWithNum = '0123456789';
-        assert.isNotTrue(containsNonAscii(strWithNum));
+        assert.isTrue(containsOnlyPrintableAscii(strWithNum));
 
-        const strWithSpecialChar = '\n\t\r!@#$%^&*(){}[]\\';
-        assert.isNotTrue(containsNonAscii(strWithSpecialChar));
+        const strWithSpecialChar = '!@#$%^&*(){}[]\\';
```
dayo09, line 32: I removed \n\t\r from the test.

Reasoning: 

(1) \r \n,
They are CR, LR keys which are 'new line' characters based on each os. 
> CR(\r), LF(\n)](https://m.blog.naver.com/taeil34/221325864981)
:When I hit the 'enter' key, the password quick pick returns, so they are not expected to be included in passwords.
(2) \t
When I hit the 'tab' key, the password quick pick disappears, so it is not expected to be included in passwords.

src/View/PasswordQuickInput.ts
```diff
@@ -16,9 +16,9 @@
 
 import * as vscode from 'vscode';
 
-// NOTE ASCII characters have codes ranging from u+0000 to u+007f
-function containsNonAscii(str: string): boolean {
-  return !/^[\u0000-\u007f]*$/.test(str);
+// NOTE Printable ASCII characters have codes ranging from u+0000 to u+007f
+function containsOnlyPrintableAscii(str: string): boolean {
```
dayo09, line 20: From clean code's view,
I changed negation from the function to avoid 'double negation' 

> https://refactoring.com/catalog/removeDoubleNegative.html

src/Tests/View/PasswordQuickInput.test.ts
jyoungyun: Is there any possibility that the key value will be entered through copy and paste?

dayo09: Oh.. sure. I tried a random password with tab and space, and it worked by pasting in.

dayo09: #### Allowiing \t \n by the rule:

It seems \t \n are allows by the rule even though they are control characters.

https://eslint.org/docs/latest/rules/no-control-regex

> The following elements of regular expression patterns are considered possible errors in typing and are therefore disallowed by this rule:

> Hexadecimal character escapes from \x00 to \x1F.
> Unicode character escapes from \u0000 to \u001F.
> Unicode code point escapes from \u{0} to \u{1F}.
> Unescaped raw characters from U+0000 to U+001F.
> Control escapes such as \t and \n are allowed by this rule.

#### More to consideration

It seems 0x7f is a special character(delete) too. 

https://www.geeksforgeeks.org/control-characters/

dayo09: So let me add \n, \r and \t to the regex and see if they are allowed. The other control characters look not to allow from security's view, IMO.

dayo09: ❗ 

I tried with ascii 0x1,0x2 and even 0x18(Cancel) and linux os still accept all those control ascii characters.
```
0x1 Start of Heading
0x2  Start of Text
0x18 Cancel
```

dayo09: I tried with 0xff(latin y) 0x80(euro sign), which are extended ascii. They are accepted as a linux passwd but I couldn't login to vscode remote with those passwords

src/View/PasswordQuickInput.ts
```diff
@@ -31,7 +32,7 @@ async function showPasswordQuickInput(): Promise<string|undefined> {
 
   async function validateInputIsAscii(input: string): Promise<string|undefined> {
     await new Promise(resolve => setTimeout(resolve, 1000));
-    return containsNonAscii(input) ? 'Password contains non-ASCII characters' : undefined;
+    return containsNonAscii(input) ? 'Password contains non-ASCII(0x00-0x7f) characters' : undefined;
```
dayo09, line 35:
```suggestion
    return containsNonAscii(input) ? 'Password contains non-ASCII characters' : undefined;
```

src/View/PasswordQuickInput.ts
```diff
@@ -16,8 +16,20 @@
...
+//   u+0000-u+001f: control ascii codes,    ascii (ALLOWED)
+//   u+0020-u+007f: printable ascii codes,  ascii (ALLOWED)
+//   u+0080-u+00ff: extended ascii codes,   extended ascii (NOT ALLOWED)
+//
+// By experience, Linux os takes all extended ascii characters(u+0000-u+00ff) for passwords, even
+// controll asciis. whereas 'eslint' recommends not using regex including control ascii codes.
+// Therefore, we allow users to put control ascii codes.
+//
+// By experience, the passwords including extended ascii characters are not working when loging in
+// to remote vscode. Therefore, here we allow only 'ascii characters'(0x00-0x7f).
```
dayo09, line 30: @jyoungyun I tried several tests under what you've mentioned, I summarized here what I've found out. It seems that leaving the codes as-is is enough for now. PTAL :-D

src/View/PasswordQuickInput.ts
```diff
@@ -16,8 +16,20 @@
...
 
-// NOTE ASCII characters have codes ranging from u+0000 to u+007f
+// NOTE
+//
+//   u+0000-u+001f: control ascii codes,    ascii (ALLOWED)
+//   u+0020-u+007f: printable ascii codes,  ascii (ALLOWED)
+//   u+0080-u+00ff: extended ascii codes,   extended ascii (NOT ALLOWED)
+//
+// By experience, Linux os takes all extended ascii characters(u+0000-u+00ff) for passwords, even
+// controll asciis. whereas 'eslint' recommends not using regex including control ascii codes.
```
dayo09, line 26:
```suggestion
// controll asciis, whereas 'eslint' recommends not to use the regex which includes any control ascii.
```

dayo09:
it's ready. PTAL :-D

llFreetimell (approved):

PR #1259, opened 2022-09-06
[lint] Remove no-prototype-builtins

dayo09:
This commit adds "no-prototype-builtins" rule and
fixes errors accordingly.

---

For #1253 

> 
> ### Remove no-prototype-builtins
> #### Reference
> 
> > https://eslint.org/docs/latest/rules/no-prototype-builtins
> 
> > In ECMAScript 5.1, Object.create was added, which enables the creation of objects with a specified [[Prototype]]. Object.create(null) is a common pattern used to create objects that will be used as a Map. This can lead to errors when it is assumed that objects will have properties from Object.prototype. This rule prevents calling some Object.prototype methods directly from an object.
> > 
> > Additionally, objects can have properties that shadow the builtins on Object.prototype, potentially causing unintended behavior or denial-of-service security vulnerabilities. For example, it would be unsafe for a webserver to parse JSON input from a client and call hasOwnProperty directly on the resulting object, because a malicious client could send a JSON value like {"hasOwnProperty": 1} and cause the server to crash.
> > 
> > To avoid subtle bugs like this, it’s better to always call these methods from Object.prototype. For example, foo.hasOwnProperty("bar") should be replaced with Object.prototype.hasOwnProperty.call(foo, "bar").
> 
> #### Example
> 
> > 
> > Examples of incorrect code for this rule:
> > var hasBarProperty = foo.hasOwnProperty("bar");
> > ....
> > Examples of correct code for this rule:
> > var hasBarProperty = Object.prototype.hasOwnProperty.call(foo, "bar");

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -26,6 +26,7 @@
       "semi": "off",
       "no-extra-semi": "warn",
       "no-case-declarations": "warn",
-      "no-useless-escape" : "warn"
+      "no-useless-escape" : "warn",
+      "no-prototype-builtins": "warn"
   }
 }
--- a/media/CircleGraph/view.js
+++ b/media/CircleGraph/view.js
@@ -581,7 +581,7 @@ view.View = class {
         this._clearSelection();
 
         // set new selection
-        if (selection.hasOwnProperty('names')) {
+        if (Object.prototype.hasOwnProperty.call(selection, 'names')) {
             // selection is by names
             const names = selection.names;
             let scrollToSelects = [];  // elements to make visible by scroll to
@@ -643,7 +643,7 @@ view.View = class {
         //      (3) interate all graph nodes and set if exist in opname, clear if not.
         //      here, (3) is implemented
 
-        if (!message.hasOwnProperty('partition')) {
+        if (!Object.prototype.hasOwnProperty.call(message, 'partition')) {
             return;
         }
         const partition = message.partition;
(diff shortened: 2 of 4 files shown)


media/CircleGraph/view.js
```diff
@@ -9,7 +9,7 @@
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
dayo09, line 12:
```suggestion
 * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```

seanshpark (approved):
LGTM thank you!

dayo09:
@Samsung/one-vscode PTAL :-D

mhs4670go (approved):
LGTM

mhs4670go:
it has some confilct.

.eslintrc.json
```diff
@@ -25,6 +25,8 @@
       "no-throw-literal": "warn",
       "semi": "off",
       "no-extra-semi": "warn",
-      "no-case-declarations": "warn"
+      "no-case-declarations": "warn",
+      "no-prototype-builtins": "warn"
+
```
mhs4670go, line 30:
```suggestion
```
extra space?

dayo09:
@mhs4670go @seanshpark It's rebased. PTAL :-D

seanshpark (approved):

mhs4670go (approved):

PR #1262, opened 2022-09-08
[res] Add example circle file

llFreetimell:
This commit adds following example circle file.
- has `While` operator, which consists of multiple subgraphs
- has `BatchMatMulV2`, which is custom operator


mhs4670go:
I think you didn't include the circle file of custom operator.

mhs4670go (approved):
LGTM

PR #1261, opened 2022-09-08
DRAFT: [lint] Remove no-unused-vars warnings

dayo09:
This commit removes 'no-unused-vars' warnings

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

---

For #1253

PR #1267, opened 2022-09-14
[lint] Remove unused variables from Toolchain

dayo09:
This commit removes unused variables and parameters from Toochain.

----

From #1261 
For #1253

### DIFF
--- a/src/Backend/Toolchain.ts
+++ b/src/Backend/Toolchain.ts
@@ -58,7 +58,7 @@ class Toolchain {
   installed(): Command {
     throw Error('Invalid installed call');
   }
-  run(cfg: string): Command {
+  run(_cfg: string): Command {
     throw Error('Invalid run call');
   }
 }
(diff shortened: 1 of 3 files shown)


llFreetimell (approved):
LGTM

src/Backend/Toolchain.ts
```diff
@@ -58,7 +58,7 @@ class Toolchain {
   installed(): Command {
     throw Error('Invalid installed call');
   }
-  run(cfg: string): Command {
+  run(_cfg: string): Command {
```
dayo09, line 61: NOTE: It seems that those classes designed as a super class don't need to follow the rule, because they indicaterequired parameters. So I plan to put some underscore and add the rule 

```
{ "argsIgnorePattern": "^_" }
```
in eslint.

mhs4670go (approved):
LGTM

PR #1274, opened 2022-09-15
[lint] Remove unused generics from MultiStepInput

dayo09:
This commit removes unused generics from MultiStepInput.

----

From #1261 
For #1253

### DIFF
--- a/src/Utils/MultiStepInput.ts
+++ b/src/Utils/MultiStepInput.ts
@@ -57,11 +57,13 @@ interface InputBoxParameters {
 
 /* istanbul ignore next */
 class MultiStepInput {
+  // eslint-disable-next-line no-unused-vars
   static async run<T>(start: InputStep) {
     const input = new MultiStepInput();
     return input.stepThrough(start);
   }
 
+  // eslint-disable-next-line no-unused-vars
   static async runSteps<T>(steps: InputStep[]) {
     if (steps.length === 0) {
       throw new Error('not enough steps');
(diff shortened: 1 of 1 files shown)


src/Utils/MultiStepInput.ts
```diff
@@ -57,12 +57,12 @@ interface InputBoxParameters {
 
 /* istanbul ignore next */
 class MultiStepInput {
-  static async run<T>(start: InputStep) {
+  static async run(start: InputStep) {
```
dayo09, line 60: @jyoungyun PTAL :-D I guess this generic is not used at all.

jyoungyun: As I know, this code came from the vscode example. Do we have to fix?

dayo09: @jyoungyun I think the examples are not the best practices. I also found some unnecessary codes or a bit weird implementations.

jyoungyun: I'm not saying that the example is perfect, but I mean that it's easier to manage the code in future not to modify the upstream code. If a change occurs in the upstream code, we will have to consider more things rather than apply them as they are. :)

dayo09: But I doubt there will be an update related to that generic. The example repository does look like appending more example codes for other devs, rather than maintaining its existing project... aside from the fact that it's not been updated since 2020 May.

And if we're really gonna use the generics, then we can adapt that on the spot.

src/Utils/MultiStepInput.ts
```diff
@@ -57,11 +57,13 @@ interface InputBoxParameters {
 
 /* istanbul ignore next */
 class MultiStepInput {
+  // eslint-disable-next-line no-unused-vars
```
dayo09, line 60: @jyoungyun PLAL :-D

jyoungyun (approved):
LGTM Thank you :)

PR #1282, opened 2022-09-20
DRAFT : [Circle Editor] GUI Updates

rudalsd:
### The following are the main changes on the GUI:

1. Relocated the edit buttons from top of each section to the end of each input areas.

2. When clicking the edit button, it is now designed to allow users to change each input areas instead of the whole section.

### These are our current progress:

1. Modified Attributes Edit feature
 
2. Classified I/O Edit
 
3. Copied Gircle Graph directory to Circle Graph Original to detach the original viewer features from the editor features we are working on
 
4. Convert buffers into binary format when partially modified
 
5. Finalized the commands to use with vscode APIs when modifying circle file


llFreetimell:
Following things should be considered.

- Attributes
  - Attribute keys for custom operators should be modifiable
  - Attributes for custom operators should addable and removable
- Inputs/Outputs
  - `type` and `values` should be modifiable

And these are not urgent, but better to consider.

- Changing operation type may be needed
- `edit` and `show`/`less` can be substituted to icon

---

FYI, I am reviewing with a perspective that supporting GUI should be equal with editing JSON.
Let's imagine that circle editor shows JSON string, which users can do almost everything like changing operation type, values and so on by editing the string.
Then, which would be more useful to users? :)

llFreetimell:
Overall design looks good :)
js files are updated too?

llFreetimell:
Could you check following bugs? :)

- After changing attribute value and click save, box is removed.
- After changing the shape, graph is not updated
- Attribute key for custom operator cannot be changed
- Buffer values are not saved even after values are added

llFreetimell:
Could you check following bugs? :)

- After changing the shape, graph is not updated
- Attribute key for custom operator cannot be changed
- Buffer values are not saved even after values are added

Issue #1271, opened 2022-09-15
[OneToolchain]  Introduce OneToolchain 


YeonJeongLee00:
### What
Let’s introduce **OneToolchain**.

### Why

Currently, the toolchain cannot be installed unless Backend extensions are installed. `OneToolchain` will be installed without a Backend extension to support the use of `run onecc` without a backend.

#### Scenario Video
[Scenario Video](https://user-images.githubusercontent.com/67946956/190300606-050fd78c-0abe-4eaf-83a1-1fb013ed1ca7.webm)

#### Step by step
1. click the `+` button to select a backend (in this scenario it means ONEToolchain; implemetation of Backend interface)

![1](https://user-images.githubusercontent.com/67946956/190300135-bc2cb46a-c12f-41fc-8dfc-f1cc19373b2f.PNG)

2. Select Toolchain type (We’re going to support only the “latest” type for now)  

![2](https://user-images.githubusercontent.com/67946956/190300182-f8ed634c-2223-4220-a88d-561b8d6300f4.PNG)

3. Select Toolchain version (latest version)

![3](https://user-images.githubusercontent.com/67946956/190300199-1e082276-e910-40fa-ae67-d619e8a6f616.PNG)

### TODO

- [x] #1289 
    - [x]  name() (maybe,  OneToolchain)
    - [x] `ToolchainCompiler` (temporary name) implementation based on Compiler interface
        - [x]  getToolchains()
        - [x]  getInstalledToolchains()
        - [x]  prerequsitedsForGetToolchains()
            - [x]  shell script  (for preparing to use toolchains)
                - [x]  add ppa address to /etc/apt/source.list
                - [x]  install onecc-docker(temporary name)
    - [x] modify DebianToolchain (to run onecc-docker)
    - [x] Add OneToolchain Guide in Tutorial.md
    - [x] #1285
    

    onecc-docker : [https://github.com/Samsung/ONE/issues/9697#issuecomment-1244794823](https://github.com/Samsung/ONE/issues/9697#issuecomment-1244794823)
   
    related issue : #1265 

/cc **[@Samsung/ootpg](https://github.com/orgs/Samsung/teams/ootpg)**

jyoungyun:
I think the toolchain name should be `ONE` without `Toolchain`. 

I agree to start by installing the `onecc-docker` debian package as the first step. Let's think about how to use the docker after `onecc-docker` is well applied. :)

lettersto:
[...]

PR #1283, opened 2022-09-20
[ONEToolchain] Introduce ONEToolchain

Psalmist-KIM:
This will introduce ONEToolchain class as Backend implementations.

Related Issue : #1271 

ONE-vscode-DCO-1.0-Signed-off-by: Nakhyun Kim [dss02094@gmail.com](mailto:dss02094@gmail.com)

### DIFF
--- a/src/Tests/Toolchain/ToolchainProvider.test.ts
+++ b/src/Tests/Toolchain/ToolchainProvider.test.ts
@@ -26,6 +26,7 @@ import {BackendNode, BaseNode, NodeBuilder, ToolchainNode, ToolchainProvider} fr
 import {MockCompiler, MockCompilerWithMultipleInstalledToolchains, MockCompilerWithNoInstalledToolchain} from '../MockCompiler';
 
 suite('Toolchain', function() {
+  const oneBackendName = 'ONE';
   const compiler = new MockCompiler();
   const toolchainEnv = new ToolchainEnv(compiler);
   const backendName = 'dummy_backend';
@@ -95,19 +96,22 @@ suite('Toolchain', function() {
     suite('#createBackendNodes()', function() {
       test('creates BackendNode list', function() {
         let bnodes: BackendNode[] = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
       });
     });
     suite('#createToolchainNodes()', function() {
       test('creates ToolchainNode list', function() {
         let bnodes: BackendNode[] = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
-        let bnode: BackendNode = bnodes[0];
-        let tnodes = NodeBuilder.createToolchainNodes(bnode);
-        assert.strictEqual(tnodes.length, 1);
-        tnodes.forEach((tnode) => {
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
+
+        let bnode2: BackendNode = bnodes[1];
+        let tnodes2 = NodeBuilder.createToolchainNodes(bnode2);
+        assert.strictEqual(tnodes2.length, 1);
+        tnodes2.forEach((tnode) => {
           assert.strictEqual(tnode.backendName, backendName);
         });
       });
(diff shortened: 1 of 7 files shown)


jyoungyun:
What about adding `ONEToolchain.test.ts` file?

lettersto:
> What about adding `ONEToolchain.test.ts` file?

That's a good idea!

Psalmist-KIM:
Since ONEToolchain is also registered as a backend, I think that the test cases checked in the existing test code (`ToolchainProvider.test.ts`, `Backend.test.ts`) should be checked for both backends `ONEToolchain` and `MockupBackend`. 

I thought of two ways below.
1. Check both 'ONEToolchain' and 'MockupBackend' in the existing ToolchainProvider and Backend related tests, and in a situation where only ONEToolchain is registered in ONEToolchain.test.ts, select and write additional test codes required regardless of ToolchainProvider and Backend.

2. write additional tests and test cases related to ToolchainProvider and Backend also.

If we proceed as in step 1, the current test code specifies a random backend name and there is a code that compares the expected value with the actual value, so it seems necessary to branch according to the backend name. In this case, I would like to ask for advice on how to write the test.

jyoungyun:
I prefer step 1. I think you can keep this PR changes as it is and just add new `ONEToolchian.test.ts` file. This file will be tested for function implementations such as `prerequisitesForGetToolchains`, `getToolchainTypes`, etc. Currently, there is no implementation about those functions, so you only need to test the backend name.

src/Tests/Toolchain/ToolchainProvider.test.ts
```diff
@@ -170,17 +174,19 @@ suite('Toolchain', function() {
...
       test('gets Children with BackendNode', function(done) {
         let provider = new ToolchainProvider();
         let bnodes: BackendNode[] = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
-        let bnode: BackendNode = bnodes[0];
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
+        let bnode: BackendNode = bnodes[1];
```
jyoungyun, line 189:
```suggestion
        // Ignore bnodes[0] because it is ONE Toolchain backend.
        let bnode: BackendNode = bnodes[1];
```

src/Tests/Toolchain/ToolchainProvider.test.ts
```diff
@@ -95,29 +96,32 @@ suite('Toolchain', function() {
...
-        assert.strictEqual(bnodes[0].label, backendName);
-        let bnode: BackendNode = bnodes[0];
-        let tnodes = NodeBuilder.createToolchainNodes(bnode);
-        assert.strictEqual(tnodes.length, 1);
-        tnodes.forEach((tnode) => {
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
+
+        let bnode2: BackendNode = bnodes[1];
```
jyoungyun, line 111:
```suggestion
        // Ignore bnodes[0] because it is ONE Toolchain backend.
        let bnode2: BackendNode = bnodes[1];
```

src/Tests/Toolchain/ToolchainProvider.test.ts
```diff
@@ -95,29 +96,32 @@ suite('Toolchain', function() {
...
       });
     });
     suite('#createToolchainNodes()', function() {
       test('NEG: creates ToolchainNode list using invalid backend node', function() {
         const bnodes: BackendNode[] = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
-        const tnodes1 = NodeBuilder.createToolchainNodes(bnodes[0]);
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[1].label, backendName);
```
jyoungyun, line 123:
```suggestion
        assert.strictEqual(bnodes[0].label, oneBackendName);
        assert.strictEqual(bnodes[1].label, backendName);
```

I add this line for code consistency because you added this line elsewhere.

src/Tests/Toolchain/ToolchainProvider.test.ts
```diff
@@ -226,9 +232,10 @@ suite('Toolchain', function() {
       test('requests uninstall', function() {
         const provider = new ToolchainProvider();
         const bnodes = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
-        const tnodes = NodeBuilder.createToolchainNodes(bnodes[0]);
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
+        const tnodes = NodeBuilder.createToolchainNodes(bnodes[1]);
```
jyoungyun, line 238:
```suggestion
        // Ignore bnodes[0] because it is ONE Toolchain backend.
        const tnodes = NodeBuilder.createToolchainNodes(bnodes[1]);
```

src/Tests/Toolchain/ToolchainProvider.test.ts
```diff
@@ -271,9 +278,10 @@ suite('Toolchain', function() {
       test('request setDefaultToolchain', function() {
         const provider = new ToolchainProvider();
         const bnodes = NodeBuilder.createBackendNodes();
-        assert.strictEqual(bnodes.length, 1);
-        assert.strictEqual(bnodes[0].label, backendName);
-        const tnodes = NodeBuilder.createToolchainNodes(bnodes[0]);
+        assert.strictEqual(bnodes.length, 2);
+        assert.strictEqual(bnodes[0].label, oneBackendName);
+        assert.strictEqual(bnodes[1].label, backendName);
+        const tnodes = NodeBuilder.createToolchainNodes(bnodes[1]);
```
jyoungyun, line 284:
```suggestion
        // Ignore bnodes[0] because it is ONE Toolchain backend.
        const tnodes = NodeBuilder.createToolchainNodes(bnodes[1]);
```

PR #1299, opened 2022-09-26
[lint] Add no-unused-vars rule

dayo09:
This commit adds no-unused-vars rule to linter.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

---

Waiting for #1274
Waiting for #1300

PR #1300, opened 2022-09-26
[lint] Remove unused variables from .js files

dayo09:
This commit removes unused variables from .js files.

---

For #1299 
From #1253

### DIFF
--- a/media/CircleGraph/index.js
+++ b/media/CircleGraph/index.js
@@ -492,7 +492,7 @@ host.BrowserHost = class {
         this._view.setSelection(message);
     }
 
-    _msgColorTheme(message) {
+    _msgColorTheme(_message) {
         this._view.updateThemeColor();
         this._view.applyThemeColor();
     }
@@ -529,7 +529,7 @@ host.BrowserHost = class {
         this._document.head.appendChild(style);
     }
 
-    _msgReload(message) {
+    _msgReload(_message) {
         this._modelData = [];
 
         this._view.reset();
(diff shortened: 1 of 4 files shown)


media/CircleGraph/index.js
```diff
@@ -492,7 +492,7 @@ host.BrowserHost = class {
         this._view.setSelection(message);
     }
 
-    _msgColorTheme(message) {
+    _msgColorTheme() {
```
seanshpark, line 495: caller is `this._msgColorTheme(message);` please check caller.

media/CircleGraph/index.js
```diff
@@ -135,7 +135,7 @@ host.BrowserHost = class {
                     this._msgSelection(message);
                     break;
                 case 'colorTheme':
-                    this._msgColorTheme(message);
+                    this._msgColorTheme();
```
seanshpark, line 138: I mean, do not remove `message`

seanshpark (approved):
LGTM thank you!

media/CircleGraph/index.js
```diff
@@ -492,7 +492,7 @@ host.BrowserHost = class {
         this._view.setSelection(message);
     }
 
-    _msgColorTheme(message) {
+    _msgColorTheme(_message) {
```
yunjayh, line 495: I could not fully understand why not to remove the argument. 😢 As the title of this pr, both `message` isn't used in the method. 
@seanshpark @dayo09 Was there any background to remain the argument as variable but adding underscore in front of the name?

seanshpark: because I wanted to preserve initial design of using message argument. currently I don't have time to take another look with other codes it would be appropriate to remove or not.

adding underscore will make ignore the tool used for unused variable.

dayo09: @yunjayh Well I think so too. :sweat_smile:  I just put the author's mind as the top priority. I introduced the underscore-ignoring rules to make a room for it.

yunjayh (approved):
LGTM Thank you!!!

Issue #1269, opened 2022-09-14
[Circle Editor] Plans for new Circle Editor feature


yoojincha:
# [Circle Editor] Plans for new Circle Editor feature

> **WHAT**
> 

We are planning to implement an editor feature to the currently embedded netron circle viewer.
<br>
> **WHY**
> 

There were demands for modifying data in circle file. However, it is difficult to manipulate data in circle file, which is in binary format. Therefore, we plan to increase convenience by enabling modification on the current circle viewer in GUI form.
<br>
> **HOW**
> 

We plan to add an edit button to the sidebar that appears when Circle graph is clicked. Our goal is to let users modify the data in Attribute/Input/Output sections by turning them into input areas when clicking the edit button.

Current Circle Viewer is set to be Read-Only, and does not allow modification. We plan to expand the existing CustomViewer, Provider, CustomDocument codes to enable data CRUD. Once the edit button is clicked on our new GUI, user input data will be transferred to our CustomEditor(updated version of the current CircleViewer) to apply the changes to the original Circle file.

Below are two designs we are considering for the edit button:

1) Pencil icon
![pencil icon](https://user-images.githubusercontent.com/55440076/190074967-7829b1f2-156c-485d-a220-d642d479e844.png)

2) VS Code style button
![bluebutton](https://user-images.githubusercontent.com/55440076/190075014-d3de8cf8-c94b-4d1f-a0af-ef3c88868f7e.png)

llFreetimell:
### Test Model File

If you follow below steps, you can can inception_v3 circle model file!

- Download https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz and extract it. Then you can find `inception_v3.tflite`.
- Download https://github.com/Samsung/ONE/releases/download/1.21.0/nncc-1.21.0.tar.gz and extract it. Then you can find `bin/tflite2circle`.
- Execute `./tflite2circle inception_v3.tflite inception_v3.circle`
- Open `inception_v3.circle` with the editor!

YuyeonHan:
## [Circle Editor] Add files for making Circle model object #1305

We have completed issue https://github.com/Samsung/ONE-vscode/issues/1269 and now ready to send pull request per each function.

These are brief explanations of the files we would like to merge:

1. src/CircleEditor/circle_schema_generated.ts
- ts file auto generated with flatbuffers according to circle schema.
[...]

PR #1317, opened 2022-10-01
[CircleEditor] Preprocessing for Editing Circle Files

Seonhyuk:
This PR introduces preprocessing for editing circle file. It enables saving the data from open window view, adding data necessary for data modification, and adding vscode cmds and types of custom operator attributes.

ONE-vscode-DCO-1.0-Signed-off-by: Seonhyuk Ahn <gomduridkstjsgur@gmail.com>

---

Related issue : #1269
Draft : #1298

### DIFF
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -149,7 +149,7 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
-                // TODO add message listener 'setCustomOpAttrT'
+                    // TODO add message listener 'setCustomOpAttrT'
             }
         });
 
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -46,8 +46,7 @@
 
 var host = {};
 
-var vscode = acquireVsCodeApi();
-const builtinOperatorType = builtinOperatorType || {};
+const vscode = acquireVsCodeApi();
 
 const viewMode = {
     viewer: 0,
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -149,9 +149,7 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
-                case 'getCustomOpAttrT':
-                    this._msgGetCustomOpAttrT(message);
-                    break;
+                // TODO add message listener 'setCustomOpAttrT'
             }
         });
 
@@ -543,19 +541,7 @@ host.BrowserHost = class {
         vscode.postMessage({command: 'loadmodel', offset: '0'});
     }
 
-    _msgGetCustomOpAttrT(message) {
-        const data = message.data;
-        const graphs = this._view._model._graphs;
-        const types = data._type;
-        const node = graphs[data._subgraphIdx]._nodes[data._nodeIdx];
-        for (const key of Object.keys(types)) {
-            for (let i in node.attributes) {
-                if (node.attributes[i].name === key) {
-                    node.attributes[i]._type = types[key];
-                }
-            }
-        }
-    }
+    // TODO add function msgSetCustomOpAttrT
 };
 
 host.Dropdown = class {
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -149,8 +149,8 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
-                case 'customType':
-                    this._msgGetType(message);
+                case 'getCustomOpAttrT':
+                    this._msgGetCustomOpAttrT(message);
                     break;
             }
         });
@@ -543,9 +543,9 @@ host.BrowserHost = class {
         vscode.postMessage({command: 'loadmodel', offset: '0'});
     }
 
-    _msgGetType(message) {
+    _msgGetCustomOpAttrT(message) {
         const data = message.data;
-        const graphs = data._view._model._graphs;
+        const graphs = this._view._model._graphs;
         const types = data._type;
         const node = graphs[data._subgraphIdx]._nodes[data._nodeIdx];
         for (const key of Object.keys(types)) {
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -149,9 +149,6 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
-                case 'edit':
-                    this._msgLoadModel(message);
-                    break;
                 case 'customType':
                     this._msgGetType(message);
                     break;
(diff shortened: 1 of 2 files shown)


media/CircleEditor/index.js
```diff
@@ -146,6 +150,12 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
+                case 'edit':
+                    this._msgLoadModel(message);
+                    break;
+                case 'customType':
+                    this._msgGetType(message);
```
llFreetimell, line 157: It's hard to think about the meaning of this function :(
What is this function for?

media/CircleEditor/view.js
```diff
@@ -821,6 +825,24 @@ view.View = class {
                     if (this._page !== 'default') {
                         this.show('default');
                     }
+                    for (let idx = 0; idx < model.graphs.length; idx++) {
+                        model._graphs[idx]['_subgraphIdx'] = idx;
+                        for (let jdx = 0; jdx < model.graphs[idx].nodes.length; jdx++) {
+                            model._graphs[idx]._nodes[jdx]['_subgraphIdx'] = idx;
+                            if (builtinOperatorType
+                                    [model._graphs[idx]._nodes[jdx]._type.name.toUpperCase()] ===
```
llFreetimell, line 833: As far as I know, `CUSTOM` is defined in `builtinOperatorType` as 32.
Isn't it error when `builtinOperatorType` is undefined?

media/CircleEditor/view.js
```diff
@@ -168,6 +168,10 @@ view.View = class {
             }
         }
         this._page = page;
+
+        if (typeof (this._host._viewingNode) === 'number') {
```
llFreetimell, line 172: How about following?
At first, I misunderstood that `_viewingNode` can be another type than `number` :)

```suggestion
        if (this._host._viewingNode !== null) {
```

media/CircleEditor/view.js
llFreetimell: And this is just optional, but it would be better to introduce a function like `builtinOperatorTypeAsNumber(string)` :)

media/CircleEditor/index.js
```diff
@@ -46,7 +46,8 @@
 
 var host = {};
 
-const vscode = acquireVsCodeApi();
+var vscode = acquireVsCodeApi();
```
llFreetimell, line 49:
```suggestion
let vscode = acquireVsCodeApi();
```

media/CircleEditor/index.js
Seonhyuk: For attributes of the custom operator, the type is not recognized by netron. So it's a function that takes the type from the model and fills in the attributes type of the custom operator.

media/CircleEditor/view.js
Seonhyuk: Thank you! I think it's a good idea!

media/CircleEditor/index.js
Seonhyuk: It doesn't have to be declared again. I will use `const` as it is. Thank you!

media/CircleEditor/view.js
Seonhyuk: I was found a custom operator classification method in the netron model. I will use the if statement as like `model._graphs[idx]._nodes[jdx]._type?.category === 'custom'`. Thank you!!!

media/CircleEditor/index.js
llFreetimell: If `acquireVsCodeApi()` is called twice, it would cause an error.
The solution is creating a file like `media/CfgEditor/vscodeapi.js` and importing it :)

media/CircleEditor/index.js
llFreetimell: As there are so many kinds of "type", it is really confused :(
For example,
- `customType` : Is there any type like this? Which type in custom operator, attributes' or inputs' or outputs'?
- `msgGetType` : Which type can be gotten? All operators or only some of operators?

Could you clarify the names?

media/CircleEditor/index.js
```diff
@@ -146,6 +149,12 @@ host.BrowserHost = class {
                 case 'reload':
                     this._msgReload(message);
                     break;
+                case 'edit':
+                    this._msgLoadModel(message);
```
llFreetimell, line 153: Only seeing the `command`, `edit` seems not related with loading model.
Furthermore, this case is exactly same with `loadmodel`.
What is this case for?

media/CircleEditor/index.js
```diff
@@ -537,10 +546,19 @@ host.BrowserHost = class {
...
     }
 
-    /**
-     * TODO Implement _msgGetType(message)
-     * Get custom operator attributes type
-     */
+    _msgGetType(message) {
+        const data = message.data;
+        const graphs = data._view._model._graphs;
+        const values = data._type;
```
llFreetimell, line 552: For consistency, following seems better :)

```suggestion
        const types = data._type;
```

media/CircleEditor/index.js
Seonhyuk: I separated the `edit` command considering the possibility of partial modification in the future, but I think it can be combined into `loadmodel` :)

media/CircleEditor/index.js
Seonhyuk: Thank you!

media/CircleEditor/index.js
Seonhyuk: There could be some confusion... `msgGetCustomOperatorAttributesType` I think it will be a little long to express them all clearly, do you have any good ideas?

llFreetimell: Hmm... how about `msgGetCustomOpAttrT`?
It's little bit longer too but shorter than full name...

Seonhyuk: I'd like that. I'll apply it! Thank you!

media/CircleEditor/view.js
```diff
@@ -821,6 +825,22 @@ view.View = class {
...
                         this.show('default');
                     }
+                    for (let idx = 0; idx < model.graphs.length; idx++) {
+                        model._graphs[idx]['_subgraphIdx'] = idx;
+                        for (let jdx = 0; jdx < model.graphs[idx].nodes.length; jdx++) {
+                            model._graphs[idx]._nodes[jdx]['_subgraphIdx'] = idx;
+                            if (model._graphs[idx]._nodes[jdx]._type?.category === 'custom') {
+                                model._graphs[idx]._nodes[jdx]._isCustom = true;
+                                vscode.postMessage({
+                                    command: 'getCustomOpAttrT',
```
llFreetimell, line 835: I am little bit confused...
It seems that there are two `getCustomOpAttrT`, which are in `index.js` and `CircleEditorProvider.ts`.
Then the two message listeners are really for **getting** custom operator attributes type? not **setting**?
Furthermore, there is no `CircleEditorProvider.ts` to handle this message...

Seonhyuk: I'll change `get-` to `set-` in `index.js`. 
Would it be better to upload the related parts of the message after `CircleEditorProvider.ts` is uploaded? 
I think that sending and receiving messages is always complementary, so what should we do with the commands that already exist?

llFreetimell: > Would it be better to upload the related parts of the message after CircleEditorProvider.ts is uploaded?

Leaving it as TODO and then implementing after message listeners in `CircleEditorProvider.ts` is introduced would be better :)

> so what should we do with the commands that already exist?

What does this mean...? T.T

Seonhyuk: I left the related part as TODO..

llFreetimell (approved):
LGTM :)

YuyeonHan (approved):
LGTM :)

PR #1319, opened 2022-10-02
DRAFT: [Metadata Manager] Draft of Metadata

chshin59:
This commit introduces metadata.ts, which contains functions that manipulate the metadata of file system.

ONE-vscode-DCO-1.0-Signed-off-by: Changho Shin <chshin59@gmail.com>

### DIFF
--- a/src/MetadataManager/Metadata.ts
+++ b/src/MetadataManager/Metadata.ts
@@ -20,7 +20,6 @@ import * as vscode from 'vscode';
 import {ToolchainInfo} from '../Backend/Toolchain';
 import {obtainWorkspaceRoot} from '../Utils/Helpers';
 
-import {} from './PathToHash';
 import {getStats, isValidFile} from './Utils';
 
 type BuildInfoKeys = 'onecc'|'toolchain'|'cfg';
@@ -31,7 +30,7 @@ interface BuildInfoObj {
 
 export class BuildInfo {
   private static _map = new Map<string, BuildInfoObj>();
-  
+
   public static get(metadata: any, uri: vscode.Uri) {
     const path = vscode.workspace.asRelativePath(uri);
     const info = BuildInfo._map.get(path);
@@ -80,8 +79,7 @@ export class Metadata {
     const jsonUri = vscode.Uri.joinPath(
         vscode.Uri.file(workspaceroot),
         `.meta/hash_objects/${hash.substring(0, 2)}/${hash.substring(2)}.json`);
-    await vscode.workspace.fs.writeFile(
-      jsonUri, Buffer.from(JSON.stringify(obj, null, 4), 'utf8'));
+    await vscode.workspace.fs.writeFile(jsonUri, Buffer.from(JSON.stringify(obj, null, 4), 'utf8'));
   }
 
   public static async getEntry(uri: vscode.Uri, hash: string) {
@@ -94,25 +92,24 @@ export class Metadata {
     return undefined;
   }
 
-  public static async setEntry(uri: vscode.Uri, hash: string, obj: any) {
+  public static async setEntry(uri: vscode.Uri, hash: string, entry: any) {
     const workspaceroot = obtainWorkspaceRoot();
     const metadataUri = vscode.Uri.joinPath(
-      vscode.Uri.file(workspaceroot),
-      `.meta/hash_objects/${hash.substring(0, 2)}/${hash.substring(2)}.json`);
+        vscode.Uri.file(workspaceroot),
+        `.meta/hash_objects/${hash.substring(0, 2)}/${hash.substring(2)}.json`);
 
     const relPath = vscode.workspace.asRelativePath(uri);
     let metadata = await Metadata.getObj(hash);
     if (!metadata[relPath]) {
       metadata[relPath] = {};
     }
-    for (let key in obj) {
-      metadata[relPath][key] = obj[key];
+    for (let key in entry) {
+      metadata[relPath][key] = entry[key];
     }
-    await vscode.workspace.fs.writeFile(metadataUri, Buffer.from(JSON.stringify(metadata, null,
-    4), 'utf8'));
+    await vscode.workspace.fs.writeFile(
+        metadataUri, Buffer.from(JSON.stringify(metadata, null, 4), 'utf8'));
   }
 
-  // TODO: 덮어쓰기 추가
   public static async createDefault(uri: vscode.Uri, hash: string) {
     if (!vscode.workspace.workspaceFolders) {
       return;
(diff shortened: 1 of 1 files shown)


src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,173 @@
...
+    let info = BuildInfo._map.get(relPath);
+    if (info === undefined) {
+      info = {onecc: '', toolchain: undefined, cfg: undefined};
+      BuildInfo._map.set(relPath, info);
+    }
+    info[key] = value;
+  }
+}
+
+export class Metadata {
```
dayo09, line 58: Thanks for introducing `Metadata` class :smile:

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+    // TODO: Error Handling
+    return undefined;
+  }
+
+  public static async setObj(hash: string, obj: object) {
+    const workspaceroot = obtainWorkspaceRoot();
+    const jsonUri = vscode.Uri.joinPath(
+        vscode.Uri.file(workspaceroot),
+        `.meta/hash_objects/${hash.substring(0, 2)}/${hash.substring(2)}.json`);
```
dayo09, line 96: This may require a big change, but could you rename the directory..? `.meta/` -> `.one-vscode/info/`

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,189 @@
...
+    metaObj[relPath]['name'] = filename;
+    metaObj[relPath]['file-extension'] = filename.split('.').pop();
+    metaObj[relPath]['create-time'] = stats.birthtime;
+    metaObj[relPath]['modified-time'] = stats.mtime;
+    metaObj[relPath]['is-deleted'] = false;
+    await Metadata.setObj(hash, metaObj);
+  }
+
+  // NOTE When the deleted file is recovered, metadata is also recovered.
+  //    For that situation, metadata is not deleted but deactivated.
```
dayo09, line 156:
```suggestion
  // For that situation, metadata is not deleted but deactivated.
```

chshin59:
We found a problem that takes too much time to load if there are large model and circle files in user workspace.

src/MetadataManager/Metadata.ts
chshin59: We will apply the comments. I think replacing `.meta/` -> `.one-vscode/info/` will be not that hard.

dayo09:
Could you make this draft into a PR if it's ready?

src/MetadataManager/Metadata.ts
dayo09: Thanks :-D

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    // step 3. deactivate (set 'is_deleted') that path.
+    metaObj[relPath]['is-deleted'] = true;
+    await Metadata.setObj(hash, metaObj);
+  }
+
+  public static async delete(uri: vscode.Uri, hash: string) {
+    const metaObj = await Metadata.getObj(hash);
+    const relPath = vscode.workspace.asRelativePath(uri);
+    delete metaObj[relPath];
```
dayo09, line 197: I wonder what happens when metaObj[relPath] not exists. If it throws, it's critical. Could you make a unit tests for that, maybe after this PR?

chshin59: When metaObj[relPath] does not exist, if you try to delete, nothing happens.
[Reference here](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Operators/delete#%EC%84%A4%EB%AA%85)
After this PR, I will add this case to the unit test just in case.

dayo09 (dismissed):
LGTM, Thanks :-D

gandi0330:
good!

gandi0330 (dismissed):

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  public static async save(metaEntry: any, uri: vscode.Uri) {
+    if (!uri.fsPath.endsWith('.circle')) {
+      return;
+    }
+
+    const stat = fs.statSync(uri.fsPath);
+    const operations: any = {};
+    // TODO deal with large files
+    if (stat.size > 10000) {
```
llFreetimell, line 80: AFAIK, `stat.size` is byte, not KB or MB.
If so, this logic is denying circle file over than 10000byte, which is 9.7KB.
I think 9.7KB is not that large... @.@

gandi0330:
This part seems to be a mistake code. 

We have never had a problem because of the size of the circle, and when we read the 96MB model, we had a problem, which seems to have been misrepresented here. 

Of course, it is true that it takes a long time to read a model or circle file or its capacity if it is too large, but specifying its size limit seems to require a lot of testing.

 So for now, it would be better to remove the restrictions.

src/MetadataManager/Metadata.ts
gandi0330: It seems that we took a long time in the circle operator extraction process using flatbuffer

Of course, it is true that it takes a long time to read a model or circle file or its capacity if it is too large, but specifying its size limit seems to require a lot of testing. 

So for now, it would be better to remove the restrictions.

dayo09: IMO, if it slows down the extension, it's critical, as it gathers up the metadata on extension's startup, the user must wait long. Instead, we decided to support on-click showing of the circle-op information. 
I think it's better to set conservative number for now, but if there is any number you recommends, let us know. @llFreetimell

chshin59: I want to let you know that the function is executed not at the startup but when you run cfg.

And, I think there are 2 options.

> So for now, it would be better to remove the restrictions.

1. Remove restrictions

> I think it's better to set conservative number for now

2. Set conservative number for now

Which one should I choose...?

llFreetimell: > but if there is any number you recommends, could you let us know? @llFreetimell

Unfortunately, I have no idea for it.
Nevertheless, the reason why I wrote about this was... if 9.7KB is large, almost all circle model files cannot be saved because 9.7KB is really small model.
(FYI, `res/samples/circle/while_000.circle` is 1.5KB even it has less than 10 operators)

Since this restriction is only applied for metadata manger, I am not against with going as it is :)
Please let me know @dayo09's decision, I will follow you!

dayo09: > I want to let you know that the function is executed not at the startup but when you run cfg.

Oh.. Sure. I remember the line. Then, how long does it increase the onecc-running time..? (like, doubled? 20%?)
 
When I skim through the internal model list, 10MB(to support over 60%) or 1MB(more conservatively) looks fine for the restriction.

Plus, please defined the value as a const variable somewhere and add a comment there.

chshin59: The number 10000 was a meaningless value set without testing.
In fact, when I ran a function on a 92MB file, it took me about 30 seconds...
I'm thinking of removing the restriction for consistency with ONEIT team, so if you allow me, I'll fix it and push it.
@dayo09 @llFreetimell

llFreetimell: > I'm thinking of removing the restriction for consistency with ONEIT team

Since two teams are handling different features, you don't need to think about it :)
I left a comment with maintainer's point of view, not OneIT team!

dayo09: > In fact, when I ran a function on a 92MB file, it took me about 30 seconds...

If so, let's take 1MB restriction. I think affecting the running time is not good. If it's calculated when we click 'View Metadata', it doesn't matter how long time it would take. (<-And you can write down this as TODO.)

gandi0330 (approved):

dayo09 (approved):
LGTM, Thanks :-D

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,211 @@
...
+  public static async save(metaEntry: any, uri: vscode.Uri) {
+    if (!uri.fsPath.endsWith('.circle')) {
+      return;
+    }
+
+    const stat = fs.statSync(uri.fsPath);
+    const operations: any = {};
+
+    // Note maximum size is limited to 1MB at present in order not to affect the running time.
+    if (stat.size > Operator._maxSize) {
```
dayo09, line 84: 👍

chshin59: 😂

src/MetadataManager/Metadata.ts
```diff
@@ -0,0 +1,211 @@
...
+      metaEntry['onecc-version'] = info['onecc'];
+      metaEntry['toolchain-version'] = info['toolchain'] ?.version ?.str();
+      metaEntry['cfg-settings'] = info['cfg'];
+    }
+    BuildInfo._map.delete(path);
+  }
+}
+
+export class Operator {
+  private static _maxSize = 1000000;
```
llFreetimell, line 59: Sorry for bothering but 1MB = 1024 * 1024 T.T

llFreetimell (approved):
LGTM // 1000000 is not 1MB but since there is no comment for it, don't care :)

PR #1326, opened 2022-10-04
[CircleGraph] Fix to find file system path

llFreetimell:
Currently, `uri.path` is used for finding path.
However, it causes an error at Windows native environment. This commit fixes the error.

### DIFF
--- a/src/CircleGraph/CircleViewer.ts
+++ b/src/CircleGraph/CircleViewer.ts
@@ -75,7 +75,7 @@ export class CircleViewerDocument implements vscode.CustomDocument {
 
   public openView(panel: vscode.WebviewPanel, extensionUri: vscode.Uri) {
     let view = new CircleViewer(panel, extensionUri);
-    view.initGraphCtrl(this.uri.path, undefined);
+    view.initGraphCtrl(this.uri.fsPath, undefined);
     view.loadContent();
     this._circleViewer.push(view);
 


seanshpark:
> Windows native environment

@llFreetimell , did we had a plan to support for Windows?
Can you explain how you used in Windows?

- I don't have any objection with this change but want to know the situation
- I didn't had any consideration with Windows platform
- Do we have plans for MacOSX ?
- If we are to support Windows officialy I think we need to post an issue and gather opinion.

seanshpark:
@lemmaa , PTAL. I maybe not know anything with Windows. Can you please share for this topic if any?

lemmaa:
> @lemmaa , PTAL. I maybe not knowing anything with Windows. Can you please share for this topic if any?

Well, there are no special plans, but considering the characteristics of vscode, I don't think there is any reason not to support Windows or MacOS. Rather, as a vscode user, isn't that what everyone expect by default?

llFreetimell:
@seanshpark 

I never expected that this PR is such a critical issue... @.@

AFAIK, there is no official plan to support Windows native.
However, custom editors were unofficially working on Windows native except CircleViewer.
After some investigation, the cause of the difference was quite simple, `uri.path`.
So I thought that it would be better to fix it because we can see circle file in Windows Native too only with simple fix.

However, if this PR requires further discussion and opinions, I will close this PR :)

seanshpark:
> However, if this PR requires further discussion and opinions, I will close this PR :)

seanshpark (approved):
LGTM

seanshpark:
Ah, one more thing... why I didn't think of Windows
- ONE-vscode needs ONE
- ONE is only supported for Linux
- To use Windows/MacOSX, one needs to use Docker, which is not free to use for our case
- Maybe need to aquire some license?

PR #1333, opened 2022-10-05
[Toolchain] Introduce local toolchain

dayo09:
This commit introduces local toolchain.
When `onecc` command is found within the system, this 'local toolchain' will use it.

Co-authored-by: Semi Jeong <smjeong53@gmail.com>

---

Originated from https://github.com/Samsung/ONE-vscode/pull/1312

---

## GIF

1005-local-toolchain

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@types/mocha": "^9.1.0",
         "@types/node": "^12.11.7",
         "@types/vscode": "^1.46.0",
+        "@types/which": "^2.0.1",
         "@typescript-eslint/eslint-plugin": "^4.9.0",
         "@typescript-eslint/parser": "^4.9.0",
         "@vscode/test-electron": "^2.1.3",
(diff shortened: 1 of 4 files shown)


jyoungyun:
(Q) Why do we need this toolchain? What is the difference from the `ONE` backend toolchain?

dayo09:
@jyoungyun It will be useful if (1) User don't have docker environment (2) User want to simply use the existing local toolchain.

And from project's view, I tried to install the docker toolchain but it has put some errors while 'prerequisites' settings, I cannot get the log to report because it's a spawned process though

jyoungyun:
If there is a problem with One backend itself, it seems right to correct it. The design to add a `One(installed, local) backend` has not been discussed so far. ONE-Vscode is designed to allow users to install and remove Toolchains using this tool. It is correct that existing installed tools should also be shown through One backend in consideration of the installed Toolchain version. This addition is related to the large directioanality of ONE-vscode. Before merging this PR, we need to get together with all ONE-vscode people and discuss it.

dayo09:
@jyoungyun I remember that we have discussed about this local toolchain before, like two weeks ago. You don't remember? I am not sure if there is real reason to against this feature. I agree with the quick discussion about this then.

jyoungyun:
At the beginning of the SSAFY project, I agree to implement it quickly first because it took time for the One backend implmentation to be completed. However, since One backend is already implemented now, so it would be better to use it.

dayo09:
@jyoungyun In that case I wonder if there is an exact reason that you are against this change. You agreed that time and the only difference is that ONE-docker toolchain is introduced. The effect and advantages of those two toolchains are fairly different. Anyways it doesn't work at my environment, so to speak.

dayo09:
### Reporting one-docker toolchain bug

[peek3WQMT1.webm](https://user-images.githubusercontent.com/17171963/194024059-d89809a6-708a-4261-b740-989cb4e32d3c.webm)

lemmaa:
I do not have an exact understanding about operation of the toolchain that ONE-vscode has. However, it is currently understood that the use of the toolchain is determined by the backend extension of ONE-vscode, not ONE-vscode itself. If so, I agree with @jyoungyun's opinion.

IIUC, what the feature @dayo09 wants to introduce is that allow ONE-vscode to install and use its own toolchain from the frontend point of view? If so, it seems to be the priority to come up with a conflict-free method for the current backend-centered toolchain management policy.

Anyway, doesn't @dayo00's  need be solved by fixing the current problem of docker-based one-toolchain? If so, why don't try this first?

dayo09:
I moved the discussion to #1334, this PR will be pending until it's resolved.

Issue #1335, opened 2022-10-05
Reporting failure of one-docker toolchain: Running prerequisites failed


dayo09:
### Reporting one-docker toolchain bug

[peek3WQMT1.webm](https://user-images.githubusercontent.com/17171963/194024059-d89809a6-708a-4261-b740-989cb4e32d3c.webm)

_Originally posted by @dayo09 in https://github.com/Samsung/ONE-vscode/issues/1333#issuecomment-1268162214_

PR #1336, opened 2022-10-05
[OneToolchain] Returns the output value as it is

jyoungyun:
This commit returns the output value as it is.
This allows the log to be displayed and the output to be handled.

~~fix this bug: https://github.com/Samsung/ONE-vscode/pull/1333#issuecomment-1268162214, #1335~~

### DIFF
--- a/script/prerequisitesForGetToolchains.sh
+++ b/script/prerequisitesForGetToolchains.sh
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 
 # This shell script add PPA to get onecc-docker package for ONEToolchain.
 
 PPA_NAME="ppa:one-compiler/onecc-docker"
 
-add-apt-repository -y ${PPA_NAME} > /dev/null 2>&1 &&
-        apt-get update > /dev/null 2>&1
+add-apt-repository -y ${PPA_NAME}
+apt-get update -yqq


Psalmist-KIM (approved):

lettersto:
Is this to see logs when prerequisites fails?

jyoungyun:
> Is this to see logs when prerequisites fails?

script/prerequisitesForGetToolchains.sh
```diff
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/bin/bash
```
lettersto, line 1: I see the change in shebang. So, do we have to change `sh` to `bash` in `prerequistesForGetToolchains` too?

jyoungyun: Our project usually uses bash shell. So I changed it too. :)

jyoungyun:
The #1335 error does not occur in my test environment. However, printing the log will help you determine what happened. So I want to merge this PR.

script/prerequisitesForGetToolchains.sh
lettersto: I see :) Thank you!

lettersto (approved):

llFreetimell (approved):

script/prerequisitesForGetToolchains.sh
```diff
@@ -1,8 +1,8 @@
...
 
 # This shell script add PPA to get onecc-docker package for ONEToolchain.
 
 PPA_NAME="ppa:one-compiler/onecc-docker"
 
-add-apt-repository -y ${PPA_NAME} > /dev/null 2>&1 &&
-        apt-get update > /dev/null 2>&1
+add-apt-repository -y ${PPA_NAME}
+apt-get update -yqq
```
llFreetimell, line 8: AFAIK, `-yqq` is not related with printing output value, isn't it?

jyoungyun: My intention is to print the `add-apt-repository` logs. I thought the `apt-get update` log was too long and unnecessary.

llFreetimell: OK, I see it!

jyoungyun: Ah, `-yqq` means that no output except for errors. It does not related to output value. :)

PR #1341, opened 2022-10-05
[CircleEditor] Add Attributes Edit function

rudalsd:
This commit adds attributes edit function.
 -BuiltInOperator attribute value edit function.
 -CustomOperator attribute name, value and edit function.
 -CustomOperator attribute type edit function will be implemented.

Postmessage will be modified when the DS code is merged.

ONE-vscode-DCO-1.0-Signed-off-by: Gyeongmin Kwon rudalsd@naver.com

---

Related issue : #1269
Draft : #1298

PR #1343, opened 2022-10-05
[CircleEditor] Update files for attribute type

rudalsd:
This commit updates files for attribute type.
type.js and index.html, the main files of this pull request.

ONE-vscode-DCO-1.0-Signed-off-by: Gyeongmin Kwon <rudalsd@naver.com.


llFreetimell:
Could you leave the reason why they are updated in commit message?

media/CircleEditor/type.js
```diff
@@ -15,7 +15,7 @@
  */
 
 /* eslint-disable */
-let typeName = {
+var typeName = {
```
llFreetimell, line 18: I think `typeName` should be modified too because it is not matched well T.T
These are not typename but map whose key is `OptionName` and value is `OptionValue`.

Could you change this name to `optionValues`?
```suggestion
var optionValues = {
```

media/CircleEditor/type.js
```diff
@@ -28,14 +28,12 @@ let typeName = {
...
-    'UINT64', 'UINT32'
-];
+var tensorType =
+    ['FLOAT32', 'FLOAT16', 'INT32', 'UINT8', 'INT64', 'BOOLEAN', 'INT16', 'INT8', 'FLOAT64'];
 
-let customType = ['int', 'boolean', 'string'];
+var customType = ['int', 'boolean', 'string'];
 
-let builtinOperatorType = {
+var builtinOperatorType = {
```
llFreetimell, line 36: This may be too.... T.T

```suggestion
var builtinOperatorTypeEnum = {
```

rudalsd: Since the function is not implemented in this PR, I will modify it to another PR  #1341 .

llFreetimell:
As I wrote in https://github.com/Samsung/ONE-vscode/pull/1343#issuecomment-1269245393, 
I wonder why `let` cannot be used :)

rudalsd:
The logic so far is a method of re-declaring a variable at the top and using it, but if we don't re-declare it, we can use const or let.

llFreetimell:
OK, then just leave the reason at the commit message and make commits as one :)

llFreetimell:
Please make commits as one...!

llFreetimell (approved):
LGTM

YuyeonHan (approved):
LGTM

PR #1344, opened 2022-10-06
[CircleEditor] Add Json editor feature

tanficial:
This commit adds Json editor feature.
 - loadJson
 - editJsonModel
 - toJson function of BigInt type

ONE-vscode-DCO-1.0-Signed-off-by: Taeyoon Jeong <tanficial9574@gmail.com>

---
Related issue : #1269

### DIFF
--- a/src/CircleEditor/CircleEditorDocument.ts
+++ b/src/CircleEditor/CircleEditorDocument.ts
@@ -17,9 +17,19 @@
 import * as flatbuffers from 'flatbuffers';
 import * as vscode from 'vscode';
 
+import {Balloon} from '../Utils/Balloon';
 import {Disposable} from '../Utils/external/Dispose';
 
 import * as Circle from './circle_schema_generated';
+import * as Types from './CircleType';
+
+/**
+ * Make BigInt data to string type.
+ * This is called by JSON.stringify function.
+ */
+(BigInt.prototype as any).toJSON = function() {
+  return this.toString();
+};
 
 /**
  * Custom Editor Document necessary for vscode extension API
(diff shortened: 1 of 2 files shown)


src/CircleEditor/CircleEditorDocument.ts
```diff
@@ -191,4 +201,176 @@ export class CircleEditorDocument extends Disposable implements vscode.CustomDoc
       }
     };
   }
+
+  /**
+   * edit _model state when user modified model through json editor
+   */
+  editJsonModel(message: any) {
```
llFreetimell, line 208: Since `message.data` is only usage for `message`, how about fixing it?

```suggestion
  editJsonModel(newModelString: string) {
```

src/CircleEditor/CircleEditorDocument.ts
```diff
@@ -191,4 +201,176 @@ export class CircleEditorDocument extends Disposable implements vscode.CustomDoc
...
+      let newModel = JSON.parse(message.data);
+
+      // Copying model from message to _model starts from here.
+      // This is required as parsed json object do not have any prototypes necessary for
+      // Circle.ModelT object.
+
+      // version
+      this._model.version = newModel.version;
+
+      // operatoreCodes
```
llFreetimell, line 220:
```suggestion
      // operatorCodes
```

src/CircleEditor/CircleEditorDocument.ts
```diff
@@ -191,4 +201,176 @@ export class CircleEditorDocument extends Disposable implements vscode.CustomDoc
...
+
+        // operators
+        data.operators = data.operators.map((operator: Circle.OperatorT) => {
+          // case1 : custom operator
+          if (this._model.operatorCodes[operator.opcodeIndex].deprecatedBuiltinCode === 32) {
+            if (operator.builtinOptionsType || operator.builtinOptions) {
+              throw new Error;
+            }
+            // case2 : builtin operator
+          } else {
```
llFreetimell, line 294: Please do not make a distance between if and else

```suggestion
          } else {
          // case2 : builtin operator
```

src/CircleEditor/CircleEditorDocument.ts
YuyeonHan: We will fix this with change argument in function onMessage in CircleEditorProvider.ts
It will be from document.editJsonModel(message) to document.editJsonModel(message.data)

src/CircleEditor/CircleEditorDocument.ts
```diff
@@ -191,4 +201,176 @@ export class CircleEditorDocument extends Disposable implements vscode.CustomDoc
...
+
+  /**
+   * Send model of JSON format to webview.
+   * TODO: Implement feature to load json with multiple times as this does not work when size of
+   * JSON is too large.
+   */
+  loadJson() {
+    let jsonModel = JSON.stringify(this._model, null, 2);
+    const numberArrayStrings = jsonModel.match(/\[[0-9,\s]*\]/gi);
+    if (numberArrayStrings) {
```
llFreetimell, line 367: I think `if` is not needed because when `numberAarayStrings` is empty, `forEach` is not iterated.

src/CircleEditor/CircleEditorDocument.ts
```diff
@@ -17,9 +17,19 @@
...
 import {Disposable} from '../Utils/external/Dispose';
 
 import * as Circle from './circle_schema_generated';
+import * as Types from './CircleType';
+
+/**
+ * Make BigInt data to string type.
+ * This is called by JSON.stringify function.
+ */
+(BigInt.prototype as any).toJSON = function() {
```
llFreetimell, line 30: Quite interesting... does this have an effect only on this file? or imported file too?

src/CircleEditor/CircleEditorDocument.ts
tanficial: match function returns null instead of empty array when there are no matched string with regex. So, I checked it is null before use forEach function.

llFreetimell (approved):
LGTM

YuyeonHan (approved):
LGTM :)

Issue #1352, opened 2022-10-14
[CircleGraphCtrl] Simplify CircleGraphEvent


seanshpark:
CircleGraphEvent provides explicit message handler,
but this forces other existing listeners to change when new message is introduced.

Lets' simplify this.

seanshpark:
from https://github.com/Samsung/ONE-vscode/issues/1350#issuecomment-1278615737

seanshpark:
done

PR #1351, opened 2022-10-14
DRAFT initial VisquvProvider

seanshpark:
on-going draft to provide VisquvProvider

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #1353, opened 2022-10-14
[CircleGraphCtrl] Relocate handlers

seanshpark:
This will relocate onDidReceiveMessage handlers.

### DIFF
--- a/src/CircleGraph/CircleGraphCtrl.ts
+++ b/src/CircleGraph/CircleGraphCtrl.ts
@@ -145,27 +145,31 @@ export class CircleGraphCtrl {
   private registerEventHandlers() {
     // Handle messages from the webview
     this._webview.onDidReceiveMessage(message => {
+      switch (message.command) {
+        case MessageDefs.alert:
+          Balloon.error(message.text);
+          return;
+        case MessageDefs.request:
+          this.handleRequest(message.url, message.encoding);
+          return;
+        case MessageDefs.loadmodel:
+          this.handleLoadModel(parseInt(message.offset));  // to number
+          return;
+        case MessageDefs.finishload:
+          this.handleFinishLoad();
+          return;
+        default:
+          break;
+      }
       this.handleReceiveMessage(message);
     }, null, this._ctrlDisposables);
   }
 
   protected handleReceiveMessage(message: any) {
     switch (message.command) {
-      case MessageDefs.alert:
-        Balloon.error(message.text);
-        return;
-      case MessageDefs.request:
-        this.handleRequest(message.url, message.encoding);
-        return;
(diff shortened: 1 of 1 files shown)


src/CircleGraph/CircleGraphCtrl.ts
```diff
@@ -145,27 +145,29 @@ export class CircleGraphCtrl {
...
+          return;
+        case MessageDefs.request:
+          this.handleRequest(message.url, message.encoding);
+          return;
+        case MessageDefs.loadmodel:
+          this.handleLoadModel(parseInt(message.offset));  // to number
+          return;
+        case MessageDefs.finishload:
+          this.handleFinishLoad();
+          return;
```
llFreetimell, line 160: Although `handleReceiveMessage` has no `default` handling, maybe it would be better to have? :)

```suggestion
          return;
        default:
          Balloon.error('No command found');
```

---

And this is just for curious, why only some of commands are relocated?

llFreetimell, line 160: At first, I was to recommend to add `default` but it prevents going to `handleReceiveMessage` function.
So I wonder why only some of commands are relocated, not all of them :)

seanshpark: Thanks for the comment :)
Relocated message have no or little relation with event handlers.
My direct target is to simplify `interface CircleGraphEvent` so that this won't depend on individual message types changes. The whole change + others exist in #1351 but maybe hard to read -_-;

seanshpark:
forgot to add a link to the issue:  #1352

src/CircleGraph/CircleGraphCtrl.ts
llFreetimell: I see it :)
Then could you add following comments? Just not to forget..!

```suggestion
          return;
        // default:
        //   TODO Handling undefined messages
```

seanshpark: As below `this.handleReceiveMessage(message);` is called and it will stay, there will be no `default` handling.

seanshpark: I've added `default` case for fall through.

src/CircleGraph/CircleGraphCtrl.ts
```diff
@@ -145,27 +145,31 @@ export class CircleGraphCtrl {
...
+        default:
+          break;
```
seanshpark, line 162: although this change is relocation of existing changes, `default` is added to express default fallthrought of not handled message.

llFreetimell (approved):
LGTM

Issue #1350, opened 2022-10-14
visq viewer from ONE


seanshpark:
This issue is about to support ONE quantization error visualizing feature.

In detail;
- there will be _visq_ tool that will produce quantization error, qerror, for each node
- information is stored in JSON format
- we can utilize existing CircleGraph (powered by netron) for network visualize
- customization is needed for presentaion difference

seanshpark:
Issue of ONE; https://github.com/Samsung/ONE/issues/9694

seanshpark:
TODO
- [ ] add new Provide, maybe `VisqProvider`.
- [ ] ... (need to breakdown)

seanshpark:
while extending existing codes, current structure gives needless changes
it would be better to simplify this; `CircleGraphEvent`

seanshpark:
done with initial version

PR #1356, opened 2022-10-16
[media/CircleGraph] Introduce region for legend

seanshpark:
This will introduce region for legend to show legend of quantization error (visquv).


seanshpark:
for #1350, from draft #1351

this region is to show legend of quantization error range (visq)

image

mhs4670go (approved):
LGTM

llFreetimell (approved):

PR #1359, opened 2022-10-17
DRAFT: [OneExplorer] Support refactoring by renaming

dayo09:
This commit enables refactoring by renaming.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

----

For #768

## GIF
1017-refactor

PR #1360, opened 2022-10-17
[OneExplorer] Revive rename command

dayo09:
This commit revives rename command which is deprecated by #1243.

---

FOR REVIEWERS,

This is a baseline for 'Cfg Refactoring' feature, which replaces input_file names with the new one when the mode file name is changed.

For #1359

## GIF

NOTE That the behavior is the same as before, it's just a revival of the same old 'rename' command.

1017-revive

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -641,20 +641,29 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
     // TODO: prohibit special characters for security ('..', '*', etc)
     let warningMessage;
 
-    if (node.type === NodeType.baseModel) {
-      // TODO automatically change the corresponding files
-      warningMessage = `WARNING! ${
-          node.getChildren()
-              .map(node => `'${node.name}'`)
-              .toString()} will disappear from the view.`;
-    } else if (node.type === NodeType.product) {
-      // TODO automatically change the corresponding files
-      warningMessage = `WARNING! '${node.name}' may disappear from the view.`;
-    } else if (node.type === NodeType.config) {
-      // DO NOTHING
-      // Renaming config doesn't affect ONE Explorer view
-    } else if (node.type === NodeType.directory) {
-      assert.fail('Renaming is not allowed for directories.');
+    switch (node.type) {
+      case NodeType.baseModel:
+        // TODO automatically change the corresponding files
+        warningMessage = `WARNING! ${
+            node.getChildren()
+                .map(node => `'${node.name}'`)
+                .toString()} will disappear from the view.`;
+        break;
+      case NodeType.product:
+        // TODO automatically change the corresponding files
+        warningMessage = `WARNING! '${node.name}' may disappear from the view.`;
+        break;
+      case NodeType.config:
+        // DO NOTHING
+        // Renaming config doesn't affect ONE Explorer view
+        break;
+      case NodeType.directory:
+        assert.fail('Renaming is not allowed for directories.');
+        break;
+      default: {
+        // ..
+        break;
+      }
     }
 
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -631,6 +632,69 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+              .toString()} will disappear from the view.`;
+    } else if (node.type === NodeType.product) {
+      // TODO automatically change the corresponding files
+      warningMessage = `WARNING! '${node.name}' may disappear from the view.`;
+    } else if (node.type === NodeType.config) {
+      // DO NOTHING
+      // Renaming config doesn't affect ONE Explorer view
+    } else if (node.type === NodeType.directory) {
+      assert.fail('Renaming is not allowed for directories.');
+    }
```
mhs4670go, line 658: (optional) It would be better to use switch statement for readability and for removing multiple if statement.
```suggestion
  switch(node.type) {
    case NodeType.baseModel: {
      // TODO automatically change the corresponding files
      warningMessage = `WARNING! ${
          node.getChildren()
              .map(node => `'${node.name}'`)
              .toString()} will disappear from the view.`;
    }
    case NodeType.product: {
      // TODO automatically change the corresponding files
      warningMessage = `WARNING! '${node.name}' may disappear from the view.`;
    }
    case NodeType.config: {
      // DO NOTHING
      // Renaming config doesn't affect ONE Explorer view
    }
    case NodeType.directory: {
      assert.fail('Renaming is not allowed for directories.');
    }
    default: {
      // ..
    }
  }
```

mhs4670go:
Describing simple comment why deprecated module should be revived would be good for the record.

mhs4670go (dismissed):
LGTM

src/OneExplorer/OneExplorer.ts
dayo09: Cool, thanks :-D

src/OneExplorer/OneExplorer.ts
```diff
@@ -631,6 +632,77 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+    let warningMessage;
+
+    switch(node.type) {
+      case NodeType.baseModel: {
+        // TODO automatically change the corresponding files
+        warningMessage = `WARNING! ${
+            node.getChildren()
+                .map(node => `'${node.name}'`)
+                .toString()} will disappear from the view.`;
+      }
```
mhs4670go, line 651: Oh, `break` keywords are missing.

dayo09:
@mhs4670go I added the comment :-D

> FOR REVIEWERS,
> 
> This is a baseline for 'Cfg Refactoring' feature, which replaces input_file names with the new one when the mode file name is changed.

seanshpark (approved):

src/OneExplorer/OneExplorer.ts
```diff
@@ -631,6 +632,78 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+        break;
+      case NodeType.config:
+        // DO NOTHING
+        // Renaming config doesn't affect ONE Explorer view
+        break;
+      case NodeType.directory:
+        assert.fail('Renaming is not allowed for directories.');
+        break;
+      default: {
+        // ..
```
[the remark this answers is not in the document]
dayo09: Oh, I just accepted the suggestion, so it just came in. Let me remove it.

dayo09, line 664:
```suggestion
        assert.fail('Cannot reach here');
```

mhs4670go (approved):
LGTM

Issue #768, opened 2022-06-09
Edit cfg automatically when the related model file name is changed in the filesystem


dayo09:

dayo09:
As this job is considered to be lowered priority, I remove this job from the milestone.

dayo09:
@lemmaa I made this 'refactoring cfg' feature out as can be seen in GIF, I wonder if you have any better idea about the UI :-D

> This commit enables refactoring by renaming, so that the model file renaming can be automatically applied to related config files. It renames all the changed model file names in children config files.
> 
> ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee [dayoung.lee@samsung.com](mailto:dayoung.lee@samsung.com)
> 
> For #768
> 
> ## GIF
> ![1017-refactor](https://user-images.githubusercontent.com/17171963/196078229-e93b2eae-cfb2-40ce-9782-409dc8071064.gif)

dayo09:
### How vscode support 'Refactor' of symbols

![1019-refactor-vscode](https://user-images.githubusercontent.com/17171963/196597822-96d086c0-de28-4652-be7b-a89f9fedf832.gif)


- Command name: 'Rename Symbol'
- Options:Enter to 'Rename' & Shift+Enter to 'Preview'
- 'Preview' shows Renaming Candidates
- It edits the document, not saving it.

dayo09:
(Step 1) Firstly, I will introduce the simple refactoring command which we have discussed so far. It will ask user if he/she will do 'Refactor', and if user answers 'yes', it will alter all the related configs and save. Command name can be 'Rename Base Model'
(Step 2) I will make it not to save but to only edit using CfgEditor.
(Step 3) Consider if I can show the 'panel'(as you can see in above gif, vscode shows the list of lines to be changed) to let user select which config file to be changed.

PR #1361, opened 2022-10-17
[OneExplorer] Support refactoring by renaming

dayo09:
This commit enables refactoring by renaming,
so that the model file renaming can be automatically applied to related
config files.
It renames all the changed model file names in children config files.

----

Waiting for #1388 to be merged
For #768

## GIF (Updated, 2022-10-19)

NOTE: Command name is changed from `Refactor` to `Rename Model` after taking this video. 
1019-refactor

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -533,6 +533,8 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
           }),
       vscode.commands.registerCommand('one.explorer.delete', (node: Node) => provider.delete(node)),
       vscode.commands.registerCommand('one.explorer.rename', (node: Node) => provider.rename(node)),
+      vscode.commands.registerCommand(
+          'one.explorer.refactor', (node: Node) => provider.refactor(node)),
     ];
 
     if (provider.isLocal) {
@@ -692,6 +694,68 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
         });
       }
     });
+  }
+
+  /**
+   * Refactor BaseModel or Product(TBD)
+   * It renames the file and changes the corresponding path in its referring config files.
+   * @command one.explorer.refactor
+   * @todo support refactoring of Product
+   * @todo prohibit special characters from new name for security ('..', '*', etc)
+   */
+  async refactor(node: Node): Promise<void> {
+    assert.ok(node.type === NodeType.baseModel);
+
+    // Ask the new name of the model file
+    const newname = await this.askNewName(node);
+    if (!newname) {
+      return;
+    }
+
+    const newpath = `${path.dirname(node.uri.fsPath)}/${newname}`;
+    const children = node.getChildren();
+
+    // If it has no child, simply rename it
+    if (children.length === 0) {
+      vscode.workspace.fs.rename(node.uri, vscode.Uri.file(newpath)).then(() => {
+        this.refresh();
+      });
+      return;
+    }
+
+    // Ask whether the user want to change the config files
+    const askChangingCfgs = () => vscode.window.showInformationMessage(
+        `Change corresponding fields in these following files?`, {
+          detail: `${children.length} file(s): ${children.map(node => ' ' + node.name).toString()}`,
+          modal: true
+        },
+        'Yes');
+
+    if (!await askChangingCfgs()) {
+      return;
+    }
+
+    // Refactor config files
+    const refactorCfgs = () => {
+      return Promise.all(children.map(child => {
+        const cfgObj = OneStorage.getCfgObj(child.path);
+        if (cfgObj) {
+          const oldpath = node.path;
+          return cfgObj.updateBaseModelField(oldpath, newpath).then(() => {
+            return Logger.info(
+                'OneExplorer', `Replaced ${oldpath} with ${newpath} in ${child.path}`);
+          });
+        }
+      }));
+    };
+
+    refactorCfgs()
+        .then(() => vscode.workspace.fs.rename(node.uri, vscode.Uri.file(newpath)).then(() => {
+          this.refresh();
+        }))
+        .catch(() => {
+          Logger.error('OneExplorer', `Refactor`, `Failed to refactor ${node.path}`);
+        });
   }
 
   /**
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -742,7 +742,7 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
         if (cfgObj) {
           const oldpath = node.path;
           return cfgObj.updateBaseModelField(oldpath, newpath).then(() => {
-            return Logger.info(
+            Logger.info(
                 'OneExplorer', `Replaced ${oldpath} with ${newpath} in ${child.path}`);
           });
         }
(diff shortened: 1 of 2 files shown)


dayo09:
@lemmaa I made this 'refactoring cfg' feature out as can be seen in GIF, I wonder if you have any better idea about the UI :-D

lemmaa:
> @lemmaa I made this 'refactoring cfg' feature out as can be seen in GIF, I wonder if you have any better idea about the UI :-D

@dayo09 ,  cool! Has it been tested on a model with multiple cfgs? :)

Apart from that, it would be good to think about whether `rename` is appropriate or `refactoring` or something other is good. UI such as `warning message` should be updated.

One more thing, does it work for files such as `*.circle` hanging under the `*.cfg` file?

dayo09:
@lemmaa 

> @lemmaa I made this 'refactoring cfg' feature out as can be seen in GIF, I wonder if you have any better idea about the UI :-D
(I moved the original comment to the issue, https://github.com/Samsung/ONE-vscode/issues/768#issuecomment-1280477169)

-----

> Has it been tested on a model with multiple cfgs? :)

Sure, it works well :-D

> Apart from that, it would be good to think about whether rename is appropriate or refactoring or something other is good. UI such as warning message should be updated.
> One more thing, does it work for files such as *.circle hanging under the *.cfg file?

Yes, I have thought about that. I persisted the command name as 'rename' because 'rename' command works naturally with the other files(*.circle, *.cfg).

(1) *.cfg: 'Rename' command works with *.cfg, without changing anything in ONE Explorer view. (changing cfg file's name doesn't affect any relationships between *.tflite or *.circle nodes)

(2) *.circle: When user renames *.circle, I expected that the user don't want to 'refactor' those in cfgs. Because it's an 'artifacts', if the user changes the name, it would be natural that it doesn't affect the cfg file.

> ...whether rename is appropriate or refactoring or something other is good. 

I think the command name for '*.tflite' can be changed to 'refactor', but for me, 'rename' sounds more straightforward. But I agree that we can come up with better name. How about `Rename (refactor)` ??

lemmaa:
> Yes, I have thought about that. I persisted the command name as 'rename' because 'rename' command works naturally with the other files(*.circle, *.cfg).
> 
> * *.cfg: 'Rename' command works with *.cfg, without changing anything in ONE Explorer view. (changing cfg file's name doesn't affect any relationships between *.tflite or *.circle nodes)

`Rename` and `Refactoring` are different operations. So, it is better to separate the two. It only confuses the user that the same command behaves differently depending on the target. It's not easy to understand either.

Shall we not allow rename for refactoring target? If so, we can keep the rename function.

> * *.circle, *.log, *.tvn, ...: When user renames *.circle, I expected that the user don't want to 'refactor' those in cfgs. IMO, because they are 'artifacts', if the user changes the name, it would be natural that it doesn't affect the cfg file.

But, it also exist multiple times in `*.cfg` file. So in terms of a single `*.cfg` file, for refactoring, `*.circle` mentioned multiple times may be more useful than `*.tflite` used only once. How can we determine the user's mind in this situation? :)

lemmaa:
> I think the command name for '*.tflite/onnx/pb' can be changed to 'refactor', but for me, 'rename' sounds more straightforward. But I agree that we can come up with the better name. How about `Rename (refactor)` ??

I think 'Rename' and 'Refactor' should exist as separate commands.

dayo09:
> I think 'Rename' and 'Refactor' should exist as separate commands.

Sounds good!

- `*.tflite/onnx/pb` : Refactor only
- `*.cfg`: Rename only
- `*.circle/log/tvn`: Refactor and Rename

Like this?

dayo09:
@lemmaa I have updated the PR based on what we discussed.

You can see that...
(1) Rename command is only supported for config, without warning messages at all.
(2) Refactoring command is only supported for base models, without warning messages at all.
(NOTE: Refactoring for `product`(circle, log, trace.json...) node is not yet supported, it's TBD)

IMO, Showing command name as "Rename All" might be possible, instead of "Refactor". Or, "Rename All(Refactor)" can be an option, too. For me, "Refactor" is a bit difficult to guess the behavior...

lemmaa:
> I personally suggest: showing command name as "Rename All" instead of "Refactor", or, possibily "Rename All(Refactor)". For me, "Refactor" is a bit difficult to guess the behavior...

Looking at the implementation again, especially the popup dialog, I could have a different thought about this.

As now, if the popup informs that other files can be changed together and receives the user's selection about it, it would not be bad to unify `Rename` and `Refactor` to `Rename`. This is because the popup itself sufficiently explains the refactoring and obtains the user's consent. :)

lemmaa:
@dayo09 , this is a separate story, i.e. you can proceed independently of this PR.

Is it a good choice to save the modified `*.cfg` file by default when refactoring? Wouldn't it be better to open and maintain it in the editor without saving it? 
Do you know how other IDEs handle it?

I can't remember well because it's been so long, but I've used the refactoring function in android studio in the past. It is usually a task to change the variable or function name throughout the project. It would be good to check how other IDEs work for refactoring. :)

dayo09:
> Therefore, if Cancel is selected, wouldn't it be appropriate to cancel the whole refactor operation itself?

I agree. I missed that. I will fix it.

> Do you know how other IDEs handle it?

I haven't checked yet. I will check how it goes.

dayo09:
### How vscode support 'Refactor' of symbols

1019-refactor-vscode


- Command name: 'Rename Symbol'
- Options:Enter to 'Rename' & Shift+Enter to 'Preview'
- 'Preview' shows Renaming Candidates
- It edits the document, not saving it.

_Originally posted by @dayo09 in https://github.com/Samsung/ONE-vscode/issues/768#issuecomment-1283410366_

dayo09:
@lemmaa Considering how vscode supports 'Rename Symbol', I plan:

(Step 1) Firstly, I will introduce the simple refactoring command which we have discussed so far. It will ask user if he/she will do 'Refactor', and if user answers 'yes', it will alter all the related configs and save. Command name can be 'Rename Base Model' 
(Step 2) I will make it not to save but to only edit using CfgEditor.
(Step 3) Consider if I can show the 'panel'(as you can see in above gif, vscode shows the list of lines to be changed) to let user select which config file to be changed.

dayo09:
@Samsung/one-vscode It's ready. PTAL :-D (you can find the latest GIF on the top)

mhs4670go (dismissed):
LGTM

dayo09:
I have decided the command name as 'Rename Model', finally. Please give me any opinion about this. I will reflect the idea after this PR.

src/OneExplorer/OneExplorer.ts
```diff
@@ -694,6 +696,68 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+    }
+
+    // Refactor config files
+    const refactorCfgs = () => {
+      return Promise.all(children.map(child => {
+        const cfgObj = OneStorage.getCfgObj(child.path);
+        if (cfgObj) {
+          const oldpath = node.path;
+          return cfgObj.updateBaseModelField(oldpath, newpath).then(() => {
+            return Logger.info(
```
llFreetimell, line 745: Maybe? :D

```suggestion
            Logger.info(
```

dayo09:
@llFreetimell Applied :-D @mhs4670go A subtle changes, PTAL :-D

lemmaa (approved):
LGTM. Thank you for your effort!

Question: Isn't it possible to keep the Expand/Collapse of the model node in its original state after executing the 'Rename Model' command? It seems to be inconvenient to collapse unconditionally.

mhs4670go (approved):

PR #1365, opened 2022-10-17
[CircleGraphCtrl] Add notes for received messages

seanshpark:
This will add notes for received messages that was removed while refactoring.


seanshpark:
related  #1352

llFreetimell (approved):

PR #1367, opened 2022-10-17
[MetadataViewer] Filter properties in for-in loop

llFreetimell:
This commit enables filtering properties in for-in loop by if statement.

---

This is to resolve static analysis warning, which is https://rules.sonarsource.com/javascript/RSPEC-1535.
Although diff is little bit complicated, only if statements are added.


llFreetimell:
@dayo09 
PTAL :D 
Although diff is dirty, actual code diff is adding if :)

dayo09 (approved):

PR #1371, opened 2022-10-17
[Toolchain] Return consistent type

llFreetimell:
`run` function in `ToolchainProvider` was not returning consistent type. This commit will fix it.

### DIFF
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -287,6 +287,8 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
     if (proceed) {
       return this._run(cfg);
     }
+
+    return false;
   }
 
   public setDefaultToolchain(tnode: ToolchainNode): boolean|undefined {


src/Toolchain/ToolchainProvider.ts
```diff
@@ -287,6 +287,8 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
     if (proceed) {
       return this._run(cfg);
     }
+
+    return false;
```
dayo09, line 291: It's return type is `Promise<boolean|undefined>`, and it seems that this._run(cfg) returns the execution result in boolean, whereas it returns undefined when it's not proceeded. Are you sure that this change will not affect the functionality?

llFreetimell: I thought that `false` was okay because `_run` function returns `false` when it is not ready to run.

https://github.com/Samsung/ONE-vscode/blob/aa7cade6e75f239f0b501eaedb494e45a824033e/src/Toolchain/ToolchainProvider.ts#L275-L281

But I am not sure because I am not the original author... :(
@jyoungyun I wonder why both `boolean` and `undefined` are used for the function...! And which would be correct for this situation?

dayo09: I will remove myself from reviewer :-D I think her approval will be enough!

jyoungyun: Well, I don't think this change affects the original process. Since the original operation is to return the `boolean|undefined` type, this change does not matter because it returns the boolean.

> I wonder why both boolean and undefined are used for the function...! And which would be correct for this situation?

It doesn't mean much. We can change it so that it does not return undefined values.

jyoungyun: Related draft PR: #1383

llFreetimell:
This PR is replaced with #1383 :)
Close!

PR #1366, opened 2022-10-17
[VisqProvider] Load visq.json

seanshpark:
This will enable to load visq.json file to visualize quantization error.

ONE-vscode-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #1374, opened 2022-10-17
[visq] Use path.join

seanshpark:
This will revise to use path.join method.


seanshpark:
from https://github.com/Samsung/ONE-vscode/pull/1366#discussion_r996797308

mhs4670go (approved):
LGTM

PR #1381, opened 2022-10-18
[OneExplorer] Allow rename command only for config files

dayo09:
This commit allows rename command only for config files. By this, base model and products renaming is restricted
and it will be replaced with 'Refactoring' command.

This commit also include some code cleanings.

----

FOR REVIEWERS,

This commit is from discussion https://github.com/Samsung/ONE-vscode/pull/1361
It's decided to enable rename command only for config files,
and support refactoring command for other files.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -206,7 +206,7 @@ class NodeFactory {
     //
     // A 'TreeDataProvider<element>' expects every elements (Node) to be correspond to visible
     // TreeItem, so let's not build hidden nodes.
-    if (attr && attr?.canHide === true && OneTreeDataProvider.didHideExtra === true) {
+    if (attr && attr?.canHide && OneTreeDataProvider.didHideExtra) {
       return undefined;
     }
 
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -206,8 +206,7 @@ class NodeFactory {
     //
     // A 'TreeDataProvider<element>' expects every elements (Node) to be correspond to visible
     // TreeItem, so let's not build hidden nodes.
-    if(attr && attr?.canHide === true && OneTreeDataProvider.didHideExtra === true)
-    {
+    if (attr && attr?.canHide === true && OneTreeDataProvider.didHideExtra === true) {
```
seanshpark, line 209: Q) can we use `if (attr && attr?.canHide && OneTreeDataProvider.didHideExtra) {` without `=== true)` ?

seanshpark (approved):
LGTM

src/OneExplorer/OneExplorer.ts
dayo09: Yes :-D I just explicitly wrote it down as those fields can be 'undefined' or 'false'(either of them is false), but I think it's also clear with the shortened form.

dayo09, line 209:
```suggestion
    if (attr && attr?.canHide && OneTreeDataProvider.didHideExtra ) {
```

seanshpark (approved):
LGTM

jyoungyun (approved):
LGTM

PR #1382, opened 2022-10-18
[ConfigObject] Add updateBaseModelField

dayo09:
This commit add a function to update base model field in the config object. It finds 'input_path' field and changes the old path to the new path.

This commit also includes some Cfg type definitions which is required subsequently.

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -17,12 +17,25 @@
 import * as fs from 'fs';
 import * as ini from 'ini';
 import * as path from 'path';
+import {TextEncoder} from 'util';
 import * as vscode from 'vscode';
 
 import {RealPath} from '../Utils/Helpers';
 import {Logger} from '../Utils/Logger';
 
 import {Artifact, Locator, LocatorRunner} from './ArtifactLocator';
+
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

seanshpark:
About OneExplorer vs CfgEditor...
- IMHO, changing `.cfg` file is main role of CfgEditor
- We may revise to support one-infer, one-partition and others, that may have to update this too
- If CfgEditor provides some service (not sure how as of now), OneExplorer can ask CfgEditor to change the file path, and update the cfg file

dayo09:
@seanshpark Sure :-D 

Speaking about it, I think the current CfgEditor can apply the refactoring's change immediately (as it did to the changes from text editor) but I will check it again when I make the main PR.

seanshpark:
> current CfgEditor can show the refactoring's change immediately in the editor 

I mention about the owner of responsibility of changing `.cfg` files, not the reflection of changes of `.cfg` files.

dayo09:
@seanshpark I know. I just added it.

jyoungyun (approved):
LGTM

(Q) Where does the `updateBaseModelField` function call?
If you have a draft PR for this change, please add the link.

PR #1383, opened 2022-10-18
[Toolchain] Do not return undefined type in ToolchainProvider

jyoungyun:
This commit does not return undefined type in Toolchain Provider file.

### DIFF
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -218,7 +218,7 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
   }
 
   /* istanbul ignore next */
-  public uninstall(tnode: ToolchainNode): boolean|undefined {
+  public uninstall(tnode: ToolchainNode): boolean {
     const notifyUninstalled = () => {
       vscode.window.showInformationMessage(`Uninstallation was successful.`);
       if (DefaultToolchain.getInstance().isEqual(tnode.toolchain)) {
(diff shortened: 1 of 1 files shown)


llFreetimell:
@jyoungyun Could you convert this PR to official PR? :D

llFreetimell (approved):
LGTM, thank you :)

llFreetimell:
@dayo09 PTAL :)

src/Toolchain/ToolchainProvider.ts
```diff
@@ -243,9 +243,10 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
     gToolchainEnvMap[backendName]
         .uninstall(tnode.toolchain)
         .then(() => notifyUninstalled(), () => notifyError());
```
dayo09, line 245: Shouldn't it be 'return false' for notifyError() case..?

dayo09: Q. Do you define the function to return 'true' if it has tried running, no matter what the result is?

jyoungyun: The return value of this function does not mean much. Only valid case is when testing, but the test for installation and uninstall does not reach the case you mentioned.

dayo09: > The return value of this function does not mean much.

If so, how about returning void? Or if it's gonna return boolean, could you elaborate it somewhere (comment of so)

jyoungyun: Sure. I will fix it in next PR. :)

dayo09 (approved):

PR #1384, opened 2022-10-18
[media/CircleGraph] Apply visq value and color

seanshpark:
This will apply visq quantization error value in node and color value according to the error value.

### DIFF
--- a/media/CircleGraph/index.js
+++ b/media/CircleGraph/index.js
@@ -347,6 +347,28 @@ host.BrowserHost = class {
         }
     }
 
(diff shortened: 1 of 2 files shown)


seanshpark:
for #1350, from #1351

seanshpark:
this change will apply node name and colors like in https://github.com/Samsung/ONE-vscode/issues/1350#issuecomment-1278612881

jyoungyun (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #1386, opened 2022-10-19
[Visquv] review and revision


seanshpark:
Initial version has landed but with not much of thoughts with UI and features.
This issue is to gather usage feedback and add todo actions for better usage.

seanshpark:
CC @jinevening

seanshpark:
TODOs (to be added more)
- [ ] show quantization error value in `NODE PROPERTIES`
- [ ] walk through menu features and remove if not necessary for `Visquv`.
- [ ] ...

seanshpark:
Internal refactoring
- [ ] add quantization error value to `node` property
- [ ] ...

seanshpark:
Show in properties side bar;
![image](https://user-images.githubusercontent.com/4616940/196581639-5d0fb7cb-6fe4-4a23-84ac-b5523bca44b3.png)

@jinevening , I've used `quant-error` field name... do you have any better display name for this?

seanshpark:
Internal bug fixes
- [ ] fix VisqViewer to reflect model path change
- [ ] ...

jinevening:
> @jinevening , I've used quant-error field name... do you have any better display name for this?

Is it possible to use the metric name in the 'meta' section?

seanshpark:
> Is it possible to use the metric name in the 'meta' section?

Ah, yes :)

seanshpark:
done for now...

PR #1387, opened 2022-10-19
DRAFT revise node properties

seanshpark:
on-going draft to revise graph node properties.


seanshpark:
for #1386

seanshpark:
done

PR #1388, opened 2022-10-19
[OneExplorer] Revisit rename command

dayo09:
This commit cleans up rename command.
It extracts 'askNewName' and 'validateNewPath' functions from rename command so that they can be used by other functions.

---

For #1359


seanshpark (approved):
LGTM

llFreetimell (approved):

PR #1396, opened 2022-10-20
[CircleViewer] Enable reload when changed

seanshpark:
This will enable to reload when view model has changed in filesystem.

### DIFF
--- a/src/CircleGraph/CircleViewer.ts
+++ b/src/CircleGraph/CircleViewer.ts
@@ -45,6 +45,8 @@ class CircleViewer extends CircleGraphCtrl {
 export class CircleViewerDocument implements vscode.CustomDocument {
   private readonly _uri: vscode.Uri;
   private _circleViewer: CircleViewer[];
+  private _fileWatcher: vscode.FileSystemWatcher|undefined;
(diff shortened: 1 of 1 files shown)


src/CircleGraph/CircleViewer.ts
```diff
@@ -83,10 +91,27 @@ export class CircleViewerDocument implements vscode.CustomDocument {
...
+    if (this._reloadTimer) {
+      clearTimeout(this._reloadTimer);
+    }
+    this._reloadTimer = setTimeout(() => {
+      this._circleViewer.forEach((view) => {
+        if (view.owner(panel)) {
+          view.loadContent();
+        }
+      });
+    }, 500);
```
seanshpark, line 113: delay with timer is used to not to make too many reloads while model is in some continuous update.

mhs4670go (approved):
LGTM

dayo09 (approved):

src/CircleGraph/CircleViewer.ts
dayo09: (optional) Could you add that as a comment..? It will be helpful to understand the code later.

PR #1408, opened 2022-10-20
[CircleGraph] Add setModel

seanshpark:
This will add setModel method to set/update model path.


mhs4670go (approved):
LGTM

seanshpark:
This will be used for following updates to provide refresh feature when loaded `*.visq.json` file has changed.

Issue #1399, opened 2022-10-20
[Visquv] Export to image


seanshpark:
Let's find out how we can provide export visq graph to an image.
calling `view.export()` doesn work maybe cause of security or implementation reasons.

seanshpark:
from https://github.com/Samsung/ONE-vscode/issues/1386#issuecomment-1283179415

seanshpark:
This feature came with offline talk with @jinevening
- developers can communucate better with posting or sending exported images through email or github issues.
- we can provide a viewer tool in ONE, like the one I made in draft PR as alternative solution

seanshpark:
exported somehow ...
[mobilenet_v1_keepdims_relu6.0to33.circle.svg.zip](https://github.com/Samsung/ONE-vscode/files/9828375/mobilenet_v1_keepdims_relu6.0to33.circle.svg.zip)

seanshpark:
done

PR #1410, opened 2022-10-21
[media/CircleGraph] Style title for visquv

seanshpark:
This will add title for visquv styles to fast query.


seanshpark:
for #1399 , from #1387

seanshpark:
FYI https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet

mhs4670go (approved):
LGTM

PR #1412, opened 2022-10-21
[media/CircleGraph] Revise blob export

seanshpark:
This will revise blob export method to post message to vscode extension.


seanshpark:
for #1399 , from #1387

yunjayh (approved):
LGTM

PR #1402, opened 2022-10-20
DRAFT: [OneExplorer] Revise Refresh to re-build only the changed node

dayo09:
This commit revises refresh command and make it more efficient to re-build only the changed node.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

----

For #1401

PR #1416, opened 2022-10-21
[OneExplorer] Revise Refresh to re-build only the changed node

dayo09:
This commit revises refresh command and make it more efficient
to re-build only the changed node.

---

From #1402

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -177,6 +177,10 @@ abstract class Node {
     return this._childNodes!;
   }
 
+  resetChildren(): void {
+    this._childNodes = undefined;
+  }
+
   get path(): string {
     return this.uri.fsPath;
   }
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -505,9 +509,49 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+                case NodeType.baseModel:
+                  OneStorage.resetBaseModel(node.path);
+                  break;
+                case NodeType.config:
+                  OneStorage.resetConfig(node.path);
+                  break;
+                default:
+                  break;
+              }
+            }
```
dayo09, line 548:
```suggestion
            };
```

dayo09:
### Before
Refreshing all the nodes on deletion of any file.

1024-refresh-before

### After
Refreshing only the parent node of the deleted file.
1024-refresh

src/OneExplorer/OneStorage.ts
```diff
@@ -162,4 +162,18 @@ export class OneStorage {
...
   }
+
+  public static resetBaseModel(path: string): void {
+    delete OneStorage.get()._baseModelToCfgsMap[path];
+    Logger.debug('OneStorage', `Base Mode Path(${path}) is removed.`);
+  }
+
+  public static resetConfig(path: string): void {
+    delete OneStorage.get()._cfgToCfgObjMap[path];
+    Object.entries(OneStorage.get()._baseModelToCfgsMap).forEach(([modelpath]) => {
```
llFreetimell, line 173: I thought following code is correct but not sure... @.@
Key of `_baseModelToCfgsMap` is array?

```suggestion
    Object.entries(OneStorage.get()._baseModelToCfgsMap).forEach((modelpath) => {
```

dayo09: Object.entries returns the given object's [key, value] list in an array.
For example,
[ ['model.tflite', 'model.cfg'],
  ['A.tflite', 'A.cfg']]

 will be `Object.entries(OneStorage.get()._baseModelToCfgsMap)`

I obtained the first value as modelpath within forEach statuement by destructuring the array. So it's the key.

llFreetimell (approved):
LGTM

dayo09:
@Samsung/one-vscode PTAL :-D

mhs4670go (approved):
LGTM

PR #1418, opened 2022-10-24
[OneExplorer] Remove unused getCfgList()

dayo09:
This commit removes unused function getCfgList().
It's been replaced with OneStorage class's other functions and it's not used anywhere now.


llFreetimell (approved):
LGTM

Issue #1417, opened 2022-10-23
[OneExplorer] Multi-root workspaces


dayo09:
## What?

Let's support multi-root workspaces

Q. Can it be simply implemented by supporting multiple workspace root directories..?

dayo09:
### Let's Inspect Workspace Folders 

#### No folder opened

When there is no folder opened, 
![1024-no-folder](https://user-images.githubusercontent.com/17171963/197429699-1bc33fab-b0d0-4997-aad1-de813636a2db.png)
vscode.workspace.workspaceFolders is [], not undefined but an empty array.

#### When a folder is added/removed from the workspace,

The extension is always refreshed, therefore `extensions.ts` runs every time.

```ts
/**
 * @brief Get Workspace root folder as string
 * @note  will throw if not working in workspace mode.
 */
export function obtainWorkspaceRoot(): string {
  const workspaceFolders = vscode.workspace.workspaceFolders;
  if (!workspaceFolders) {
    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceFolders');
    // TODO revise message
    throw new Error('Need workspace');
  }

  // TODO support active workspace among the multiple workspaceFolders
  // TODO support multi-root workspace
  if (workspaceFolders.length > 1) {
    Balloon.info('Warning: Only the first workspace directory is currently supported');
  }
  const workspaceRoot = workspaceFolders[0].uri.path;
  if (!workspaceRoot) {
    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceRoot');
    // TODO revise message
    throw new Error('Need workspace');
  }
  Logger.debug(logTag, 'obtainWorkspaceRoot:', workspaceRoot);

  return workspaceRoot;
}
```

dayo09:
## GIF

![1025-muilti-root](https://user-images.githubusercontent.com/17171963/197670215-3736851d-f60f-41e9-b880-490b5beb0ef2.gif)

_Originally posted by @dayo09 in https://github.com/Samsung/ONE-vscode/issues/1419#issuecomment-1289903701_
      


Done as can be seen above, by #1419

PR #1419, opened 2022-10-24
[OneExplorer] Support multi-root workspaces

dayo09:
This commit supports multi-root workspaces.

---

For #1417

Waiting for #1424 to be merged

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -21,7 +21,7 @@ import {TextEncoder} from 'util';
 import * as vscode from 'vscode';
 
 import {CfgEditorPanel} from '../CfgEditor/CfgEditorPanel';
-import {getErrorMessage, obtainWorkspaceRoot} from '../Utils/Helpers';
+import {obtainWorkspaceRoots} from '../Utils/Helpers';
 import {Logger} from '../Utils/Logger';
 
 import {ArtifactAttr} from './ArtifactLocator';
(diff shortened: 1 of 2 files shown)


dayo09:
## GIF

1025-muilti-root

src/OneExplorer/OneExplorer.ts
```diff
@@ -893,23 +888,23 @@ input_path=${modelName}.${extName}
 
   getChildren(element?: Node): vscode.ProviderResult<Node[]> {
     if (!element) {
-      element = this.getTree();
+      return this.getTree();
```
dayo09, line 891: FOR REVIEWERS
Here, returning this.getTree() is enough. Because the getTree() returns `the first directory node(s)` which are virtually child(ren) node(s) of the ONE Explorer View's root.

yunjayh (approved):
LGTM

mhs4670go (approved):
LGTM

src/OneExplorer/OneExplorer.ts
```diff
@@ -893,23 +888,23 @@ input_path=${modelName}.${extName}
...
   }
 
   /**
    * Get the root of the tree
    */
-  private getTree(): Node|undefined {
-    if (!this.workspaceRoot) {
+  private getTree(): Node[]|undefined {
+    if (!this._workspaceRoots || this._workspaceRoots.length === 0) {
       return undefined;
```
llFreetimell, line 902: Not directly related with this PR, but how about returning empty array instead of `undefined`? :)
I am not sure but `getChildren(element)` can return `undefined` if `getTree()` returns `undefined`...!
(But it seems to work well currently @.@)

yunjayh (approved):

src/OneExplorer/OneExplorer.ts
dayo09: Yes, I understand the point of your suggestion. That's actually the exact strategy I used in the previous PR(introducing obtainWorkspaceRoots) to clarify the return and simplify the caller. Here, it works correctly in both ways and the caller function doesn't bother as it returns vscode.ProviderResult<Node[]>. I think it's optional, so I would like to proceed as is -  I wouldn't object to it if you create the PR, though :-D

llFreetimell (approved):
LGTM

PR #1420, opened 2022-10-24
[Helpers] Fix obtainWorkspaceRoot

dayo09:
This commit rewrites obtainWorkspaceRoot.
It fixes the bug of throwing unknown error by accessing workspaceFolders[0].uri.path when worksapceFolder is an empty array but not undefined.

---

For #1419

### DIFF
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -67,27 +67,24 @@ export class RealPath {
 
 /**
  * @brief Get Workspace root folder as string
- * @note  will throw if not working in workspace mode.
+ * @return Return only the first workspaceFolder if multiple root exists, with showing a notfication
+ *     balloon.
+ *         Throws if there is no workspace.
  */
 export function obtainWorkspaceRoot(): string {
   const workspaceFolders = vscode.workspace.workspaceFolders;
-  if (!workspaceFolders) {
-    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceFolders');
-    // TODO revise message
+
+  if (!workspaceFolders || workspaceFolders.length === 0) {
+    Logger.info(logTag, 'obtainWorkspaceRoot', 'No WorkspaceFolders');
     throw new Error('Need workspace');
   }
 
-  // TODO support active workspace among the multiple workspaceFolders
   // TODO support multi-root workspace
   if (workspaceFolders.length > 1) {
     Balloon.info('Warning: Only the first workspace directory is currently supported');
   }
+
   const workspaceRoot = workspaceFolders[0].uri.path;
-  if (!workspaceRoot) {
-    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceRoot');
-    // TODO revise message
-    throw new Error('Need workspace');
-  }
   Logger.debug(logTag, 'obtainWorkspaceRoot:', workspaceRoot);
 
   return workspaceRoot;
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -68,8 +68,7 @@ export class RealPath {
 /**
  * @brief Get Workspace root folder as string
  * @return Return only the first workspaceFolder if multiple root exists, with showing a notfication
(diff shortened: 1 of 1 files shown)


jyoungyun (dismissed):
LGTM

src/Utils/Helpers.ts
```diff
@@ -67,27 +67,24 @@ export class RealPath {
 
 /**
  * @brief Get Workspace root folder as string
- * @note  will throw if not working in workspace mode.
+ * @return Return only the first workspaceFolder if multiple root exists, with showing a notfication
+ *     balloon.
+ *         Throws if there is no workspace.
```
seanshpark, line 72:
```suggestion
 * @return Return only the first workspaceFolder if multiple root exists, with showing a notfication
 *         balloon. Throws if there is no workspace.
```

src/Utils/Helpers.ts
```diff
@@ -67,27 +67,24 @@ export class RealPath {
...
  */
 export function obtainWorkspaceRoot(): string {
   const workspaceFolders = vscode.workspace.workspaceFolders;
-  if (!workspaceFolders) {
-    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceFolders');
-    // TODO revise message
```
seanshpark, line 76: `Need workspace` is as-is but removed this `TODO`. Do we not need to change this message?
IMHO, this looks not friendly with short `Need workspace`.

yunjayh (approved):
LGTM

src/Utils/Helpers.ts
dayo09: There are so many tests that checks this 'Need workspace' message. And I thought that as many people remained it, it's good to keep as is. However if you allow, we can revise it to `There is no workspace.`, which looks rather formal.

src/Utils/Helpers.ts
```diff
@@ -67,27 +67,23 @@ export class RealPath {
 
 /**
  * @brief Get Workspace root folder as string
- * @note  will throw if not working in workspace mode.
+ * @return Return only the first workspaceFolder if multiple root exists, with showing a notfication
+ *         balloon. Throws if there is no workspace.
```
dayo09, line 71:
```suggestion
 *         balloon. Throw if there is no workspace.
```

src/Utils/Helpers.ts
seanshpark: > And I thought that as many developers kept it while using it

Seems this is another context. It would be better to remove this `TODO` in another PR.

> However if you allow

I don't think I'm the person to allow or not for this message.

> we can revise it to There is no workspace. in the next PR, which looks rather formal.

This seems a choice to revise the mssage.

Anyway, for this PR, it would be better not to remove `TODO` message in this PR.

dayo09: If you don't mean to revise the message, then it's no meaning that the comment remain there. Anyways I will revive the TODO message then.

src/Utils/Helpers.ts
```diff
@@ -67,27 +67,23 @@ export class RealPath {
...
+ *         balloon. Throw if there is no workspace.
  */
 export function obtainWorkspaceRoot(): string {
   const workspaceFolders = vscode.workspace.workspaceFolders;
-  if (!workspaceFolders) {
-    Logger.error(logTag, 'obtainWorkspaceRoot: NO workspaceFolders');
-    // TODO revise message
+
+  if (!workspaceFolders || workspaceFolders.length === 0) {
+    Logger.info(logTag, 'obtainWorkspaceRoot', 'No WorkspaceFolders');
```
dayo09, line 77:
```suggestion
    Logger.info(logTag, 'obtainWorkspaceRoot', 'No WorkspaceFolders');
    // TODO revise message
```

seanshpark (approved):
LGTM

yunjayh (approved):

dayo09:
@jyoungyun PTAL :-D

jyoungyun (approved):
LGTM

PR #1421, opened 2022-10-24
[OneExplorer] Simplify getting workspaceRoot 

dayo09:
This commit simplifies the error handling logic
when ONE Explorer sets workspace root.

---

For #1419

### DIFF
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -69,7 +69,7 @@ export function getErrorMessage(error: unknown) {
   if (error instanceof Error) {
     return error.message;
   }
-  return String(error);
+  return JSON.stringify(error);
 }
 
 /**
--- a/src/Utils/Helpers.ts
+++ b/src/Utils/Helpers.ts
@@ -65,6 +65,13 @@ export class RealPath {
   }
 }
 
+export function getErrorMessage(error: unknown) {
+  if (error instanceof Error) {
+    return error.message;
+  }
+  return String(error);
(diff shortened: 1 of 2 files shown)


src/Utils/Helpers.ts
```diff
@@ -65,29 +65,33 @@ export class RealPath {
   }
 }
 
+export function getErrorMessage(error: unknown) {
+  if (error instanceof Error) {
+    return error.message;
+  }
+  return String(error);
```
yunjayh, line 72: Q) Does this work for every type of instances?

yunjayh (approved):
LGTM

src/Utils/Helpers.ts
dayo09: I just copy-pasted the line, so I will check. Thanks for pointing it out.

dayo09: https://stackoverflow.com/questions/5612787/converting-an-object-to-a-string

It seems that the javascript String(); function converts any object into string.

Or, `JSON.stringify(yourobject)` looks more explicit, so I can replace it with this.

yunjayh: As the link explains, `String(obj)` might not give sufficient information for user. How about replacing it to `JSON.stringify`? It can be done in another PR :smile:

yunjayh (approved):
LGTM Thank you!!!

llFreetimell (approved):
LGTM

PR #1423, opened 2022-10-24
[visquv] Extract common method makeModelPath

seanshpark:
This will extract common method makeModelPath.


mhs4670go (approved):
LGTM

PR #1431, opened 2022-10-28
[CI] Change pull_request_target into pull_request

dayo09:
Let's change pull_request_target into pull_request to prevent security hazard.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung.lee <dayoung.lee@samsung.com>


llFreetimell:
Thanks for your effort but #1432 was merged :D

PR #1433, opened 2022-10-28
[CI] Make CI fail when license verification is failed

llFreetimell:
Until now, CI did not fail even when license verification is failed. It was because maintainers can judge the result by comment. However, as result of license verification is not created anymore, it is better for CI to fail when the verification is failed.


yunjayh (approved):
LGTM :+1:

PR #1435, opened 2022-10-28
[OneExplorer] Optimize Hide&Show

dayo09:
This commit optimizes hide&show command.
Previously, it reset all the internal data, the node tree and node map, and built them from scratch.
Now, it keeps all the internal data and only changes the getChildren() function's behavior.

Note that TreeView does build the tree by:
(1) getting children nodes from getChildren() and
(2) mapping the nodes to OneNode(Tree Item) using getTreeItem() Therefore, to decide whether to display node or not, getChildren() should make decision.

ONE-vscode-DCO-1.0-Signed-off-by: Dayoung.lee <dayoung.lee@samsung.com>


jyoungyun (approved):
LGTM

yunjayh (approved):
LGTM

Issue #1422, opened 2022-10-24
[OneExplorer] Keep collapse state of node


dayo09:
## What?

Let's support https://github.com/Samsung/ONE-vscode/issues/768#issuecomment-1283423734

> LGTM. Thank you for your effort!
> 
> Question: Isn't it possible to keep the Expand/Collapse of the model node in its original state after executing the 'Rename Model' command? It seems to be inconvenient to collapse unconditionally.

_Originally posted by @lemmaa in https://github.com/Samsung/ONE-vscode/pull/1361#pullrequestreview-1147570225_

### Prerequisite

- [x] #1401

dayo09:
### How?
```
Node node -> node.getTreeItem() -> OneNode oneNode
// oneNode.node === node
```

OneNode has a collapsible state

When refreshed, we need to pass 'initialCollapsibleState' from Node to OneNode, which must be disposed once it's used.
Otherwise, the state will be applied everytime the node is re-builded.....

dayo09:
### How? (2)

By using `onDidRenameFiles`, we don't need to re-create all the Node&OneNode but just replace the url and the file name.
Then, the other contexts including collapsible state will be easy to keep

```
      vscode.workspace.onDidRenameFiles((e)=>{
        e.files.forEach(file=>{
          console.log(file.oldUri, file.newUri);
        });
      }),
```

dayo09:
### DID IT!

![1031-rename-keep](https://user-images.githubusercontent.com/17171963/198916737-1dce7471-0dd5-4982-9215-7a8d2ed36448.gif)


- [ ] Replace rename api from vscode.fs to vscode.workspace.applyEdit and event listener from fileSystemWatcher to vscode.workspace.onDidRenameFiles.
- [ ] Add `id` to Node
- [ ] Update the node's uri while persisting the Node using rename api.

PR #1438, opened 2022-10-31
[OneExplorer] Add id to Node

dayo09:
This commit adds id to Node class.
`id` is a member variable kept by TreeItem,
which makes it possible to keep existing node when its name is changed.

---

For #1422


mhs4670go (approved):
LGTM

yunjayh (approved):
LGTM

PR #1439, opened 2022-10-31
[OneExplorer] Revisit NodeFactory.create

dayo09:
This commit revisits NodeFactory.create

- BUG FIX: It fixes a possible bug of hide & show.
If onecc was run in hide mode, the extra files would have never shown.
Let's fix it by removing this redundant legacy codes.

- CLEANING: It uses switch-case instead of if statement

---

Related to #1435 (This should have been applied with the PR)


yunjayh (approved):
LGTM

PR #1441, opened 2022-10-31
[OneExplorer] Use chai instead of assert

dayo09:
This commit is to use chai instead of assert.


mhs4670go (approved):
LGTM

PR #1448, opened 2022-11-03
[OneStorage] Rename function

dayo09:
This commit renames _initCfgList to _getCfgList.
because it doesn't keep the cfglist and just use and dispose it.

---

From #1446


dayo09:
@Samsung/one-vscode PTAL :-D

mhs4670go (approved):
LGTM

PR #1449, opened 2022-11-03
[OneStorage] Clarify ConfigObjMaps value

dayo09:
This commit removes null type from ConfigObjMap's value. Let's return undefined when the key is not found,
as the same as the other maps do.

---

From #1446


mhs4670go (approved):
LGTM

PR #1450, opened 2022-11-03
[OneStorage/test] Rearrange suites

dayo09:
This commit rearrange test suites of OneStorage.
before introducing more classes and its tests to OneStorage class.


mhs4670go (approved):
LGTM

PR #1452, opened 2022-11-04
[OneExplorer/Test] Clarify tests

dayo09:
This commit clarifies the OneExplorer unit tests by

(1) elaborate `create a base model` test.
(2) Add isDefined checker to easily notice undefined objects
(3) Replace `?` with `!` to assert unexpected undefined objects

Issue #1442, opened 2022-11-01
[OneExplorer] Add more APIs to OneStorage


dayo09:
Let's add more APIs to OneStorage.

Currently, there is only
- getCfgs
- getCfgObjs
- getNode

Let's add
- addCfg
- addBaseModel
- addProduct
- deleteCfg
- deleteBaseModel
- deleteProduct
- replaceCfg (rename)
- replaceBaseModel (rename)
- replaceProduct (rename)

PR #1453, opened 2022-11-04
[OneExplorer] Introduce BaseModelToCfgMap/CfgToCfgObjMap

dayo09:
This commit refactors OneStorage class
to divide its functionalities into CfgToCfgObjMap and BaseModelToCfgMap.
It includes unit tests.

----

Related to #1442
From #1446
Waiting for #1452

### DIFF
--- a/src/OneExplorer/OneStorage.ts
+++ b/src/OneExplorer/OneStorage.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {assert} from 'chai';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


src/OneExplorer/OneStorage.ts
```diff
@@ -24,12 +25,157 @@ import {Logger} from '../Utils/Logger';
...
+        assert.isOk(false, `Cannot reach here`);
+        break;
+    }
+  }
+
+  /**
+   * Update the data when cfg file's path is changed or the content is changed.
+   * @param type NodeType
+   * @param path config path to update
+   * @param newpath (optional) if exists, change the file path
```
yunjayh, line 75: I cannot fully understand this method to review. 😓 When will this method be called without the `newpath`? AFAIU it will reload the `ConfigObj` with the exactly same path before.

dayo09: This map is designed to handle the events defined in OneExplorer.

> When will this method be called without the newpath?

- When a config file is changed and `onChangedFile` event is triggered -> update the CfgObj (newpath === undefined) with the same old key
- When a config file is renamed and `onDidRenameFile` event is triggered -> update the CfgObj (newpath !== undefined) with the new key

yunjayh (approved):
LGTM Thank you!

src/OneExplorer/OneStorage.ts
```diff
@@ -24,12 +25,157 @@ import {Logger} from '../Utils/Logger';
 import {ConfigObj} from './ConfigObject';
 import {Node, NodeType} from './OneExplorer';
 
-interface StringListMap {
-  [key: string]: string[];
+export {
+  BaseModelToCfgMap as _unit_test_BaseModelToCfgMap,
+  CfgToCfgObjMap as _unit_test_CfgToCfgObjMap,
```
llFreetimell, line 30: This is not test-related files but they are exported as `_unit_test_*`, why...?

src/OneExplorer/OneStorage.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {assert} from 'chai';
```
llFreetimell, line 17: AKAIK, `chai` is for writing testcases.
Maybe following is what you intended?

```suggestion
import {assert} from 'assert';
```

dayo09: @llFreetimell

> AKAIK, chai is for writing testcases.

I thought chai is more advanced assertion library, supporting BDD also.

https://www.chaijs.com/guide/styles/

You can find that they introduce some assert/BDD apis. 

I searched on stackoverflow people discussing about it. It seems there is not exact guide whether to use chai to replace the assert, but there is no reason to do not so. https://stackoverflow.com/questions/36434068/node-js-assert-library-vs-other-assert-libraries.

src/OneExplorer/OneStorage.ts
llFreetimell: I see it :)
Just FYI, there are some workaround patterns such as https://stackoverflow.com/a/54116079 and https://stackoverflow.com/a/65422568.

llFreetimell (approved):

src/OneExplorer/OneStorage.ts
dayo09: Yes, actually the workaround used in the links are the same to this - it's by aliasing its internal functions when exporting :-D

PR #1456, opened 2022-11-14
[One Explorer] Set default value for delete function

llFreetimell:
Instead of using undefined, let's use default value.


dayo09 (approved):
👍

PR #1467, opened 2023-01-12
[OneExplorer] Fix ArtifactLocator

dayo09:
This commit fixes ArtifactLocator not to parse wrongly formatted ini file.

For example,
>
>  [one-import-tflite]
>  input_path="model.tflite"

This is parsed by python 'ini' module as
[one-import-tflite]=true
...


dayo09:
https://en.wikipedia.org/wiki/INI_file

There are many ini parsers and it seems they varies in its format parsing. This wiki pages does not mention about indentation.

llFreetimell (approved):
LGTM

jyoungyun (approved):
LGTM

PR #1469, opened 2023-01-13
[Toolchain] Fix the problem of one toolchain installation failure

jyoungyun:
This commit removes the package name from the apt-cache result.
Installation fails indefinitely due to an unexpected package name.


llFreetimell (approved):
LGTM

PR #1476, opened 2023-01-17
[OneStorage] Fix unit tests

dayo09:
This commit inserts OneStorage.reset() if a test case includes any file system change.
OneStorage's singleton instance reflects the filesystem context at the time it is first referred.
Therefore, to update OneStorage instance, OneStorage.reset() must be called after the writeFileSync in test.

Then, how come those tests passed so far? One possible guess is that Github Action's runner somehow keeps previous test's filesystem,
so that those tests have been randomly succeeded.
Let's fix the test.


llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

PR #1477, opened 2023-01-17
[ci] Use node16 by default in GitHub Actions

jyoungyun:
This commit uses checkout v3 and setup-node v3 by default in GitHub Actions workflows. The Node.js 12 actions are deprecated. The checkout v3 and setup-node v3 use Node.js 16.


llFreetimell (approved):
LGTM

Issue #1473, opened 2023-01-16
Publish extension to marketplace


jyoungyun:
## What

Let's publish ONE-vscode extension to public marketplace.

### Reference

- [Publishing Extensions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension)

jyoungyun:
### Account

- id : nnfw
- publisher : Samsung

### Test

- Use website page
   - https://marketplace.visualstudio.com/manage/publishers/Samsung
   - Successful
   - ![marketplace](https://user-images.githubusercontent.com/7223627/212627813-c40710fe-608e-4806-9234-d3891c361d4e.png)
   - (I didn't expect this file to be upload successfully. 😮 But it was a success. So,,,, v0.3.0 has been uploaded..... 👀)

- Use command line interface

jyoungyun:
Done!

PR #1480, opened 2023-01-17
[ci] Publish extension to marketplace in official release workflow

jyoungyun:
This commit publishes vsix extension file to marketplace with Samsung Publisher in Official release workflow.
This job requires VSCE_PAT secret environment variable.

Related issue: #1473 
Draft workflow: https://github.com/jyoungyun/vscode-test/actions/runs/3938041397/jobs/6736269110

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -190,6 +190,7 @@ jobs:
     uses: ./.github/workflows/publish-extension.yml
     with:
       prerelease: ${{ github.event.inputs.prerelease }}
+      publish: ${{ !inputs.draft && !inputs.prerelease }}
     secrets: inherit
 
   release:
--- a/.github/workflows/publish-extension.yml
+++ b/.github/workflows/publish-extension.yml
@@ -7,10 +7,17 @@ on:
         description: 'Identify the release as a prerelease (default: false)'
         type: string
         default: false
+      publish:
+        description: 'Publish extension to marketplace'
+        type: boolean
+        default: false
     outputs:
       package_name:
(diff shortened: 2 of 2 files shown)


.github/workflows/build-release.yml
```diff
@@ -190,6 +190,7 @@ jobs:
     uses: ./.github/workflows/publish-extension.yml
     with:
       prerelease: ${{ github.event.inputs.prerelease }}
+      publish: ${{ github.event.inputs.draft == 'false' && github.event.inputs.prerelease == 'false' }}
```
llFreetimell, line 193: According to following codes, both `draft` and `prerelease` type are boolean not string :)

https://github.com/Samsung/ONE-vscode/blob/36f886e3b09048a3fc981550169a721edc2e4be8/.github/workflows/build-release.yml#L11

jyoungyun: The `github.event.inputs.<name>` always string even its type is boolean.

> Note: The workflow will also receive the inputs in the github.event.inputs context. The information in the inputs context and github.event.inputs context is identical except that the inputs context preserves Boolean values as Booleans instead of converting them to strings.

from https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

jyoungyun: The latest guideline suggests to use `input` context instead of `github.event.inputs`. So I will change it.

llFreetimell (approved):
LGTM

dayo09 (approved):

Issue #1488, opened 2023-02-01
[cfgEditor] Implement run configuration by Ctrl+F7


stamalakhov:
### What?

Let's implement compilation of .cfg (runCfg) using  'Ctrl+F7' shortcut (see https://github.com/Samsung/ONE-vscode/issues/1482#issuecomment-1404607640)
### Why?

Currently clicking the button may not the best choice for a user, because some users prefer using key shortcuts.

stamalakhov:
Seems like we can close this one due to merged #1489.

PR #1489, opened 2023-02-02
[OneExplorer] Implement configuration run by Ctrl+F7

stamalakhov:
This commit implements configuration run by  'Ctrl+F7' shortcut.

Issue: #1488
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,9 +156,14 @@
       },
       {
         "command": "one.explorer.runCfg",
-        "title": "Run with the toolchain",
+        "title": "Run with the toolchain (Ctrl+F7)",
         "category": "ONE",
         "icon": "$(play)"
+      },
+      {
+        "command": "one.explorer.runSingleSelectedCfg",
+        "title": "Run selected with the toolchain",
+        "category": "ONE"
       },
       {
         "command": "one.explorer.rename",
@@ -245,6 +250,13 @@
         "command": "one.viewer.metadata.showFromOneExplorer",
         "title": "ONE: Show Metadata",
         "category": "ONE"
+      }
+    ],
+    "keybindings": [
+      {
+          "command": "one.explorer.runSingleSelectedCfg",
+          "key": "ctrl+f7",
+          "when": "OneExplorerView.active && one.explorer:hasSelectedCfg"
       }
     ],
     "menus": {
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -482,6 +482,8 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
 
   public static didHideExtra: boolean = false;
 
+  public static hasSelectedCfg: boolean = false;
+
   public static register(context: vscode.ExtensionContext) {
     const provider = new OneTreeDataProvider(context.extension.extensionKind);
 
@@ -489,6 +491,9 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
       treeDataProvider: provider,
       showCollapseAll: true,
       canSelectMany: true,
+    });
+    provider._treeView.onDidChangeSelection(() => {
+      provider.refreshCfgSelection();
     });
 
     let registrations = [
@@ -559,6 +564,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
       vscode.commands.registerCommand("one.explorer.runCfg", (node: Node) => {
         vscode.commands.executeCommand("one.toolchain.runCfg", node.uri.fsPath);
       }),
+      vscode.commands.registerCommand(
+        "one.explorer.runSingleSelectedCfg",
+        () => {
+          let selectedCfg = provider.getSelectedCfg();
+          if (selectedCfg && selectedCfg.length === 1) {
+            vscode.commands.executeCommand(
+              "one.toolchain.runCfg",
+              selectedCfg[0].uri.fsPath
+            );
+          } else if (!selectedCfg || selectedCfg.length === 0) {
+            // TODO: handle for none selection
+            vscode.window.showErrorMessage("No selected cfg file to run.");
+          } else if (selectedCfg.length > 1) {
+            // TODO: handle for multiple selection
+            vscode.window.showErrorMessage(
+              "Cannot run multiple cfg files. Please select a single cfg file."
+            );
+          }
+        }
+      ),
       vscode.commands.registerCommand("one.explorer.delete", (node: Node) =>
         provider.delete(node)
       ),
@@ -599,6 +624,11 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
       "setContext",
       "one.explorer:didHideExtra",
       OneTreeDataProvider.didHideExtra
+    );
+    vscode.commands.executeCommand(
+      "setContext",
+      "one.explorer:hasSelectedCfg",
+      OneTreeDataProvider.hasSelectedCfg
     );
   }
 
@@ -671,10 +701,29 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
         OneStorage.reset();
         // Reset the root in order to build from scratch (at OneTreeDataProvider.getTree)
         this._tree = undefined;
+        // Update selection info
+        this.refreshCfgSelection();
       }
       this._onDidChangeTreeData.fire(undefined);
     } else {
       this._onDidChangeTreeData.fire(node);
+    }
+  }
+
+  /**
+   * Refresh hasSelectedCfg parameter
+   */
+  refreshCfgSelection(): void {
+    const selectedCfg = this.getSelectedCfg();
+    const newValue: boolean =
+      selectedCfg !== undefined && selectedCfg!.length > 0;
+    if (OneTreeDataProvider.hasSelectedCfg !== newValue) {
+      OneTreeDataProvider.hasSelectedCfg = newValue;
+      vscode.commands.executeCommand(
+        "setContext",
+        "one.explorer:hasSelectedCfg",
+        OneTreeDataProvider.hasSelectedCfg
+      );
     }
   }
 
@@ -947,6 +996,25 @@ input_path=${modelName}.${extName}
           }
         });
       });
+  }
+
+  /**
+   * This function returns selected config items
+   * @returns selected config items
+   */
+  private getSelectedCfg(): Node[] | undefined {
+    if (this._treeView?.selection === undefined) {
+      return undefined;
+    }
+
+    const configs: Node[] = [];
+    for (var node of this._treeView!.selection) {
+      if (node.type === NodeType.config) {
+        configs.push(node);
+      }
+    }
+
+    return configs;
   }
 
   /**


seanshpark:
FYI, I don't think @hyunsik-yoon can participate the review and not sure @Seojeonguk can review too (I think @Seojeonguk was one of SSAFY member '22 or '21)

seanshpark:
@stamalakhov , if you like, I'll participate the review.

stamalakhov:
> @stamalakhov , if you like, I'll participate the review.

Ok. Thank you.

src/OneExplorer/OneExplorer.ts
```diff
@@ -562,6 +562,18 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+        () => {
+          let selectedCfg = provider.getSelectedCfg();
+          if (selectedCfg) {
+            vscode.commands.executeCommand(
+              "one.toolchain.runCfg",
+              selectedCfg.uri.fsPath
+            );
+          }
+        }
+      ),
```
seanshpark, line 576: maybe put this new block after `runCfg`

stamalakhov: Ahhh... sorry

src/OneExplorer/OneExplorer.ts
```diff
@@ -1015,4 +1027,16 @@ input_path=${modelName}.${extName}
 
     return this._tree;
   }
+
+  private getSelectedCfg(): Node | undefined {
```
seanshpark, line 1031: maybe put this after `createCfg` ?

src/OneExplorer/OneExplorer.ts
stamalakhov: done

src/OneExplorer/OneExplorer.ts
stamalakhov: ok

stamalakhov: done

stamalakhov: Ah. Ok. I see.

stamalakhov: fixed

src/OneExplorer/OneExplorer.ts
```diff
@@ -559,6 +559,18 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
       vscode.commands.registerCommand("one.explorer.runCfg", (node: Node) => {
         vscode.commands.executeCommand("one.toolchain.runCfg", node.uri.fsPath);
       }),
+      vscode.commands.registerCommand(
+        "one.explorer.runSingleSelectedCfg",
+        () => {
+          let selectedCfg = provider.getSelectedCfg();
+          if (selectedCfg && selectedCfg.length === 1) {
```
seanshpark, line 566: (sorry for adding comments one-by-one)
I'm not sure about current UI guide but I think running this with no selection or with two or more selected items won't do any action, resulting user to just wait and wait...
May be add a `TODO: handle for none or multiple selection` for now?

stamalakhov: ok

stamalakhov: fixed

seanshpark (dismissed):
LGTM thank you!

package.json
```diff
@@ -247,6 +252,12 @@
         "category": "ONE"
       }
     ],
+    "keybindings": [
+      {
+          "command": "one.explorer.runSingleSelectedCfg",
+          "key": "ctrl+f5"
```
jyoungyun, line 258: What about using another key binding? In Visual studio, Ctrl+F5 is used as the meaning of execution. I think it would be more appropriate to use this key binding to run tvn on simulator or target.

jyoungyun, line 258: What about using `ctrl+f7`? Because this operation compiles the cfg file to generate the tvn binary. In Visual studio, `ctrl+f7` uses as the meaning of compilation. It would be more appropriate to use this key binding to run tvn or simulator or target.

src/OneExplorer/OneExplorer.ts
```diff
@@ -559,6 +559,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+              "one.toolchain.runCfg",
+              selectedCfg[0].uri.fsPath
+            );
+          } else if (!selectedCfg || selectedCfg.length === 0) {
+            // TODO: handle for none selection
+            vscode.window.showErrorMessage("No selected cfg file to run.");
+          } else if (selectedCfg.length !== 1) {
+            // TODO: handle for multiple selection
+            vscode.window.showErrorMessage(
+              "No single selected cfg file to run."
```
dayo09, line 577: How about this message instead? 
```suggestion
              "Cannot run multiple cfg files. Please select a single cfg file."
```

src/OneExplorer/OneExplorer.ts
```diff
@@ -559,6 +559,26 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+          let selectedCfg = provider.getSelectedCfg();
+          if (selectedCfg && selectedCfg.length === 1) {
+            vscode.commands.executeCommand(
+              "one.toolchain.runCfg",
+              selectedCfg[0].uri.fsPath
+            );
+          } else if (!selectedCfg || selectedCfg.length === 0) {
+            // TODO: handle for none selection
+            vscode.window.showErrorMessage("No selected cfg file to run.");
+          } else if (selectedCfg.length !== 1) {
```
dayo09, line 574:
```suggestion
          } else if (selectedCfg.length > 1) {
```

How about this condition? It's more specific.

package.json
```diff
@@ -247,6 +252,12 @@
...
+      }
```
dayo09, line 259: I recommend adding 'when' condition here. 
Current way does enable your "runCfg" shortcut everywhere. 
You can make it only enabled when 
(1) One Explorer View is active.
(2) It has single selected cfg.

like
"when" : "OneExplorerView.active && one.explorer:hasSingleSelectedCfg"

'OneExplorerView.active' is a context value provided by vscode by default.
We can defined our own context one.explorer:hasSingleSelectedCfg here.

stamalakhov: Ok.

package.json
dayo09, line 258:
```suggestion
          "key": "ctrl+f5",
          "when" : "OneExplorerView.active && one.explorer:hasSingleSelectedCfg"
```

package.json
dayo09: You can find the example here: 
https://github.com/dayo09/ONE-vscode/tree/0203-review-1489

I checked it works well :-D

package.json
stamalakhov: But IMHO doing this way there is no chance for a user to get knowledge what actions to do for run configuration using shortcut (see https://github.com/Samsung/ONE-vscode/pull/1489#discussion_r1094316314).

stamalakhov: @dayo09 
IMHO even when there is just a single but not selected .cfg file, some users may still want to run it anyway using shortcut, because there is just a single .cfg file. IMHO. May be just 'OneExplorerView.active" would be sufficient to provide a user some hints how to run .cfg file ?

dayo09: Oh.... Sure :-D Maybe we can change this to one.explorer:hasSelectedCfg or so. 

IMO, anyways enabling this only for cfg nodes would make sense. 
Then maybe we can guide our users about the number of the selected cfgs, like the way the current pr does.

stamalakhov: ok

dayo09: > IMHO even when there is just a single but not selected .cfg file, some users may still want to run it anyway using shortcut, because there is just a single .cfg file. IMHO. May be just 'OneExplorerView.active" would be sufficient to provide a user some hints how to run .cfg file ?

Oh I just checked this comment.

> IMHO. May be just 'OneExplorerView.active" would be sufficient to provide a user some hints how to run .cfg file ?

I recommend putting `(ctrl+f7)` next to `run cfg` icon (play icon). That way, user will directly notice that the shortcut exists. 

>  IMHO even when there is just a single but not selected .cfg file, some users may still want to run it anyway using shortcut, because there is just a single .cfg file. 

You mean the case of there is only one cfg file in the workspace and user hit the shortcut? I think the case could be supported but I think we can discuss about this in another pr or issue.

stamalakhov: Ok. I'm working on it.

jyoungyun: @dayo09 

> > IMHO. May be just 'OneExplorerView.active" would be sufficient to provide a user some hints how to run .cfg file ?
> 
> I recommend putting `(ctrl+f7)` next to `run cfg` icon (play icon). That way, user will directly notice that the shortcut exists. 

Apart from this issue, 
in ONE explorer, it uses ▶️ icon to run cfg. If the `run tvn` feature is introduced afterwards, the meaning of `run` may be a little confusing. If there is a good icon to represent compilation, it would be nice to use this icon for `run cfg` command. :)

dayo09: @jyoungyun How do you expect `run tvn` would be like? Sorry that I don't have any idea about that. I thought that it would be about cfg, not `run cfg`. I would appreciate your explanation, maybe in another issue.

package.json
dayo09: Thanks, it works well :-D

src/OneExplorer/OneExplorer.ts
```diff
@@ -671,13 +701,32 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
       this._onDidChangeTreeData.fire(undefined);
     } else {
       this._onDidChangeTreeData.fire(node);
     }
   }
 
+  /**
+   * Refresh hasSelectedCfg parameter
+   */
+  refreshSelection(): void {
```
dayo09, line 716: How about 'refreshCfgSelection()' for the name? By the name, reader might misunderstand that the function refreshes general selection of nodes.

package.json
stamalakhov: My pleasure to be useful )

src/OneExplorer/OneExplorer.ts
stamalakhov: Ahhh... Ok. I see.

dayo09: Thanks!

dayo09:
### GIF

0203-shortcut

(When I hit ctrl+f7 with circle file, it did not respond as expected.)

Thanks :-D

dayo09 (approved):

stamalakhov:
> (When I hit ctrl+f7 on the circle file, it did not respond as expected.)

@dayo09 
Sorry. But IMHO user clicked .circle file and seems like user intends to work with .circle file (m.b. edit ... ), so it doesn't respond to 'ctrl+f7'. May i ask what are the expectations in this case?

dayo09:
Oh it worked well without failure.

I just wanted to mention that it worked along the expectation, because it was not seen in the gif.

Thanks for your work!

stamalakhov:
> Oh it worked well without failure.
> 
> I just wanted to mention that it worked well along the expectation- by not responding with circle file. I commened it just because it was not seen in the gif.
> 
> Thanks for your work!

My pleasure to be helpful!

jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

Issue #1487, opened 2023-01-31
[CfgEditor] Implement default values


stamalakhov:
### What?

Let's implement  `Automatically fill default options` from https://github.com/Samsung/ONE-vscode/issues/1482#issuecomment-1404607640

### Why?

Currently setting  all values manually is laborious, so it seems that setting them to default can save user efforts.

stamalakhov:
Seems like `?` isn't allowed for a shortcut (see https://code.visualstudio.com/docs/getstarted/keybindings#_accepted-keys).
![image](https://user-images.githubusercontent.com/112689352/217243029-a56e88fa-c1dc-4893-b597-dce11593b324.png).
So let it be `Ctrl+d` for now.

stamalakhov:
Seems like we can close this one because #1493 has been merged.

PR #1493, opened 2023-02-08
[CfgEditor] Fill .cfg parameters  by default values

stamalakhov:
This commit fills .cfg paramters by default values using keyboard shortcut.

This commit fills default values in .cfg file (input and output paths for import/optimization/quantization).
Currently its optimal usage would be:
1) create new configuration
2) check needful steps 
3) press 'ctrl+shift +/' for filling .cfg parameters to default values. ('ctrl+d' is busy)

Issue: #1487

### DIFF
--- a/media/CfgEditor/index.js
+++ b/media/CfgEditor/index.js
@@ -58,6 +58,9 @@ function main() {
       case "displayCfgToEditor":
         displayCfgToEditor(message.text);
         break;
+      case "setDefaultValues":
+        setDefaultValues(message.name);
+        break;
       case "applyDialogPath":
         document.getElementById(message.elemID).value = message.path;
         switch (message.step) {
@@ -99,6 +102,55 @@ function main() {
   });
 
   postMessageToVsCode({ type: "requestDisplayCfg" });
+}
+
+function setDefaultValues(name) {
+  // import step
+  let importedName = name;
+  let importedExt = name + ".circle";
+  const importTypeInfo = {
+    pb: ["PBOutputPath", updateImportPB],
+    saved: ["SAVEDOutputPath", updateImportSAVED],
+    keras: ["KERASOutputPath", updateImportKERAS],
+    tflite: ["TFLITEOutputPath", updateImportTFLITE],
+    onnx: ["ONNXOutputPath", updateImportONNX],
+  };
+  const curInputType = document.getElementById("importInputModelType").value;
+  if (curInputType in importTypeInfo) {
+    const fieldId = importTypeInfo[curInputType][0];
+    const updateFunc = importTypeInfo[curInputType][1];
+
+    document.getElementById(fieldId).value = importedExt;
+    updateFunc();
+  } else {
+    return;
+  }
+
+  // optimization step parameters
+  let optimizedName = importedName + ".opt";
+  let optimizedExt = optimizedName + ".circle";
+  document.getElementById("optimizeInputPath").value = importedExt;
+  document.getElementById("optimizeOutputPath").value = optimizedExt;
+  updateOptimize();
+
+  // quantization step parameters
+  const qType = document.getElementById("DefaultQuantQuantizedDtype").value;
+  const qSuffix = qType === "uint8" ? ".q8" : ".q16";
+  const optimizationUsed = document.getElementById("checkboxOptimize").checked;
+  let quantizedName = optimizationUsed
+    ? optimizedName + qSuffix
+    : importedName + qSuffix;
+  let quantizedExt = quantizedName + ".circle";
+  // input
+  document.getElementById("DefaultQuantInputPath").value = optimizationUsed
+    ? optimizedExt
+    : importedExt;
+  // output
+  document.getElementById("DefaultQuantOutputPath").value = quantizedExt;
+  updateQuantizeDefault();
+
+  // apply
+  applyUpdates();
 }
 
 function registerSteps() {
--- a/package.json
+++ b/package.json
@@ -251,6 +251,11 @@
         "command": "one.viewer.metadata.showFromOneExplorer",
         "title": "ONE: Show Metadata",
         "category": "ONE"
+      },
+      {
+        "command": "one.editor.cfg.setDefaultValues",
+        "title": "ONE: Set Default Values",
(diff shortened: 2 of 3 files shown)


media/CfgEditor/index.js
```diff
@@ -101,6 +104,82 @@ function main() {
...
 }
 
+function setDefaultValues(name) {
+  const optimizationUsed = document.getElementById("checkboxOptimize").checked;
+
+  let importedPath = name;
+  let optimizedPath = importedPath + ".opt";
+  let quantizedPath = optimizationUsed
+    ? optimizedPath + ".Q"
+    : importedPath + ".Q";
```
seanshpark, line 114: using capital letters in filename makes typing annoying.
how about using `q8` for `uint8` and `q16` for `int16` from `DefaultQuantQuantizedDtype` ?
this convention is used for our internal release.

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -223,11 +240,19 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
             "one.explorer.revealInOneExplorer",
             document.fileName
           );
         }
+        if (
+          !webviewPanel.visible &&
+          this._activeWebviewPanel === webviewPanel
+        ) {
+          this._activeWebviewPanel = undefined;
+        }
```
seanshpark, line 255: Q) can you explain why this block is needed and what it does?

media/CfgEditor/index.js
stamalakhov: Ok. I'll try!

src/CfgEditor/CfgEditorPanel.ts
stamalakhov: > Q) can you explain why this block is needed and what it does?


It makes `_activeWebviewPanel` inactive in case it is invisible. If the user selects another view in oneexplorer, e. g. `CircleEditor' or just edits .cfg file in text mode, it's inactive. But   i may try to use `when' clause instead, like the following
```
 "when": "CfgEditor.active"
```
in `package.json` file

seanshpark: > i may try to use when' clause instead, like the following ...

IIUC, `ctrl+D` won't do anything when this View is hidden.
If it works, this block can be removed in this PR?

stamalakhov: > IIUC, `ctrl+D` won't do anything when this View is hidden.

Ahh... Ok. I see. Thank you.

> If it works, this block can be removed in this PR?

Yes. It will be removed.

seanshpark (dismissed):
LGTM thank you!

package.json
```diff
@@ -251,13 +251,23 @@
...
     "keybindings": [
       {
           "command": "one.explorer.runSingleSelectedCfg",
           "key": "ctrl+f7",
           "when": "OneExplorerView.active && one.explorer:hasSelectedCfg"
+      },
+      {
+        "command": "one.cfgEditor.setDefaultValues",
+        "key": "ctrl+shift+/",
+        "when": "one.CfgEditor:activeCfgIsVisible"
```
dayo09, line 270:
```suggestion
        "when": "activeCustomEditorId == one.editor.cfg"
```

You can use this context, it's given by vscode. :-D

src/CfgEditor/CfgEditorPanel.ts
```diff
@@ -246,6 +283,18 @@ export class CfgEditorPanel implements vscode.CustomTextEditorProvider {
...
+      if (this._activeWebviewPanel === webviewPanel) {
+        // there are race conditions with visibility
+        // another webviewPanel becomes visible while this one is disposed
+        // so 'activeCfgIsVisible' will not be set to false
+        vscode.commands.executeCommand(
+          "setContext",
+          "one.CfgEditor:activeCfgIsVisible",
+          false
+        );
+      }
```
dayo09, line 296: By using the given context below, you don't need to be bothered to implement this custom context.

> "when": "activeCustomEditorId == one.editor.cfg"

dayo09 (changes requested):

media/CfgEditor/index.js
```diff
@@ -101,6 +104,61 @@ function main() {
   postMessageToVsCode({ type: "requestDisplayCfg" });
 }
 
+function setDefaultValues(name) {
+  const optimizationUsed = document.getElementById("checkboxOptimize").checked;
```
dayo09, line 108: (Not for functionality, but for readability)

How about collecting related lines?

These optimization-related lines can be moved below, near to the updating lines,
```
const optimizationUsed = document.getElementById("checkboxOptimize").checked;
let optimizedPath = importedPath + ".opt";
```

and these quantization-related lines too.
```
  const qType = document.getElementById("DefaultQuantQuantizedDtype").value;
  const qSuffix = qType === "uint8" ? ".q8" : ".q16";
  let quantizedPath = optimizationUsed
    ? optimizedPath + qSuffix
    : importedPath + qSuffix;
```

src/CfgEditor/CfgEditorPanel.ts
stamalakhov: Ahh... Ok. Thank you. I'll try!

package.json
```diff
@@ -251,13 +251,23 @@
         "command": "one.viewer.metadata.showFromOneExplorer",
         "title": "ONE: Show Metadata",
         "category": "ONE"
+      },
+      {
+        "command": "one.cfgEditor.setDefaultValues",
```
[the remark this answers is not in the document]
stamalakhov: Ok. Thank you!

package.json
stamalakhov: Ok. Thank you.

media/CfgEditor/index.js
stamalakhov: Thank you. I'll try to make it more clear.

dayo09: It looks clear, thanks!

dayo09 (approved):
Thanks a lot! 😄

seanshpark (approved):
LGTM

Issue #1496, opened 2023-02-09
[CircleGraph] Undo reformat for imported codes by prettier


seanshpark:
As format change is now `prettier`, `clang-format on/off` directives are not working and the codes hard to read the codes.

Find a way to `prettier` to ignore and rollback the changes.

seanshpark:
We can use https://prettier.io/docs/en/ignore.html

seanshpark:
done

PR #1498, opened 2023-02-09
[CircleGraph] Use prettier ignore view.js

seanshpark:
This will let prettier ignore previous code of view.js without formatting for better readability.


seanshpark:
for #1496, from  #1497

jyoungyun (approved):
LGTM

PR #1499, opened 2023-02-09
[CircleGraph] Use prettier ignore view-sidebar.js

seanshpark:
This will let prettier ignore previous code of view-sidebar.js without formatting for better readability.


jyoungyun (approved):
LGTM

PR #1505, opened 2023-02-16
 [MPQEditor] Create or edit mixed precision quantization json file.

stamalakhov:
This commit allows to create or edit mixed precision quantization json file.

ONE-vscode-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #1491, opened 2023-02-07
Introduce manual mixed precision quantization 


stamalakhov:
### What?
Let's implement manual mixed precision feature via editing quantization json file similar to `PartEditor`.

### Why?
Currently setting all values via .json is laborious, so it seems like setting them in GUI will be more convenient.

stamalakhov:
Please correct me if I'm wrong, but
IMHO
the best way to create/edit .json suitable for mixed precision quantization is to create context_menu_item (e.g. `create mixed precision quantization json`) for float `circle` model or (`edit mixed precision quantization json`) for json file.

seanshpark:
> the best way to create/edit .json suitable for mixed precision ...

As I didn't think much on this feature but I can assume you had more on this.
Please share your idea with writing some draft so that others including me can try and think :)

One thing is that one GUI can be nice but can harm harmony with others.

stamalakhov:
> > the best way to create/edit .json suitable for mixed precision ...
> 
> As I didn't think much on this feature but I can assume you had more on this. Please share your idea with writing some draft so that others including me can try and think :)
> 
> One thing is that one GUI can be nice but can harm harmony with others.
> 
> Add: once we have the feature, we can later adjust the entry interface.

Ok. Thank you. I'll be back with a draft.

stamalakhov:
All changes related to MPQ are merged, so it seems that we can close this one.

PR #1509, opened 2023-03-13
[MPQEditor] Introduce 'media' for MPQEditor

stamalakhov:
This commit introduces 'media'  for MPQEditor.

Draft: #1505
Related: #1491

### DIFF
--- a/media/MPQEditor/index.html
+++ b/media/MPQEditor/index.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en">
+<!-- 
+Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ 
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+ 
+   http://www.apache.org/licenses/LICENSE-2.0
+ 
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<head>
+    <meta charset="utf-8">
+    <meta
+        http-equiv="Content-Security-Policy"
+        content="img-src vscode;
+                style-src 'unsafe-inline' ${webview.cspSource};
+                script-src 'nonce-${nonce}';">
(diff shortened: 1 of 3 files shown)


media/MPQEditor/index.html
```diff
@@ -0,0 +1,75 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Mixed Precision Quantization editor</title>
+    <link rel="stylesheet" href="${cssUri}">
+    <link rel="stylesheet" href="${codiconUri}">
+    <script nonce="${nonce}" type="module" src="${toolkitUri}"></script>
+    <script nonce="${nonce}" type="module" src="${scriptUri}"></script>
+</head>
+<body>
+    <div class="maintitle">
+        <span>Edit mixed precision quantization json file</span>
```
dayo09, line 34: I personally think this title is not redundant. How about removing this?

stamalakhov: ok.

stamalakhov: done

dayo09:
Thanks for your work! 😄 

Inked62e0e08f-a495-4a0c-85a4-fc13f9382fa4
BTW, when I click on that 'X' icon in 'visq path' box, it only reloads the model.circle file on the right side. What does it mean to be?

stamalakhov:
> Thanks for your work! smile
> 
> Inked62e0e08f-a495-4a0c-85a4-fc13f9382fa4 BTW, when I click on that 'X' icon in 'visq path' box, it only reloads the model.circle file on the right side. What does it mean to be?

It clears `visq_view`.
By default, the `model-view` opens in `selector` mode, but in case `visq` data is loaded, `model_view` load the model in `visq` mode. So to reset to `selector` mode, we may press the `x` button. But IMHO in case we don't `visq` data, it should be disabled.

dayo09:
@stamalakhov 

Oh, I think that you have removed the code. 
So it's now unrelated but to say, I would like to suggest 'Show Graph' or 'Display Graph' instead of 'Model Graph'?

stamalakhov:
> @stamalakhov
> 
> > But IMHO in case we don't visq data, it should be disabled.
> 
> Oh, It seems that you just have removed the code. So it's now unrelated but to say, but for later, I would like to suggest 'Show Graph' or 'Display Graph' instead of 'Model Graph', for the text next to the checkbox.

Ahhh. OK. I'll rename it.

dayo09 (approved):
LGTM, Thanks.

seanshpark (approved):
LGTM

PR #1511, opened 2023-03-27
[DRAFT][MPQEditor] Introduce MPQEditor

stamalakhov:
This draft introduces more fresh version of `MPQEditor` for testing with tests and screenrecordings of its capabilities.

Naming and validation of `mpq.json`:

https://user-images.githubusercontent.com/112689352/229081677-e6ca1ade-00da-4450-b81e-6c424c8456ab.mp4

Changing quantization parameters and saving:

https://user-images.githubusercontent.com/112689352/228561235-cc420b94-f249-4ab6-9506-e6269b822c43.mp4

Adding specific layers and update of actual default layers:

https://user-images.githubusercontent.com/112689352/228562694-cef536d7-e566-4511-a196-cd2e26defeb6.mp4

### DIFF
--- a/src/MPQEditor/MPQEditor.ts
+++ b/src/MPQEditor/MPQEditor.ts
@@ -47,6 +47,12 @@ export class MPQEditorProvider implements vscode.CustomTextEditorProvider {
         }
       ),
       // Add command registration here
+      vscode.commands.registerCommand(
+        "one.editor.mpq.showFromDefaultExplorer",
+        (uri) => {
+          MPQEditorProvider.createMPQJson(uri);
+        }
+      ),
     ];
 
     registrations.forEach((disposable) =>
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -30,6 +30,7 @@ import { PartGraphSelPanel } from "./PartEditor/PartGraphSelector";
 import { ToolchainProvider } from "./Toolchain/ToolchainProvider";
 import { Logger } from "./Utils/Logger";
 import { VisqViewerProvider } from "./Visquv/VisqViewer";
+import { MPQEditorProvider } from "./MPQEditor/MPQEditor";
 
 /* istanbul ignore next */
 export function activate(context: vscode.ExtensionContext) {
@@ -76,6 +77,8 @@ export function activate(context: vscode.ExtensionContext) {
   VisqViewerProvider.register(context);
 
   MetadataViewerProvider.register(context);
+
+  MPQEditorProvider.register(context);
 
   // returning backend registration function that will be called by backend extensions
   return backendRegistrationApi();
--- a/package.json
+++ b/package.json
@@ -128,6 +128,16 @@
           }
         ],
         "priority": "default"
+      },
+      {
+        "viewType": "one.editor.mpq",
+        "displayName": "MPQ Editor",
+        "selector": [
+          {
+            "filenamePattern": "*.mpq.json"
+          }
+        ],
+        "priority": "default"
       }
     ],
     "commands": [
@@ -256,6 +266,11 @@
         "command": "one.editor.cfg.setDefaultValues",
         "title": "ONE: Set Default Values",
         "category": "ONE"
+      },
+      {
+        "command": "one.editor.mpq.showFromDefaultExplorer",
+        "title": "Create MPQ json",
+        "category": "ONE"
       }
     ],
     "keybindings": [
@@ -370,6 +385,11 @@
           "command": "one.viewer.metadata.showFromDefaultExplorer",
           "when": "resourceExtname in one.metadata.supportedFiles && !explorerResourceIsFolder",
           "group": "7_metadata@1"
+        },
+        {
+          "command": "one.editor.mpq.showFromDefaultExplorer",
+          "when": "resourceExtname == .circle",
+          "group": "1_tools"
         }
       ],
       "commandPalette": [
@@ -451,6 +471,10 @@
         },
         {
           "command": "one.viewer.metadata.showFromOneExplorer",
+          "when": "false"
+        },
+        {
+          "command": "one.editor.mpq.showFromDefaultExplorer",
           "when": "false"
         }
       ]


src/Tests/MPQEditor/MPQEditor.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      const dirPath: string = testBuilder.dirInTemp;
+
+      const content = `
+            empty content
+            `;
+
+      let targetNames: string[] = [
+        "model.mpq.json",
+        "model(1).mpq.json",
+        "model(1)(1).mpq.json",
```
dayo09, line 65: I wonder why you have decided to generate model name as such?
I think it's rather natural that when model.mpq.json and model(1).mpq.json exists, model(2).mpq.json be the next..

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,525 @@
...
+
+  /**
+   * Add layers to edit their quantization parameters
+   */
+  private hadleAddSpecificLayerFromDialog(document: vscode.TextDocument) {
+    const nodes =
+      this._mpqDataMap[document.uri.toString()].getDefaultModelLayers();
+    const pickOptions = {
+      canPickMany: true,
+    };
```
dayo09, line 414: How about applying those quickPickOptions? 😄 

```suggestion
    const pickOptions = {
      title: "Add specific layers to edit their quantization parameters",
      matchOnDescription: true,
      matchOnDetail: true,
      placeHolder: "Select layers",
      ignoreFocusOut: true,
      canPickMany: true,
    };
```

dayo09: 190e428b-2d12-4309-a964-d204e1629614

src/Tests/MPQEditor/MPQEditor.test.ts
stamalakhov: @dayo09 
Thank you.
Please see https://github.com/Samsung/ONE-vscode/pull/1505#issuecomment-1437715625.
Just adding `(1)` is easy to implement, but i can change it to another scheme if needed.

src/MPQEditor/MPQEditor.ts
dayo09: it will
- title: show a title "Add specific~"
- placeHolder: show a place holder "Select layers"
- ignoreFocusOut: stay even if you clicks other sides
- matchOnDescription/Detail: (well it won't affect cuz we don't have description/details in the list elements)

stamalakhov: Ahhh. Thank you very much!

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,530 @@
...
+        dirPath,
+        baseMPQName + MPQEditorProvider.fileExtension
+      );
+      if (!fs.existsSync(mpqPath)) {
+        return baseMPQName + MPQEditorProvider.fileExtension;
+      }
+      baseMPQName = baseMPQName + "(1)";
+    }
+    return "";
+  }
```
dayo09, line 78: This function will not work as expected when the trial repeats more than 5 times.
It will return the "", which is not expected.

On top of that, I think it's natural to have increasing numbers (1)->(2)->...
Could you revise this function?

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,530 @@
...
+  ): Promise<vscode.Uri | undefined> {
+    const content =
+      '{"default_quantization_dtype": "uint8",' +
+      '"default_granularity": "channel",' +
+      '"layers": [],' +
+      '"model_path": ' +
+      '"' +
+      circleName +
+      '"' +
+      "}";
```
dayo09, line 93:
```suggestion
      `{"default_quantization_dtype": "uint8",
      "default_granularity": "channel",
      "layers": [],
      "model_path": "${circleName}"`;
```

You can use this \`\` string format :-D

src/MPQEditor/MPQEditor.ts
stamalakhov: Yes. I can. Thank you.

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,530 @@
...
+   */
+  public static createMPQJson(uri: vscode.Uri): void {
+    const dirPath = path.parse(uri.path).dir;
+    const modelName = path.parse(uri.path).name;
+    const extName = path.parse(uri.path).ext.slice(1);
+    const circleName = path.parse(uri.path).base;
+
+    // try to guess unoccupied name for mpq.json
+    let mpqName = MPQEditorProvider.findMPQName(modelName, dirPath);
+    if (mpqName.length < 1) {
```
dayo09, line 142:
```suggestion
    if (mpqName.length == 0) {
```

I think it's more explicit to test an empty string.

src/MPQEditor/MPQEditor.ts
stamalakhov: ok. Thank you.

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,526 @@
...
+   */
+  public static createMPQJson(uri: vscode.Uri): void {
+    const dirPath = path.parse(uri.path).dir;
+    const modelName = path.parse(uri.path).name;
+    const extName = path.parse(uri.path).ext.slice(1);
+    const circleName = path.parse(uri.path).base;
+
+    // try to guess unoccupied name for mpq.json
+    let mpqName = MPQEditorProvider.findMPQName(modelName, dirPath);
+    if (mpqName.length == 0) {
```
dayo09, line 138:
```suggestion
    if (mpqName.length === 0) {
```
Oops sorry.

src/MPQEditor/MPQEditor.ts
stamalakhov: @dayo09 
I've implemented (and updated tests and draft description) simple version of it , in the form of
```
for (let i =0; ; i++)
```
do i need to constrain somehow index (by 10000 e.g.) ?
Or it was assumed to some form `glob` query , e.g.
```
1. find the longest filename with the pattern *(*).mpq.json
2. extract number from round brackets
3. return number + 1 as possible answer
```
?

src/extension.ts
```diff
@@ -77,6 +78,13 @@ export function activate(context: vscode.ExtensionContext) {
 
   MetadataViewerProvider.register(context);
 
+  MPQEditorProvider.register(context);
+
+  Logger.info("activated");
+
+  // Set context as a global
+  (global as any).extensionContext = context;
```
dayo09, line 86: Q. I wonder why is this needed?

src/MPQEditor/MPQEditor.ts
```diff
@@ -0,0 +1,528 @@
...
+  /**
+   * find unoccupied mpq file-name
+   */
+  public static findMPQName(baseMPQName: string, dirPath: string): string {
+    if (baseMPQName.length === 0) {
+      throw new Error("Invalid mixed precision quantization file name");
+    }
+    const baseName = baseMPQName;
+    let mpqName: string = "";
+    for (let i = 0; ; i++) {
```
dayo09, line 71: How about putting a large number as a limit for safety?

stamalakhov: > how about putting a large number as a limit for safety in the for loop?

 @dayo09
Thank you. Yes. It seems like an option, but which one? 100? 1000? ...

src/extension.ts
stamalakhov: > Q. I wonder why is this needed?

AFAIK no. It's needed e.g. in tests to construct `MPQEditorProvider`, but it was used `constructor` only. So it can be omitted for now, because `constructor` is trivial.

src/MPQEditor/MPQEditor.ts
stamalakhov: @dayo09 
Other option: 
we may query `fs` how many `mpq.json` files are there (`max_num`), and set maximal number of iterations as `max_num+1`.

src/MPQEditor/MPQEditor.ts
```diff
@@ -41,13 +47,148 @@ export class MPQEditorProvider implements vscode.CustomTextEditorProvider {
...
+      .showInputBox({
+        title: `Create mixed precision quantization configuration for '${modelName}.${extName}' :`,
+        placeHolder: `Enter a file name`,
+        value: mpqName,
+        valueSelection: [
+          0,
+          mpqName.length - `${MPQEditorProvider.fileExtension}`.length,
+        ],
+        validateInput: (mpqName: string): string | undefined => {
+          return MPQEditorProvider.validateMPQName(dirPath, mpqName);
```
stamalakhov, line 163: This code uses `validateMPQName`

stamalakhov:
Seems like the whole draft is merged for now, so we can close it.

PR #1530, opened 2023-04-19
[MPQEditor] Introduce 'Show graph'

stamalakhov:
This draft introduces the ability to edit specific layers in `circle-view`.

'Show graph` synchronization test:

https://user-images.githubusercontent.com/112689352/233072896-f51418be-6a91-4fb0-b350-9162bf4c0a29.mp4

'Add/Remove' synchronization test:

https://user-images.githubusercontent.com/112689352/233073477-f41f0d4c-f0e2-4162-b521-466db6564a71.mp4

'Add' with changed defaults test:

https://user-images.githubusercontent.com/112689352/233074181-ce93c481-8552-45bc-9692-9ceef7b208d9.mp4

Full draft: #1511
ONE-vscode-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #1531, opened 2023-04-19
[MPQEditor] Implement media for 'Show graph'

stamalakhov:
This commit implements media changes for 'Show graph'.

Its correctness is tested in #1530

Fresh draft: #1530
Full draft: #1505
Related: https://github.com/Samsung/ONE-vscode/issues/1491

### DIFF
--- a/media/MPQEditor/index.js
+++ b/media/MPQEditor/index.js
@@ -33,6 +33,9 @@ function main() {
       case "modelNodesChanged":
         handleModelNodesChanged(message.names);
         break;
+      case "modelGraphIsShown":
+        handleModelGraphIsShown(message.shown);
+        break;
       default:
         break;
     }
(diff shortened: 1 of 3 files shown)


media/MPQEditor/index.js
```diff
@@ -33,19 +33,27 @@ function main() {
...
+        handleModelGraphIsShown(message.shown);
+        break;
       default:
         break;
     }
   });
 
   vscode.postMessage({ type: "requestDisplayMPQ" });
   vscode.postMessage({ type: "requestModelNodes" });
+  vscode.postMessage({ type: "showModelNodes" });
```
seanshpark, line 46: Q) this will post in serial at once and the response may not guarantee in this order. Is this OK?

seanshpark (approved):
anyway LGTM

media/MPQEditor/index.js
stamalakhov: > Q) this will post in serial at once and the response may not guarantee in this order. Is this OK?

Yes. It is ok. They are independent.
`showModelNodes` just opens another panel for selection of layers to edit.

dayo09 (approved):
LGTM :-D

PR #1539, opened 2023-04-27
[MPQEditor] Implement 'dispose' and 'onViewMessage'

stamalakhov:
This commit implements 'dispose' and 'onViewMessage' for manual mixed precision quantization task.

Its correctness is tested in https://github.com/Samsung/ONE-vscode/pull/1530

Fresh draft: https://github.com/Samsung/ONE-vscode/pull/1530
Full draft: https://github.com/Samsung/ONE-vscode/pull/1505
Related: https://github.com/Samsung/ONE-vscode/issues/1491

### DIFF
--- a/src/MPQEditor/MPQCircleSelector.ts
+++ b/src/MPQEditor/MPQCircleSelector.ts
@@ -20,6 +20,7 @@ import * as vscode from "vscode";
 import {
   CircleGraphCtrl,
   CircleGraphEvent,
+  MessageDefs,
 } from "../CircleGraph/CircleGraphCtrl";
 
 export interface MPQSelectionEvent {
@@ -213,14 +214,49 @@ export class MPQSelectionPanel
   }
 
   public dispose() {
-    //TODO
+    if (!this._closedByOwner && this._mpqEventHandler) {
+      this._mpqEventHandler.onClosed(this._ownerPanel);
+    }
+
+    this.disposeGraphCtrl();
+
+    MPQSelectionPanel.panels.forEach((selPan, index) => {
+      if (
+        this._ownerId === selPan._ownerId &&
+        this._modelPath === selPan._modelPath
+      ) {
+        MPQSelectionPanel.panels.splice(index, 1);
+      }
+    });
+
+    // Clean up our resources
+    this._panel.dispose();
+
+    while (this._disposables.length) {
+      const x = this._disposables.pop();
+      if (x) {
+        x.dispose();
+      }
+    }
   }
 
   /**
    * CircleGraphEvent interface implementations
    */
-  public onViewMessage(_message: any) {
-    // TODO
+  public onViewMessage(message: any) {
+    switch (message.command) {
+      case MessageDefs.selection:
+        this._lastSelected = message.names;
+        // we need to update the document, but not save to file.
+        // pass to owner to handle this.
+        if (this._mpqEventHandler) {
+          this._mpqEventHandler.onSelection(message.names, this._document);
+        }
+        break;
+      case MessageDefs.finishload:
+        this.onForwardSelection(this._lastSelected);
+        break;
+    }
   }
 
   public onForwardSelection(selection: any) {


src/MPQEditor/MPQCircleSelector.ts
```diff
@@ -213,14 +214,49 @@ export class MPQSelectionPanel
...
   }
 
   /**
    * CircleGraphEvent interface implementations
    */
-  public onViewMessage(_message: any) {
-    // TODO
+  public onViewMessage(message: any) {
+    switch (message.command) {
+      case MessageDefs.selection:
```
dayo09, line 248: I've got some UI feedbacks. If it's not easy to handle on this pr, you may consider it later.

image

On #1505 draft, 
- when I click on 'add' node, MPQEditor seems to adds the node to the mpq editor slot. 
- when I click on 'Place holder's or 'Add 2' node, as they are not the `mpq` target, they shows the property view on the right side of the circle editor.

I think that 'click' event must result in the same behavior. 
We can choose either to open the property view same to 'add' operation or not to open the property view to other operations.

dayo09:
Hi, this is another UI feedback.
It's not related to this PR, but rather related to the draft #1505.

image

It seems that : In the (1) area, it decides the default quantization configuration and (2) area specifies the per-node(layer) quantization configuration.

At first look, I didn't understand what the (1) area would do. I propose to add 'Set default quantization' 'Set specific layer's quantization' in each areas.

src/MPQEditor/MPQCircleSelector.ts
stamalakhov: @dayo09 
Thank you. I'll try to fix it in the draft at first.

stamalakhov:
> Hi, this is another UI feedback. It's not related to this PR, but rather related to the draft #1505.
> 
> image
> 
> It seems that : In the (1) area, it decides the default quantization configuration and (2) area specifies the per-node(layer) quantization configuration.
> 
> At first look, I didn't understand what the (1) area would do. From my point of view, I propose to add 'Set default quantization' 'Set specific layer's quantization' in each areas.

@dayo09 
Thank you. I'll try to fix it in the draft at first.

src/MPQEditor/MPQCircleSelector.ts
stamalakhov: @dayo09 @seanshpark 
The same behaviour in the netron:
 click on input/output nodes shows `Model Properties`, while clicking on the internal nodes show `Nodes Properties`.
For MPQ draft clicking internal nodes results in their selection for editing, while for input/output nodes still shows `Model Properties` to observe input/output size and theirs type.
To handle this issue:
1. We can hide input/output nodes
2. We can disable clicking input/output nodes

But still there may be some issues that some users are used to observe input/output size.
IMHO.

stamalakhov:
> It seems that : In the (1) area, it decides the default quantization configuration and (2) area specifies the per-node(layer) quantization configuration.

@dayo09 
To fix layout it needs to edit media files, so may be we can handle it in another PR?

dayo09:
@stamalakhov  

> To fix layout it needs to edit media files, so may be we can handle it in another PR?

Sure :-D Thanks a lot :-D

src/MPQEditor/MPQCircleSelector.ts
dayo09: Or 3. We can show `Properties` too, when clicking on the internal nodes(`ex. Add`).

dayo09: I will approve this PR, let's go on about it on the other issue or PR. Thanks for listening to me.!

dayo09 (approved):

src/MPQEditor/MPQCircleSelector.ts
stamalakhov: > I will approve this PR, let's go on about it on the other issue or PR. Thanks for listening to me.!

Thank you very much!

seanshpark (approved):
LGTM

Issue #1541, opened 2023-04-28
[MPQEditor] Make layout more user-friendly


stamalakhov:
### What?
Let's make MPQEditor layout more user- friendly.

### Why?
See original discussion in https://github.com/Samsung/ONE-vscode/pull/1539#issuecomment-1526872463.

stamalakhov:
@dayo09 @seanshpark 
Initial draft is proposed at #1542.
Otherwise we can try to make both sections collapsible as was done at `metadataviewer `.

dayo09:
Oh, I just got another idea - how about adding 'default' just as the same as the other layers? Not as a radio button, but just as a drop box element just the same as other layers. Therefore we don't need to put some text information.

dayo09:
I just got another idea - how about adding 'default' just as the same as the other layers? Not as a radio button, but just as a drop box element just the same as other layers. Therefore we don't need to put some text information.

stamalakhov:
> I just got another idea - how about adding 'default' just as the same as the other layers? Not as a radio button, but just as a drop box element just the same as other layers. Therefore we don't need to put some text information.

@dayo09 
Like this?:
![235097735-1036982d-4a60-4b7a-97f9-ff75ae4b250a](https://user-images.githubusercontent.com/112689352/235584014-afe87375-2090-4e19-a176-5049d0d7ee6c.png)

or ?:
![235097735-1036982d-4a60-4b7a-97f9-ff75ae4b250a (1)](https://user-images.githubusercontent.com/112689352/235584779-74f1a026-3fec-42ff-9da9-5ebf519f202f.png)

dayo09:
I meant 1.) 😄

stamalakhov:
> I meant 1.) 😄

Ahh. Ok. Thank you!

PR #1548, opened 2023-05-02
[MPQEditor] Improve layout

stamalakhov:
This commit improves UI layout.
This commit improves UI layout.
The draft produces the following UI.

image

Related: https://github.com/Samsung/ONE-vscode/issues/1541

### DIFF
--- a/media/MPQEditor/index.html
+++ b/media/MPQEditor/index.html
@@ -32,20 +32,33 @@ limitations under the License.
 <body>
     <div>
         <div>
-            <div>
-                <vscode-radio-group id="DefaultDtype">
-                    <label slot="label">Quantized Data Type</label>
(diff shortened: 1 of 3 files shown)


stamalakhov:
Also we could place all model nodes AS IS in the table for editing, without default/specific separation.
I mean:
1. we could populate the table with all models nodes
2. there are a lot of nodes, so in case we have visq data, we can add additional column with `accuracy relevance` and then sort all nodes by `accuracy relevance` so that most erroneous nodes are at the top
3. in case latency data is available (e.g. profiler output), we can insert another column `latency relevance` and then sort all nodes by 'latency relevance` to have the ability for fast elimination of slowest nodes as well as the most erroneous nodes
4. at the bottom of the table we can make summary row, depicting expected 'latency' and `accuracy` improvements in case MPQ configuration will be applied for quantization

Issue #1562, opened 2023-05-23
Support model related functions


jyoungyun:
## What

Let's support model related functions such as inference, profiling and getting model information.

## Todo

- [x] Expand Backend Class
- [x] Introduce stdout result variable in JobRunner
- [x] Implement inference function
- [x] Implement profiling function
- [x] Implement function to get model information

Draft: #1552

jyoungyun:
Done!

PR #1563, opened 2023-05-23
Update Backend codes for supporting model related functions

jyoungyun:
This commit updates Backend codes for support model related functions.

Related issue: #1562

### DIFF
--- a/src/Backend/ToolchainImpl/DebianToolchain.ts
+++ b/src/Backend/ToolchainImpl/DebianToolchain.ts
@@ -14,10 +14,8 @@
  * limitations under the License.
  */
 
-import assert from "assert";
-
 import { Command } from "../Command";
-import { Toolchain, ToolchainInfo } from "../Toolchain";
+import { IToolchain, IToolCommand, ToolchainInfo } from "../Toolchain";
 
 class DebianRepo {
   uri: string;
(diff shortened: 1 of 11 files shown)


src/Backend/ToolchainImpl/DebianToolchain.ts
```diff
@@ -124,13 +106,32 @@ class DebianToolchain implements Toolchain {
     cmd.push("echo $?");
     return cmd;
   }
+}
+
+class DebianToolchain implements IToolchain<DebianTool> {
+  tool: DebianTool;
```
dayo09, line 112: Q. Why did you put install() or installed() commands inside `tool` not `toolchain`..? I wonder what is the intention of introducing `tool` because it looks good enough that `toolchain` has such functions.

dayo09: I mean it looks more natural that `toolchain` owns the installation-related functions, because it doens't own multiple tools that needs installation seperately.

jyoungyun: This design also has one tool. I just separate the install functions from other functions. I think `Toolchain` means `TRIV Toolchain`. And the way to install and uninstall the `TRIV Toolchain` varies from environment. So I don't think it is right that the two were tied together.

dayo09: Would it be better with install() get env as parameter or so? 
I just wondered you intention and I don't understand still... But it's a minor issue so I won't block this.

dayo09 (approved):

jyoungyun:
PTAL @Samsung/one-vscode

seanshpark (approved):
LGTM

Issue #1566, opened 2023-05-31
Managing SAM scores


lemmaa:
We need a review and resolution of issues related to SAM scores. Please refer to the CC report of the analysis results.

<img src="https://github.com/Samsung/ONE-vscode/assets/1185333/da342501-1fea-48da-8a9b-ec2b6d574383" width=50%></img>

- https://analysishub.sec.samsung.net/service/analyses/2115735

/cc @Samsung/one-vscode

seanshpark:
```
CC;media/CircleGraph/index.js;add;22;785;Y
CC;media/CircleGraph/view-sidebar.js;toggle;27;1077;Y
CC;media/CircleGraph/view-sidebar.js;render;33;1398;Y
CC;media/CircleGraph/view-sidebar.js;formatDocumentation;25;1570;Y
CC;media/CircleGraph/view-sidebar.js;update;31;1843;Y
CC;media/CircleGraph/view-sidebar.js;_format;53;2002;Y
CC;media/CircleGraph/view-sidebar.js;_tokenize;58;2199;Y
CC;media/CircleGraph/view-sidebar.js;_tokenizeInline;69;2476;Y
CC;media/CircleGraph/view-sidebar.js;_render;39;2805;Y
CC;media/CircleGraph/view-sidebar.js;_renderInline;21;2964;Y
CC;media/CircleGraph/view.js;add;31;1383;Y
CC;media/CircleGraph/view.js;_add;70;1541;Y
CC;media/CircleGraph/view.js;ARROW_FUNCTION_l2823:c27;28;2823;Y
```

@lemmaa , these are imported codes from netron. what is your recommendation for these?

seanshpark:
@lemmaa please share how to read these lines
```
CC;media/CircleGraph/index.js;add;22;785;Y
```

seanshpark:
@lemmaa , there are lots of `media/CircleEditor` which is copy of `media/CircleGraph` with modification from last year SSAFY project. what is your recommendation for this?

lemmaa:
> @lemmaa , these are imported codes from netron. what is your recommendation for these?

I don't think it's appropriate from a maintenance point of view that we directly modify the code of external modules. In this case, you will have to process it as exclude in the analysis target.

However, if you plan to disconnect from an external module, it would be appropriate to modify the source code and resolve the issues.

lemmaa:
[...]

PR #1567, opened 2023-05-31
[ahub/sam] Revise exclude list

seanshpark:
This will revise sam exclude list to exclude CircleGraph and CircleEditor.


seanshpark:
for #1566

jyoungyun (approved):
LGTM

PR #1568, opened 2023-06-01
Introduce stdout result variable in JobRunner

jyoungyun:
This commit introduces the stdout result variable to JobRunner. If the tool runner succeeds in executing the command, the stdout result is passed to the JobRunner.

Related issue: #1562


seanshpark (approved):
LGTM

PR #1570, opened 2023-06-01
[Toolchain] Add checkAvailableToolchain function in ToolchainProvider

jyoungyun:
This commit adds checkAvailableToolchain function. This code comes from _run() function to share this function with other functions.


seanshpark:
change itself looks ok but as there seems multiple changes and diff itself looks complicate,
can you split spit this to (1) introduce `checkAvailableToolchain` and then (2) use `checkAvailableToolchain` ?

PR #1574, opened 2023-06-01
[Toolchain] Implement inference function in ToolchainEnv

jyoungyun:
This commit implements inference function in ToolchainEnv.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

Related issue: #1562

PR #1577, opened 2023-06-04
[Toolchain] Change inference function name

jyoungyun:
This commit changes inference() to infer().

Related issue: #1562

### DIFF
--- a/src/Tests/Toolchain/ToolchainEnv.test.ts
+++ b/src/Tests/Toolchain/ToolchainEnv.test.ts
@@ -283,13 +283,13 @@ suite("Toolchain", function () {
       });
     });
 
-    suite("#inference()", function () {
-      test("NEG: requests unimplemented inference", async function () {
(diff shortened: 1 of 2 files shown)


seanshpark (approved):
LGTM

seanshpark:
@jyoungyun , changing API will break codes outside of this REPO if any.
It would be better to (1) add `infer()` (2) notify users to change codes before YY/MM (3) deprecate `inference()` at YY/MM.

I'll land this if you are sure changing this API is OK.

jyoungyun:
@seanshpark 

> @jyoungyun , changing API will break codes outside of this REPO if any.
> It would be better to (1) add `infer()` (2) notify users to change codes before YY/MM (3) deprecate `inference()` at YY/MM.

Fortunately, there is no use case for this api. I introduced this api last week (#1574). So I think it's better to change it quickly before someone uses it. :) Thank you!!!

PR #1584, opened 2023-06-22
[Draft] Connect inference operation in ONE Explorer

jyoungyun:
This commit connects infererence operation in ONE Explorer.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #1585, opened 2023-06-22
[OneExplorer] Append the extname to the contextValue

jyoungyun:
This commit appends the extname to the contextValue. The OneExplorer does not support `resourceExtname` context key. This issue can be resolved using the extname in the contextValue.

Draft PR: #1584

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -463,7 +463,15 @@ export class OneNode extends vscode.TreeItem {
     //
     // However, resourceExtname returns info of vscode Explorer view (not of OneExplorer).
     //    "when": "view == OneExplorerView && viewItem == config"
-    this.contextValue = node.typeAsString;
+
+    // NOTE
+    // this.contextValue has a format of {NodeType}.{Ext}
+    // EXAMPLE
+    //   directory
+    //   basemode.tflite
+    //   product.tvn
+    const extname = path.extname(node.uri.fsPath);
+    this.contextValue = node.typeAsString + extname;
   }
 }
 
--- a/package.json
+++ b/package.json
@@ -329,22 +329,22 @@
       "view/item/context": [
         {
           "command": "one.explorer.runCfg",
-          "when": "view == OneExplorerView && viewItem == config",
+          "when": "view == OneExplorerView && viewItem =~ /config/",
           "group": "1_run"
         },
         {
           "command": "one.explorer.runCfg",
-          "when": "view == OneExplorerView && viewItem == config",
+          "when": "view == OneExplorerView && viewItem =~ /config/",
           "group": "inline"
         },
         {
           "command": "one.explorer.rename",
-          "when": "view == OneExplorerView && viewItem == config",
+          "when": "view == OneExplorerView && viewItem =~ /config/",
           "group": "7_modification"
         },
         {
           "command": "one.explorer.refactor",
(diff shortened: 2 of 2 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -463,7 +463,10 @@ export class OneNode extends vscode.TreeItem {
     //
     // However, resourceExtname returns info of vscode Explorer view (not of OneExplorer).
     //    "when": "view == OneExplorerView && viewItem == config"
-    this.contextValue = node.typeAsString;
+    //
+    // To resolve above the issue, it appends the extname to the contextValue.
```
dayo09, line 467: The above comment is not the issue but the history. It's to explain why we should use 'viewItem'.

src/OneExplorer/OneExplorer.ts
```diff
@@ -463,7 +463,10 @@ export class OneNode extends vscode.TreeItem {
...
+    const extname = path.extname(node.uri.fsPath);
+    this.contextValue = node.typeAsString + extname;
```
dayo09, line 469: For what use would you need extname in contextValue? ❓

jyoungyun: This code requires to distinguish the tvn file from product files.
I will use the extname like this.
```
        {
          "command": "one.explorer.inferModel",
          "when": "view == OneExplorerView && viewItem =~ /tvn/",
          "group": "inline"
        },
```
from #1854

dayo09: Aha, understood :-D

src/OneExplorer/OneExplorer.ts
dayo09, line 467:
```suggestion
```

dayo09: How about removing the comment?

src/OneExplorer/OneExplorer.ts
dayo09, line 469:
```suggestion
    // NOTE
    // this.contextValue's having a format of {NodeType}.{Ext}
    // EXAMPLE
    //   directory.
    //   basemodel.tflite
    //   product.tvn
    this.contextValue = node.typeAsString + extname;
```

dayo09: But I wonder what 'extname' would be within directory nodes.

jyoungyun: If it is a directory node, there is an empty string in the extname.

jyoungyun:
@dayo09 PTAL again.

src/OneExplorer/OneExplorer.ts
dayo09: How about adding the comment? It makes the format readable at a glance.

jyoungyun: I added the comment on the `extname` line.

dayo09 (approved):

src/OneExplorer/OneExplorer.ts
dayo09: Oh sure. Thanks :-D

dayo09: Oh sure. Thanks :-D

jyoungyun:
@seanshpark PTAL :)

seanshpark (approved):
LGTM

PR #1597, opened 2023-06-29
[OneExplorer] Show profile result from profileModel command

dayo09:
This commit shows profile result file created by profileModel command.

- ONE Explorer basically shows a file written in config.json. If it's not written in config.json, it won't be shown.
- one.toolchain.profileModel command generates <model>.trace.json in the same directory. Let's show it exceptionally.


src/OneExplorer/ConfigObject.ts
```diff
@@ -325,6 +325,24 @@ export class ConfigObj {
       }),
     });
 
+    // NOTE
+    // Shows <model>.trace.json
+    // REQUIRES: <model>.tvn be written in the config file.
```
dayo09, line 330: @jyoungyun The tvn file should be written in the config file to show it. Will it be enough?

jyoungyun: Sure. I'm considering profiling with the tvn file created in ONE-VSCode, so it is satisfied with your constraints.

jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

PR #1603, opened 2023-07-04
[OneExplorer] Delete multiple nodes with shortcut

dayo09:
This commit enables multiple node deletion on shortcut.

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -619,19 +619,16 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
       vscode.commands.registerCommand(
         "one.explorer.deleteOnShortcut",
         async () => {
-          if (provider.getSelectedNodes()?.length !== 1) {
-            // Delete is only supported for single selection
-            // Do not show an error or warning message for UI's sake
-            // TODO: handle for multiple selection
-            return;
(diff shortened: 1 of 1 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -619,19 +619,16 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
-            // Do not show an error or warning message for UI's sake
-            // TODO: handle for multiple selection
-            return;
-          } else {
-            const node = provider.getSelectedNodes()![0];
-            Logger.info("OneExplorer", "Shortcut", `Delete ${node.uri.fsPath}`);
-
-            await provider.delete(node);
-            // TODO: handle for multiple selection
-            provider.refresh(node.parent);
```
jyoungyun, line 633: Is there no need to refresh in case of multiple nodes?

dayo09: Actually the phrase is redundant because the view refreshes on file system deletion.
I put it in case multiple node deletion requires manual refresh. But when I implemented it, the deletion pop-up gave it enough time to self-refresh, so it worked without the phrase.

jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

Issue #1609, opened 2023-07-20
Print token as asterisks


dayo09:
## What?

Let's convert token printed on the console as asterisks for the sake of security.

## How?

Use winston library

dayo09:
I tried to apply pino to vscode Logger, but it seems that our logger need to be streamified for that.

I simply added Regex replace step to redact.

https://github.com/Samsung/ONE-vscode/pull/1610

PR #1610, opened 2023-07-20
[Logger] Add redact step

dayo09:
This commit adds redact step to a Logger.

---

For #1609

### DIFF
--- a/src/Utils/Logger.ts
+++ b/src/Utils/Logger.ts
@@ -33,26 +33,36 @@ const isDebugMode = process.env.VSCODE_DEBUG_MODE === "true";
  * - message: watch out
  */
 function _logStr(severity: string, tag: string, ...msgs: MsgList) {
-  let logStrList = [];
-
   if (msgs.length === 0) {
     // Do not print
     return "";
   }
 
-  for (let m of msgs) {
-    if (m instanceof Error) {
-      const err = m as Error;
-      logStrList.push(
-        `\nError was thrown:\n- name: ${err.name}\n- message: ${err.message}`
-      );
-    } else if (typeof m === "object") {
-      logStrList.push(`\n${m.constructor.name}: ${JSON.stringify(m)}`);
-    } else {
-      logStrList.push(`${m}`);
+  const flatten = (msgs: MsgList) => {
+    let logStrList = [];
+    for (let m of msgs) {
+      if (m instanceof Error) {
+        const err = m as Error;
+        logStrList.push(
+          `\nError was thrown:\n- name: ${err.name}\n- message: ${err.message}`
+        );
+      } else if (typeof m === "object") {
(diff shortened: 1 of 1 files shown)


jyoungyun (dismissed):
LGTM

dayo09:
@Samsung/one-vscode_committers PTAL :-D

src/Utils/Logger.ts
```diff
@@ -33,26 +33,36 @@ const isDebugMode = process.env.VSCODE_DEBUG_MODE === "true";
...
     }
-  }
-  const msg = logStrList.join(" ");
+    return logStrList.join(" ");
+  };
+
+  const redact = (msg: string) => {
+    // Replace github tokens with ********
+    const prefix = "ghp_";
+    const regex = new RegExp(`${prefix}[a-zA-Z0-9]+`, "g");
```
batcheu, line 61: How about to add the regular expression for ['Fine Grained Personal Acess Token'](https://docs.github.com/ko/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#fine-grained-personal-access-tokens) either?
I'm not sure whether RegExp() supports 'ORed' expression, if so we can add it as OR-ed option as like,
```
const prefix_classic = "ghp_";
const prefix_fgp = "github_pat_";
`^(${prefix_classic}[a-zA-Z0-9]+|${prefix_fgp}_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59})`
```

dayo09: @batcheu Look like a good idea to add the redact for fine grained token :-D

BTW I didn't find the string length of classic token so I supposed to redact all the strings starting with 'ghp_'. 
And I cannot find the string length of fine grained token written somewhere in github page either. 

If you have found some, could you let me know the link?

If not, let me simply add find-grained Regex as github_pat_[a-zA-Z0-9]+.

dayo09: @batcheu Applied :-D

batcheu: Oops, I didn't notice that you've added comments. 😮
Sorry for late reply 😢 

I just check the regular expression of new type of access token in below link.
* [Github token validation regular expression](https://gist.github.com/magnetikonline/073afe7909ffdd6f10ef06a00bc3bc88)

And I also found below GITHUB's official annoucement related with access token today.
* https://github.blog/changelog/2021-03-31-authentication-token-format-updates-are-generally-available/

> The length of our tokens is remaining the same for now. 
> However, GitHub tokens will likely increase in length in future updates, so integrators should plan to support tokens up to 255 characters after June 1, 2021.

So, I think that your suggestion is more proper than fixed length expression.

batcheu (approved):
LGTM

jyoungyun (approved):
LGTM

dayo09:
@Samsung/one-vscode_committers PTAL :-D

Issue #1621, opened 2023-09-01
[EdgeTPU Backend] Organization of this project


hohee-hee:
# WHY

EdgeTPU is a processing unit which is optimized for tensor operation. It supports faster inference of deep learning models in on-device environment. However, it can’t inference every type of models. Only `.tflite` model that is already compiled with EdgeTPU Compiler can be executed on EdgeTPU. Therefore, EdgeTPU can run inference a TensorFlow Lite Model that is compiled with EdgeTPU Compiler provided by Google Coral.

Currently, however, the compiler is only performed in the CLI environment so it cannot provide a better developer experience. So we are going to add a new vscode extension to do above process on ONE-vscode which consists of GUI environment. Through this project, users are expected to use EdgeTPU more intuitively and conveniently.

# WHAT

This project is to make vscode extension to use EdgeTPU compiler, which supports faster inference of deep learning model in on-device environment, in ONE conveniently.

### Functions

1. Implementation of Backend Interface for EdgeTPU
    
    The main function of this project is to implement a Backend Interface for EdgeTPU Compiler in ONE-vscode. The goal is to provide an environment where users can install EdgeTPU compiler with clicking a button instead of entering command line. In addition, the extension has a capability to convert standard tflite model to EdgeTPU’s tflite model without entering CLI commands.
    
2. Visualization EdgeTPU’s tflite model in ONE Explorer 
    
    One of the most important elements of extension of vscode is to visualize all models and files with tree structure. In this case of EdgeTPU, model files are compiled to tflite format but this appears same with general tflite models. Therefore, our new extension have to support visualization after considering how to distinguish naturally those two types of tflite models(general and EdgeTPU’s).
    

# HOW

### Mockup
![BackendForEdgeTPU (1)](https://github.com/Samsung/ONE-vscode/assets/122425990/8457da59-c771-4501-8b8a-ab917b2a5d16)

1. on Toolchain View
    1. Support to install EdgeTPU Compiler with command palette menu
    2. Support to set Default Toolchain with EdgeTPU Compiler
    3. Support to uninstall EdgeTPU Compiler on Toolchain
2. on ONE Explorer
    1. show `.tflite` by Base Model
    2. after clicking the file, config window will be shown
    3. choose some compile options for EdgeTPU Compiler on the window
    4. with play button, compile the Base Model file with EdgeTPU Compiler
    5. show a compiled tflite file for EdgeTPU in the sub part of the cfg file
3. on Console
[...]

PR #1625, opened 2023-09-05
Draft : [EdgeTPU] Adding Options for Edge TPU Compiler to CfgEditor

Hyeon-Uk:
For #1621 

This Draft is to add options for `Edge TPU Compiler` to cfgEditor

ONE-vscode-DCO-1.0-Signed-off-by: Hyeon-Uk <rlagusdnr120@gmail.com>

### DIFF
--- a/media/CfgEditor/cfgeditor.html
+++ b/media/CfgEditor/cfgeditor.html
@@ -244,16 +244,6 @@ limitations under the License.
                 </div>
                 <!-- 여기에 EdgeTPU Compile에 필요한 기본옵션을 추가하기 -->
                 <div class="option">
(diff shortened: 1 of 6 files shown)


dayo09:
image

I think it's well-done :-D Thanks a lot.

One thing that we should consider is : 'how to make it on/off'?

We'd better define a module on/off flag.
It's the bad practice but I suggest a way that you temporarily define a global variable, or a as an environment variable in extension context, etc.

So that we can gather our parts together and easily modulize later.

Hyeon-Uk:
From now on, we will separate the EdgeTPU-related Editor function from the existing cfgEditor code and modularize it.

dayo09:
Good Luck!!

PR #1628, opened 2023-09-05
Draft : [EdgeTPU] Adding EdgeTPU Toolchain

rhqjatn2398:
For #1621 

This draft is to add a EdgeTPU Toolchain to compile a model using EdgeTPU Compiler in ONE-vscode.

ONE-vscode-DCO-1.0-Signed-off-by: Bumsoo Ko rhqjatn2398@naver.com


script/prerequisitesForGetToolchains.sh
```diff
@@ -3,6 +3,11 @@
...
 PPA_NAME="ppa:one-compiler/onecc-docker"
-
 add-apt-repository -y ${PPA_NAME}
+
+# This shell script add PPA to get edgetpu-compiler package for EdgeTPUToolchain.
+
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+echo "deb https://packages.cloud.google.com/apt coral-edgetpu-stable main" | sudo tee /etc/apt/sources.list.d/coral-edgetpu.list
+
 apt-get update -yqq
```
dayo09, line 13: FYI, this part is better be seperated from the current script. Let's distinguish prerequisite stage for onecc and edgetpu.

profornnan: It seems like separating the prerequisites for `onecc` and `edgetpu` would be a good idea.

Is it okay to split `prerequisitesForGetToolchains.sh` into `prerequisitesForGetOneToolchain.sh` and `prerequisitesForGetEdgeTPUToolchain.sh`?

dayo09: Yes, that's cool :_D

script/prerequisitesForGetOneToolchain.sh
```diff
@@ -3,6 +3,6 @@
 # This shell script add PPA to get onecc-docker package for ONEToolchain.
```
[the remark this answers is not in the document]
profornnan: I'd like to confirm if I understood correctly. Here's what I understood:

1. In this Draft PR, the file name change for `prerequisitesForGetToolchains.sh` is excluded.
2. Creating a new branch from the main branch of ONE-vscode, renaming the file, and then creating a PR to merge it into the main branch.

Is my understanding correct?

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
```diff
@@ -0,0 +1,287 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```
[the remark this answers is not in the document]
profornnan: There is a mistake in the copyright year comment. I have applied it in the new commit.

profornnan:
Previously, testing was based on the Backends registered in `extension.ts`. So, adding the EdgeTPU Backend led to test failures.

Therefore, in order to conduct independent tests in `extensions.ts`, I modified the tests to clear `gToolchainEnvMap` and `globalBackendMap` before and after testing.

However, this led me to contemplate the direction for implementing EdgeTPU Compiler test code.

1. Currently, `ToolchainProvider.test.ts` adds `ToolchainEnv` in its setup for testing. Is it better to add the EdgeTPU here and have a single test code that checks all ToolchainEnvs?
2. Is it better to separate `ToolchainEnv` for each Compiler and write individual `ToolchainEnv` test code, rather than adding `ToolchainEnv` in the setup?

Related: https://github.com/SSDC-onE-FiVE/ONE-vscode/pull/64

src/Backend/Version.ts
```diff
@@ -23,7 +23,7 @@ class Version {
   constructor(
     major: number,
     minor: number | undefined,
-    patch: number | undefined,
+    patch?: number | undefined,
```
jyoungyun, line 26: You can use `undefined` value instead of changing this line.

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    const splitedVersion = _version.split(".");
+
+    if (splitedVersion.length > 3) {
+      throw Error("Invalid version format.");
+    }
+
+    let major: number | string;
+    let minor: number | string;
+    let patch: number | string;
```
jyoungyun, line 117: If the patch version is not valid in EdgeTPU, you can remove this line.

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
```diff
@@ -0,0 +1,295 @@
...
+    
+    if (isNaN(major) || isNaN(minor) || isNaN(patch)) {
+      throw Error("Invalid version format.");
+    }
+
+    if (splitedVersion.length === 2 && !option) {
+      return new Version(major, minor);
+    }
+
+    return new Version(major, minor, patch, option);
```
jyoungyun, line 138:
```suggestion
    return new Version(major, minor, undefined, option);
```

jyoungyun (commented):
In this draft code, there is no code that creates EdgeTPUToolchain class. You need to add `API.registerBackend(new EdgeTPUToolchain());` this line in extension.ts file.

src/Backend/Version.ts
profornnan: Thank you for your advice~!! I modified EdgeTPU Compiler Version to use undefined patch instead of changing the Version class

https://github.com/SSDC-onE-FiVE/ONE-vscode/pull/71

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
profornnan: The latest EdgeTPU Compiler version is 16.0 which has major and minor but old versions like 2.1.302470888 have the patch.
So, I consider the case that the version has the patch.

[EdgeTPU Compiler and Runtime versions](https://coral.ai/docs/edgetpu/compiler/#compiler-and-runtime-versions)

jyoungyun:
> 1. Currently, `ToolchainProvider.test.ts` adds `ToolchainEnv` in its setup for testing. Is it better to add the EdgeTPU here and have a single test code that checks all ToolchainEnvs?

EdgeTPU will be added to the extension.ts file. Adn this is a default operation to be added the ONE toolchain and EdgeTPU toolchain when one-vscode first loads. Therefore, we need to implement our test code based on this information.

> 2. Is it better to separate `ToolchainEnv` for each Compiler and write individual `ToolchainEnv` test code, rather than adding `ToolchainEnv` in the setup?

Could you think about it in current design? `gToolchainEnvMap` will have two Toolchain by default in ONE-vscode.

profornnan:
I think I misunderstood that there would be cases where EdgeTPU is removed from extension.ts 😢 

If ONE toolchain and EdgeTPU toolchain are added by default, we will modify the test code accordingly.

We will close the PR created previously in ONE-vscode repository and create a new PR.

Issue #1622, opened 2023-09-01
[EdgeTPU Backend] How to distinguish regular tflite model from edgeTPU-compiled tflite model?


Hyeon-Uk:
### Here is our project introduction

Project Introduction : #1621 

## Background : outputs after compiling tflite model with EdgeTPU

File before compiled with EdgeTPU and file after compiled with EdgeTPU are the same type, `.tflite`.  

The image below illustrates the procedure of converting a `.tflite` model into an Edge TPU Model using the Edge TPU compiler. The red box located at the bottom represents the `.tflite` file intended for compilation, while the blue box symbolizes the resulting output of the compilation process.

![image](https://github.com/Samsung/ONE-vscode/assets/43038815/f6fd0d26-3e3e-4a27-ad83-8cce29cae2fc)

It means that baseModel and Product are all tflite file. However, tflite file compiled with EdgeTPU doesn’t need to convert to circle file because edgetpu-custom-op cannot be inferenced in ONE-vscode. So this special tflite file is not a Base Model, it is a Product.

As a result, ONE explorer has to show the tflite file in Product type instead of  Base Model type after detecting tflite file and checks whether the tflite is a regular file or is from EdgeTPU Compiler. 

### The way to distinguish Base Model from Product in ONE-vscode

1. Find the Base Model with filename extension such as `.pb`, `.tflite.` 
2. Read the Base Model’s `.cfg` file and finds the results that corresponding with Output Path.
3. Display the results.

Of course, it assumes that the user performs all works in ONE-vscode interface.

## Problem

As I mentioned, because the file type of the EdgeTPU Compiler’s result is `.tflite`, ONE-vscode reads this result two times, while finding Base Model and finding Product with Output Path.  

We considered about the way to distinguish those two types(regular tflite and tflite from EdgeTPU Compiler) of tflite files.

## Process we discussed

1. After generating Base Models, set Products of each Models corresponding to Output Path in cfg files of each Models.
2. At this point, EdgeTPU Compiler’s outputs saved seperately
3. Exclude those files from Base Model and generate the rest files as the Base Model.

However, we believe that this straightforward method might incur a `high time complexity`. So we keep thinking about the better method.

If you have a better idea about it, please give us an innovative idea.

dayo09:
> ## Process we discussed
> 1. After generating Base Models, set Products of each Models corresponding to Output Path in cfg files of each Models.
> 2. At this point, EdgeTPU Compiler’s outputs saved seperately
[...]

PR #1630, opened 2023-09-07
Draft : [EdgeTPU] To distinguish between compiled tflite files with the edgetpu compiler in Explorer

Hyeon-Uk:
Related issue : #1622 
# Before we start...

we designed the `.cfg` file as shown in the picture below.
If the "one-import-edgetpu" value in the "onec" section is true, it means that you want to compile with the edge tpu compiler.
"output_path" in the "one-import-edgetpu" section means the location of the compiled product.
image
and this method does not depend on "flatbuffer".

# How?

This method uses the "one-import-edgetpu" value in the "onecc" section and the "output_path" value in the "one-import-edgetpu" section to determine whether a file was compiled with the Edge TPU compiler.

1. Before creating the TreeView component, we first utilize the getCfgList() function from OneStorage to fetch a list of `cfg` files.
2.  Following this, we collect and save the output path generated by the EdgeTPU compiler into a variable. [code](https://github.com/SSDC-onE-FiVE/ONE-vscode/blob/7a65fdf345f911dae279382bd52611f22c83f8d3/src/OneExplorer/OneStorage.ts#L284C1-L301C4)
3.  Then, within the _buildChildren() code, we implement a conditional check to determine if the file is a `.tflite` file representing the `baseModel`. [code](https://github.com/SSDC-onE-FiVE/ONE-vscode/blob/7a65fdf345f911dae279382bd52611f22c83f8d3/src/OneExplorer/OneExplorer.ts#L257)
4. In this check, we verify whether the model exists within the previously collected variables. [code](https://github.com/SSDC-onE-FiVE/ONE-vscode/blob/7a65fdf345f911dae279382bd52611f22c83f8d3/src/OneExplorer/OneExplorer.ts#L244)
5. If we find the model in the collection, we've made the decision not to categorize it as a 'baseModel' because it originates from the EdgeTPU Compiler.

# Advantage
- Simple
- Allows custom output_path of edgetpu-compiler's product

#  Problem
- It depends on the output_path in the `.cfg` file. so if it change, explorer can't find the products

# Finally
This is not the perfect way to solve this problem. we keep thinking about it, and as a result, we found a way to use "flatbuffer". However, due to performance time problems, optimization of "flat buffers", or  other methods, must be devised.

### DIFF
--- a/media/CfgEditor/cfgeditor.html
+++ b/media/CfgEditor/cfgeditor.html
@@ -84,6 +84,7 @@ limitations under the License.
                         <vscode-radio value="tflite">tflite</vscode-radio>
                         <vscode-radio value="onnx">onnx</vscode-radio>
                         <vscode-radio value="bcq" disabled="true">bcq</vscode-radio>
+                        <vscode-radio value="edgetpu">edge_tpu</vscode-radio>
                     </vscode-radio-group>
                 </div>
             </div>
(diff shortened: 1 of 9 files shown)


dayo09:
From the issue #1622,

              To make long short; we concluded that using _edgetpu.tflite prefix for edgetpu compiled models.

_Originally posted by @dayo09 in https://github.com/Samsung/ONE-vscode/issues/1622#issuecomment-1788462541_

PR #1631, opened 2023-09-08
[API] Clarify APIs

dayo09:
This commit revisits API to split ONE-toolchain backend registration from the api itself. It aim to show the extension's api functions clearly.

### DIFF
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -20,7 +20,6 @@ import { Backend } from "./Backend";
 import { gToolchainEnvMap, ToolchainEnv } from "../Toolchain/ToolchainEnv";
 import { Logger } from "../Utils/Logger";
 import { Executor } from "./Executor";
-import { OneToolchain } from "./One/OneToolchain";
 
 /**
  * Interface of backend map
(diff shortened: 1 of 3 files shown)


src/Backend/API.ts
```diff
@@ -35,47 +34,53 @@ let globalBackendMap: BackendMap = {};
 // List of Executor extensions registered
 let globalExecutorArray: Executor[] = [];
 
-function backendRegistrationApi() {
```
dayo09, line 38: I changed this name from `backendRegistraitionAPI` to `API` because it represents general API exports from ONE-vscode. Actually it contains 'executor registration API' too.

src/Backend/API.ts
```diff
@@ -35,47 +34,53 @@ let globalBackendMap: BackendMap = {};
...
+    "Executor",
+    executor.name(),
+    "was registered into ONE-vscode."
+  );
+
+  // NOTE: This might not 100% guaratee the activating extension has been done.
+  //   - link: https://github.com/Samsung/ONE-vscode/pull/1101#issuecomment-1195099002
+  // TODO: Consider better way to refresh toolchainView after backend's registration.
+  vscode.commands.executeCommand("one.device.refresh");
+};
```
jyoungyun, line 79: I don't know whether we need this function now. However, this PR just changes the code structure. Let's think about how to do this function later.

dayo09: I agree. I thought about removing this one because it's not used anywhere. I left this one in this PR but I agree that we clean this up

mhs4670go (approved):
LGTM

PR #1632, opened 2023-09-08
[OneExplorer] Revisit ConfigObject

dayo09:
This commit removes logics from constructor.
Doing so many things in constructor is a bad practice because any unexpected error cannot be handled.

### DIFF
--- a/src/OneExplorer/ConfigObject.ts
+++ b/src/OneExplorer/ConfigObject.ts
@@ -112,12 +112,19 @@ export class ConfigObj {
     return found ? true : false;
   }
 
+  private init() {
+    this.obj = {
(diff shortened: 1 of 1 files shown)


src/OneExplorer/ConfigObject.ts
```diff
@@ -172,7 +179,10 @@ export class ConfigObj {
       return null;
     }
 
-    return new ConfigObj(uri, obj as Cfg);
+    const cfgObj = new ConfigObj(uri, obj as Cfg);
+    cfgObj.init();
+
+    return cfgObj;
```
jyoungyun, line 185: It creates `ConfigObj` whenever `createConfigObj` function is called. Is this your intention?

dayo09: Yes, it's to ease the process of undefined uri handling.

PR #1635, opened 2023-09-12
[API] Set context of backend enabled

dayo09:
This commit sets vscode context of enabled backend.

### DIFF
--- a/src/Backend/API.ts
+++ b/src/Backend/API.ts
@@ -43,6 +43,11 @@ const registerBackend = (backend: Backend) => {
   const compiler = backend.compiler();
   if (compiler) {
     gToolchainEnvMap[backend.name()] = new ToolchainEnv(compiler);
+    vscode.commands.executeCommand(
+      "setContext",
+      `one:backend.${backend.name()}`,
+      "enabled"
+    );
   }
   const executor = backend.executor();
   if (executor) {
@@ -77,6 +82,8 @@ const registerExecutor = (executor: Executor) => {
   // TODO: Consider better way to refresh toolchainView after backend's registration.
   vscode.commands.executeCommand("one.device.refresh");
 };
+
+// TODO Add removeBackend
 
 export const API = {
   registerBackend,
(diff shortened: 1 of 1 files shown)


src/Backend/API.ts
```diff
@@ -29,20 +29,33 @@ interface BackendMap {
...
+}
+
 const registerBackend = (backend: Backend) => {
   const backendName = backend.name();
   assert(backendName.length > 0);
   globalBackendMap[backendName] = backend;
   const compiler = backend.compiler();
   if (compiler) {
     gToolchainEnvMap[backend.name()] = new ToolchainEnv(compiler);
+    vscode.commands.executeCommand(
```
dayo09, line 54: FYI, this is the context that can be referred in package.json

src/Backend/API.ts
```diff
@@ -29,20 +29,33 @@ interface BackendMap {
...
 let globalExecutorArray: Executor[] = [];
 
 const logTag = "API";
 
+// TODO Move outside API.ts
+export namespace BackendContext {
+  export const isRegistered = (backendName: string) => {
+    return globalBackendMap[backendName] !== undefined;
+  };
+}
```
dayo09, line 45: FYI, this is the context that can be referred in any modules.

src/Backend/API.ts
jyoungyun: There is no `getContext` API in vscode. Therefore, both apis should be provided to get the backend status.

jyoungyun (approved):
LGTM

profornnan (approved):
It seems like it will be helpful for modularization. Thank you 😄

rhqjatn2398 (approved):
LGTM

hohee-hee:
LGTM
We'll modulize EdgeTPU Compiler based on this context😊

hohee-hee (approved):
LGTM
We'll modulize EdgeTPU Compiler based on this context😊

PR #1639, opened 2023-09-14
Draft : [EdgeTpu] Separate config logic

imsw0529:
related to #1621 

This draft is for separating the config logic of ONE and EdgeTpu.
We make new filename extension that support Edge Tpu compile configuration ".edgetpucfg". It handles products of Edge Tpu compiler so it's logic is different from onecc configuration. So I separate it's logic from ConfigObject.ts and while doing this, I separate onecc config logic too.

If there are any parts to fix, please let me know.


dayo09:
I like this overal approach to separate OneConfigSetting and EdgeTpuConfigSetting. :-D

src/OneExplorer/OneConfigSetting.ts
```diff
@@ -0,0 +1,250 @@
...
+type CfgOneImportTflite = any;
+type CfgOneImportOnnx = any;
+type CfgOneImportTf = any;
+
+export class OneConfigSetting {
+  static baseModelsLocatorRunner: LocatorRunner;
+  static productsLocatorRunner: LocatorRunner;
+  // TODO: make sections for updateBaseModelField method
+
+  static getBaseModelsLocatorRunner(): LocatorRunner {
```
dayo09, line 19: Q. Why should it be Singleton? I think it can simply be a class and function.

src/OneExplorer/ConfigObject.ts
```diff
@@ -214,37 +216,10 @@ export class ConfigObj {
...
-      artifactAttr: {
-        ext: ".onnx",
-        icon: new vscode.ThemeIcon("symbol-variable"),
-      },
-      locator: new Locator((value: string) =>
-        LocatorRunner.searchWithExt(".onnx", value)
-      ),
-    });
+    let locatorRunner = OneConfigSetting.getBaseModelsLocatorRunner();
+    if (path.extname(filePath) === ".edgetpucfg") {
```
dayo09, line 220: Where does it check if the edgetpu backend is registered or not?

src/OneExplorer/OneConfigSetting.ts
imsw0529: A. I think that if config use same setting, config use same locatorRunner. If it has problem, I will change it to simple class function.

src/OneExplorer/ConfigObject.ts
imsw0529: I temporarily check just filename. I will add the function that check if the edgetpu backend is registered or not soon.

imsw0529:
I make abstract class **ConfigSetting** that is super class of **OneConfigSetting** and **EdgeTpuConfigSetting**. While doing this, I change ConfigSetting's function from singleton to simple class function.
> A. I think that if config use same setting, config use same locatorRunner. If it has problem, I will change it to simple class function.

src/OneExplorer/ConfigObject.ts
```diff
@@ -210,41 +229,12 @@ export class ConfigObj {
    */
   private static parseBaseModels = (
     filePath: string,
-    iniObj: object
+    iniObj: object,
+    configSetting: ConfigSetting
   ): Artifact[] => {
```
imsw0529, line 234: I attended that
- config object has config setting as member.
- when config object parse base models or products, it uses it's member config setting.

But parse function is static function, so it can't use member variable. So I add configSetting as input of parse function. Is there better way?

src/OneExplorer/ConfigSetting.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```
dayo09, line 2: 2022 -> 2023

src/OneExplorer/OneConfigSetting.ts
dayo09: Making something Singleton is to use a same object across the whole project, it would be for sustaining a certain context or something. In this case, I think it's enough to go with a simple class. :-D

src/OneExplorer/ConfigObject.ts
dayo09: I agree with the overal approach. I suggest one point: how about defining 'cfg type' and branch to call the corresponding configSetting in the function? To make our cfgobj instances lighter.

src/OneExplorer/ConfigSetting.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export abstract class ConfigSetting {
+  baseModelsLocatorRunner: LocatorRunner;
+  productsLocatorRunner: LocatorRunner;
+  // TODO: make sections for updateBaseModelField method
+
+  constructor() {
+    this.baseModelsLocatorRunner = new LocatorRunner();
+    this.productsLocatorRunner = new LocatorRunner();
+    this._init();
```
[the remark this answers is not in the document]
dayo09: I should confess that I did so many things in ConfigObj constructors, to say... I am planning to change that, too :-D (See #1632)

imsw0529:
### About Backend detection

We supposed that when there are edge-tpu-backend associated files(.edgetpucfg or model_edgetpu.tflie ans so on), How OneExplorer would treat it. There are two approach:

1. **Do not show** edge-tpu-backend associated files **even though** there are that files in workspace when edge-tpu-backend is not installed.
2. **Show** edge-tpu-backend and **block** function **run cfg** if edge-tpu-backend is not installed.

We think that second approach is appropriate. What is your opinion?

imsw0529:
done :-)

PR #1641, opened 2023-09-15
[OneExplorer] Relieve function complexity

dayo09:
This commit is to relieve function complextiy in test.

### DIFF
--- a/src/Tests/OneExplorer/ConfigObject.test.ts
+++ b/src/Tests/OneExplorer/ConfigObject.test.ts
@@ -46,16 +46,15 @@ suite("OneExplorer", function () {
       });
 
       test("NEG: Create config of a file without any valid content", function () {
-        const configName = "model.cfg";
         const content = `
         empty content
         `;
 
         // Write a file inside temp directory
-        testBuilder.writeFileSync(configName, content);
+        testBuilder.writeFileSync("model.cfg", content);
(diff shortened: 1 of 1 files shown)


dayo09:
cc(Cyclomatic complexity) violation number decreased from 112 to 65

jyoungyun (approved):
LGTM

src/Tests/OneExplorer/ConfigObject.test.ts
```diff
@@ -46,16 +46,15 @@ suite("OneExplorer", function () {
       });
 
       test("NEG: Create config of a file without any valid content", function () {
-        const configName = "model.cfg";
```
shs-park, line 49: I think using const seems to be better for code readability, but it seems to increase CC violations(!!). 
I don't understand why using simple constants/variables is related to cyclomatic complexity score.
😢

shs-park (approved):
The code seems to good to me, but I left a question.

src/Tests/OneExplorer/ConfigObject.test.ts
dayo09: I totally agree... I just followed the guide anyways. 
In terms of readability, I first thought that using const is better but some people told me that writing down the file name is better... 🤔

Issue #1624, opened 2023-09-04
Prepare official release v1.0


jyoungyun:
## What

Let's prepare official release v1.0.

In order to make release preparation stable, we will do branch-out first and fix codes in release branch.
Of course the change should be merged to main branch. It is to avoid code conflict or other unexpected issues during the SSAFY project. :)

### Release date

2023.10

#### OneExplorer
- [ ] Multiple model/cfg files are shown
- [ ] Refresh
- [ ] Hide&Show extra files
- [ ] Rename cfg/model
- [ ] Refactor
- [ ] Create cfg
- [ ] Delete directory/cfg/model
- [ ] Reveal in explorer view
- [ ] Show circle in CircleViewer
- [ ] Show cfg in CfgEditor
- [ ] Auto refresh on file creation/deletion/change (only on local mode)

#### Toolchain (Compiler)
- [ ] Shows currently installed ONE/TRIV2 backend lists
- [ ] Setting prerequisites
- [ ] Setting default toolchain
- [ ] Run cfg (sample.tflite)
   - [ ] ONE backend (onecc-docker)
      - [ ] import 🟡 (Known issue, rate limit exceeded error)
   - [ ] tv2 backend (triv2-toolchain-latest)
      - [ ] import
      - [ ] codegen

### Release Action

- [ ] Branch out

### Release Criteria
- [ ] TEM coverage (>80%)
- [ ] TCM (New NEgative / New Positive > 1)
- [ ] Svace & Coverity (critical = 0)
- [ ] SAM (> 4.0) #1566

dayo09:
### SAM Score

(main branch, cad87ff)
- Typescript: 4.50
- Javascript: 4.94

<details>

>  ### SAM Score, typescript, 2023-09-15 (main branch, cad87ff)
> 
> https://analysishub.sec.samsung.net/service/analyses/2373959
> 
> ![image](https://github.com/Samsung/ONE-vscode/assets/17171963/795eccc9-4e31-4c89-acdd-484d2304c09d)
> 
> Thanks to @hseok-oh
> 
> _Originally posted by @dayo09 in https://github.com/Samsung/ONE-vscode/issues/1566#issuecomment-1720732074_

> ### Javascript SAM Score, 2023-09-15 (main branch, [cad87ff](https://github.com/Samsung/ONE-vscode/commit/cad87ff0d7b885dbc0e06efc543f0a3d5e4927de))
[...]

PR #1643, opened 2023-09-15
[media] Check origin of messages

dayo09:
This commit checks origin of messages in event handlers

----

For https://github.com/Samsung/ONE-vscode/issues/1624#issuecomment-1720750787

### DIFF
--- a/media/CircleGraph/index.js
+++ b/media/CircleGraph/index.js
@@ -147,6 +147,10 @@ host.BrowserHost = class {
     });
 
     this.window.addEventListener("message", (event) => {
+      if (window.origin !== event.origin) {
+        console.log("Unexpected Origin: ", event.origin);
+        return;
+      }
       const message = event.data;
       switch (message.command) {
         case "loadmodel":
--- a/media/CfgEditor/index.js
+++ b/media/CfgEditor/index.js
@@ -53,6 +53,10 @@ function main() {
 
   // event from vscode extension
   window.addEventListener("message", (event) => {
+    if (window.origin !== event.origin) {
+      console.log("Unexpected Origin: ", event.origin);
+      return;
+    }
     const message = event.data;
     switch (message.type) {
       case "displayCfgToEditor":
--- a/media/CircleEditor/index.js
+++ b/media/CircleEditor/index.js
@@ -134,6 +134,10 @@ host.BrowserHost = class {
     });
 
     this.window.addEventListener("message", (event) => {
+      if (window.origin !== event.origin) {
+        console.log("Unexpected Origin: ", event.origin);
+        return;
+      }
       const message = event.data;
       switch (message.command) {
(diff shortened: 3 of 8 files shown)


media/CircleGraph/index.js
```diff
@@ -147,6 +147,9 @@ host.BrowserHost = class {
     });
 
     this.window.addEventListener("message", (event) => {
+      if (window.origin !== event.origin) {
+        return;
+      }
```
dayo09, line 152: I just applied it empty-minded. 
This one makes circle viewer not working. 
Let me know if anyone knows the right way.

media/CircleGraph/index.js
```diff
@@ -430,6 +433,9 @@ host.BrowserHost = class {
     return new Promise((resolve, reject) => {
       if (url.startsWith("vscode-webview://")) {
         const messageHandler = (event) => {
+          if(window.origin === event.origin){
+            return;
+          }
```
dayo09, line 438: (ditto)
I just applied it empty-minded. 
This one makes circle viewer not working. 
Let me know if anyone knows the right way.

media/CircleGraph/index.js
```diff
@@ -147,6 +147,10 @@ host.BrowserHost = class {
     });
 
     this.window.addEventListener("message", (event) => {
+      if (window.origin !== event.origin) {
+        console.log("Unexpected Origin: ", event.origin);
+        return;
+      }
```
dayo09, line 153: @seanshpark I noticed that there is a potential security problem from `Verify the origin of the received message`.

I assumed that the messages are coming from the window's origin, would it be okay for CircleGraph?

seanshpark: I'm not sure about technical background of this, but I think `event` itself would be coming from vscode process(? or thread?, the node-js).
And not sure about the fix but if this no problem loading the model, I think it would be OK.

dayo09:
@jyoungyun PTAL :-D

media/CircleGraph/index.js
jyoungyun: I agree this changes because this is a similar solution from sonarsource.com.

https://rules.sonarsource.com/javascript/RSPEC-2819/
```
window.addEventListener("message", function(event) {

  if (event.origin !== "http://example.org") // Compliant
    return;

  console.log(event.data)
});
```

jyoungyun:
@dayo09 Could you update the commit title so that there is no `DRAFT:` prefix. :)

media/CircleGraph/index.js
dayo09: @jyoungyun I actually referred that solution :-D

jyoungyun (approved):
LGTM

seanshpark (approved):
LGTM

PR #1648, opened 2023-09-19
[MPQEditor] Add create menu from OneExplorer

dayo09:
This commit adds the 'create mpq configuration menu' from OneExplorer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -283,7 +283,12 @@
       },
       {
         "command": "one.editor.mpq.createFromDefaultExplorer",
-        "title": "Create MPQ json",
+        "title": "ONE: Create Mixed Precision Quantization Configuration",
+        "category": "ONE"
+      },
+      {
+        "command": "one.editor.mpq.createFromOneExplorer",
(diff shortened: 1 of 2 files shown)


stamalakhov (approved):

package.json
```diff
@@ -283,7 +283,12 @@
       },
       {
         "command": "one.editor.mpq.createFromDefaultExplorer",
-        "title": "Create MPQ json",
+        "title": "ONE: Create Mixed Precision Quantization Configuration",
```
jyoungyun, line 286: Do we need `ONE:` prefix in title?

dayo09: It's the menu shown from the default explorer, so I added it. 
The menu on ONE Explorer doesn't include "ONE:" instead.

dayo09: Actually, I am rather worried that ONE Explorer cannot show the created mpq configuration files for now. 😨

jyoungyun (approved):
LGTM

PR #1650, opened 2023-09-19
[Execute] Expand node entries at HostPC parent node

jyoungyun:
This commit expands node entries at HostPC parent node by default.


shs-park (approved):
LGTM!
=)

Issue #1652, opened 2023-09-19
[OneExplorer] Error in `Delete` key


jyoungyun:
## What

After installing the latest ONE-vscode as a package file, the `Delete` key does not work.

### How to reproduce

1. create package file
   ```
   ONE-vscode$ vsce package
   ```
2. install vsce package file 
   - Extensions Tab -> ... button on the top right -> Install from VSIX...

dayo09:
It works on my local environment. 
I could delete files by hitting delete key.
I tried both on tflite, circle, config and directory.

jyoungyun:
My explanation is not enough to understand this issue. :(

The delete action of ONE explorer works well. However, default delete key actions like delete string, delete file in File explorer are not working after the latest ONE-vscode is installed.

dayo09:
@jyoungyun Sure. Now I see the problem and it's the same on my computer. I will restrict the conditions. :-D

PR #1654, opened 2023-09-19
[OneExplorer] Activate keybindings with condition

dayo09:
This commit makes our keybindings activate only when ONE Explorer is visible and no input focus.

---

For #1652

### DIFF
--- a/package.json
+++ b/package.json
@@ -301,12 +301,12 @@
       {
         "command": "one.explorer.deleteOnShortcut",
         "key": "delete",
-        "when": "view.OneExplorerView.visible && !explorerResourceReadonly && !inputFocus"
+        "when": "view.OneExplorerView.visible && !inputFocus"
       },
       {
         "command": "one.explorer.renameOnShortcut",
         "key": "F2",
-        "when": "view.OneExplorerView.visible && one.explorer:hasSelectedCfg && !explorerResourceReadonly && !inputFocus"
+        "when": "view.OneExplorerView.visible && one.explorer:hasSelectedCfg && !inputFocus"
       }
     ],
     "menus": {
--- a/package.json
+++ b/package.json
@@ -291,7 +291,7 @@
       {
         "command": "one.explorer.runSingleSelectedCfg",
         "key": "ctrl+f7",
-        "when": "OneExplorerView.active && one.explorer:hasSelectedCfg"
+        "when": "view.OneExplorerView.visible && one.explorer:hasSelectedCfg"
       },
       {
         "command": "one.cfgEditor.setDefaultValues",
@@ -301,12 +301,12 @@
       {
         "command": "one.explorer.deleteOnShortcut",
         "key": "delete",
-        "when": "OneExplorerView.active"
+        "when": "view.OneExplorerView.visible && !explorerResourceReadonly && !inputFocus"
       },
       {
         "command": "one.explorer.renameOnShortcut",
         "key": "F2",
-        "when": "OneExplorerView.active && one.explorer:hasSelectedCfg"
+        "when": "view.OneExplorerView.visible && one.explorer:hasSelectedCfg && !explorerResourceReadonly && !inputFocus"
       }
     ],
     "menus": {


hseok-oh (approved):
LGTM

jyoungyun:
When I tested this PR, the selected file in ONE explorer is not deleted using `Delete` key.

If I check the context key value when I click any log file in ONE explorer, these values appear as follows.
```
view.OneExplorerView.visible: true
explorerResourceReadonly: false
inputFocus: true
```

You can use `Developer: Inspect Context Keys` in command tool to show the context value.

package.json
```diff
@@ -301,12 +301,12 @@
       {
         "command": "one.explorer.deleteOnShortcut",
         "key": "delete",
-        "when": "OneExplorerView.active"
+        "when": "view.OneExplorerView.visible && !explorerResourceReadonly && !inputFocus"
```
hseok-oh, line 304: @dayo09 Where is `explorerResourceReadonly` comes from?

jyoungyun: You can refer [this url](https://www.whitphx.info/posts/20230217-vscode-undocumented-contexts/) about undocumented contexts. :(

dayo09:
### Other shortcuts
 'moveFileToTrach' (Delete shortcut) has when conditions as `explorerResourceMoveableToTrash && filesExplroerFocus && foldersViewVisible && !explorerResourceReadonly && !inputFocus`

### Context Values
(1) !inputFocus
- TRUE when editing a file.

- When a user edits a json file and then clicks the file in explorer view, 
  - In default explorer view, `delete` key shortcut is enabled.
    - Because `inputFocus` turns false.
  - In ONE explorer view, `delete` key shortcut is still disabled.  (@jyoungyun has found out this scenario)
    - Because `inputFocus` is still true.

(2) view.OneExplorerView.visible
- TRUE when the view is visible

(3) OneExplorerView.visible, OneExplorerView.active
- TRUE always when ONE view was once active

jyoungyun:
@dayo09 When I click a json file in File Explorer, it just shows a json file. In this case, the `inputFocus` is not true. However, in ONE explorer, it always opens a json file as editing mode. Isn't this relevant?

dayo09:
@jyoungyun Yes, that is the thing. That's the explorer's behavior difference. Default explorer has supports preview. I tried that thing beforehand but I couldn't find a documentation about that. I guess the current behavior is the best we could have for now.

package.json
dayo09: @hseok-oh Actually, I thought it's a flag whether the selected node's uri file is readonly or not, but it was always true for readonly file both on default explorer and one explorer...

dayo09:
### `explorerResourceReadonly`


This is the code from microsoft/vscode.
```ts
    
/**
 * Context Keys to use with keybindings for the Explorer and Open Editors view
 */
export const ExplorerViewletVisibleContext = new RawContextKey<boolean>('explorerViewletVisible', true, { type: 'boolean', description: localize('explorerViewletVisible', "True when the EXPLORER viewlet is visible.") });
export const ExplorerFolderContext = new RawContextKey<boolean>('explorerResourceIsFolder', false, { type: 'boolean', description: localize('explorerResourceIsFolder', "True when the focused item in the EXPLORER is a folder.") });
export const ExplorerResourceReadonlyContext = new RawContextKey<boolean>('explorerResourceReadonly', false, { type: 'boolean', description: localize('explorerResourceReadonly', "True when the focused item in the EXPLORER is readonly.") });
export const ExplorerResourceNotReadonlyContext = ExplorerResourceReadonlyContext.toNegated();
/**
```

jyoungyun:
@dayo09 Does this PR work? It still uses the `inputFocus` context. However, this context key is always set to true in ONE explorer.

dayo09:
@jyoungyun 

> @dayo09 Does this PR work? It still uses the inputFocus context. However, this context key is always set to true in ONE explorer.

It works except for the case of `When a user edits a json file and then clicks the file in explorer view,`. `inputFocus` context works as expected.

jyoungyun (approved):
LGTM

Unfortunately, we can not delete log file in ONE explorer. 😅

hseok-oh (approved):

PR #1659, opened 2023-09-20
[CircleGraph] Support onnx

glistening:
Now, CircleGraph can show onnx.

onnx.js, onnx-proto.js and  onnx-metadata.json are copied from original netron r5.2.1.

To avoid name conflicts with others (e.g. tflite, circle), all occurrences of `$root` are replaced with `onnx_root` from onnx-proto.js.

ONE-vscode-DCO-1.0-Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>


dayo09 (approved):
It works very well :-D Thanks.

jyoungyun (approved):
LGTM

glistening:
Just for recording, I add more explanation why I chosen `onnx-*` from r5.2.1.

In short, the latest `onnx-*` does not work with old one in `CircleGraph`.
`view.Parameter` is renamed to `Argument` and there is no more `getter arguments()`.
It makes trouble while accessing `arguments`. For example,

<sub> media/CircleGraph/view.js </sub>
```js
  renderGraph(model, graph) {
        ...
        const rotate = graph.nodes.every(
          (node) =>
            node.inputs.filter((input) =>
              input.arguments.every((argument) => !argument.initializer)
            ).length === 0 && node.outputs.length === 0
        );
```

PR #1660, opened 2023-09-20
[Execute] Introduce SimulatorDevice and TargetDevice

jyoungyun:
This commit introduces SimulatorDevice and TargetDevice.

ONE-vscode-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #1661, opened 2023-09-20
[Execute] Register SimulatorDevice with Toolchain information

jyoungyun:
This commit registers SimulatorDevice with Toolchain information.
It does not add device according to the DeviceSpec. All devices are
added to the DeviceManager directly.

wait for #1660

### DIFF
--- a/src/Execute/DeviceViewProvider.ts
+++ b/src/Execute/DeviceViewProvider.ts
@@ -23,7 +23,7 @@ import { Executor } from "../Backend/Executor";
 import { DeviceSpec, supportedSpecs } from "../Backend/Spec";
 import { Logger } from "../Utils/Logger";
 
-import { Device } from "./Device";
+import { Device, SimulatorDevice } from "./Device";
 import { DeviceManager } from "./DeviceManager";
 
 type DeviceTreeView = DeviceViewNode | undefined | void;
@@ -39,14 +39,20 @@ export const deviceManagerList = ["local"];
 export enum NodeType {
   /**
    * A Execution system that contains multiple Devices.
+   * 1. local
+   * 2. TODO remote
    */
   deviceManager,
   /**
    * A single device dedicated to certain device.
+   * // TODO Remove device node type
    */
   device,
   /**
    * A Executor that run on certain device.
+   * 1. the actual target connected to the bridge
+   * 2. simulator on host machine
+   * 3. TODO simulator on docker image
    */
   executor,
   /**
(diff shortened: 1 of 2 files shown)


jyoungyun:
This PR changes the Device view UI. It does not use `HostPC` parent node. Instead, it adds the device item directly to the `local` parent node. Other docker or target devices are also added directly to the `local` node.

src/Tests/Execute/DeviceViewProvider.test.ts
```diff
@@ -114,28 +114,6 @@ suite("DeviceViewProvider", function () {
         done();
       });
     });
-    test("get Children under deviceManager Node", function (done) {
```
jyoungyun, line 117: This test case is related to `Executor` node type. `Executor` node type is not used anymore. So I remove this test case.

shs-park (approved):
LGTM!
=)

src/Execute/DeviceViewProvider.ts
```diff
@@ -191,42 +197,59 @@ export class DeviceViewProvider
...
-      let executorList: Executor[] = [];
-      const entries = Object.entries(globalBackendMap);
-      for (const entry of entries) {
-        const compiler = entry[1].compiler();
-        if (compiler) {
-          for (const toolchainType of compiler.getToolchainTypes()) {
-            if (compiler.getInstalledToolchains(toolchainType).length > 0) {
-              executorList.push(...entry[1].executors());
-              break;
+        Object.entries(globalBackendMap).map((entry) => {
```
dayo09, line 216:
```suggestion
        Object.entries(globalBackendMap).map((_backendName, backend) => {
```

How about using `type destructuring` for readability?
I added underbar to _backendName because the value is not used in the loop

src/Execute/DeviceViewProvider.ts
```diff
@@ -191,42 +197,59 @@ export class DeviceViewProvider
...
-      let executorList: Executor[] = [];
-      const entries = Object.entries(globalBackendMap);
-      for (const entry of entries) {
-        const compiler = entry[1].compiler();
-        if (compiler) {
-          for (const toolchainType of compiler.getToolchainTypes()) {
-            if (compiler.getInstalledToolchains(toolchainType).length > 0) {
-              executorList.push(...entry[1].executors());
-              break;
+        Object.entries(globalBackendMap).map(([_, backend]) => {
```
dayo09, line 216: 👍

dayo09 (approved):
LGTM :-D

PR #1680, opened 2023-09-25
[workflows] Use boolean type in publish extension workflow

jyoungyun:
This commit uses boolean type to process true/false value in publish extension workflow.


dayo09 (approved):
LGTM :-D

PR #1681, opened 2023-09-25
[Execute] Use forEach instead of Map

jyoungyun:
This commit uses forEach instead of Map because it does not use the return value.


jyoungyun:
Before this PR, Potential Defect Summary score is 3.97.
If tested with this PR, Potential Defect Summary score is 5.0.

PR #1682, opened 2023-09-26
[EdgeTpu] Introduce EdgeTpuConfigSetting

imsw0529:
Related to issue #1621, draft #1639

This commit introduce EdgeTpuConfigSetting that has ConfigObject logic for EdgeTpu Backend

- Introduce class EdgeTpuConfigSetting extends ConfigSetting for EdgeTpu Backend logic
- Apply EdgeTpuConfigSetting to ConfigObject
- Add Test code for EdgeTpuConfigSetting

Co-author: SeungHyunH <vkflek1232@naver.com>

ONE-vscode-DCO-1.0-Signed-off-by: Seongwon Im <imsw0529@gmail.com>


imsw0529:
We have a question.
https://github.com/Samsung/ONE-vscode/pull/1639#issuecomment-1729055215
> ### About Backend detection
> We supposed that when there are edge-tpu-backend associated files(.edgetpucfg or model_edgetpu.tflie ans so on), How OneExplorer would treat it. There are two approach:
> 
> 1. **Do not show** edge-tpu-backend associated files **even though** there are that files in workspace when edge-tpu-backend is not installed.
> 2. **Show** edge-tpu-backend and **block** function **run cfg** if edge-tpu-backend is not installed.
> 
> We think that second approach is appropriate. What is your opinion?

SeungHyunH (approved):

src/OneExplorer/ConfigObject.ts
```diff
@@ -140,6 +148,10 @@ export class ConfigObj {
     this.uri = uri;
     this.rawObj = rawObj;
     this.configType = CfgType.one;
+    const ext = path.extname(uri.fsPath);
+    if (ext === EdgeTpuConfigSetting.ext) {
+      this.configType = CfgType.edgeTpu;
+    }
```
dayo09, line 154: How about tidying these by setting configType with similar conditional statement?

SeungHyunH:
```
    let configSetting: ConfigSetting;
    this.configType = CfgType.one;
    const ext = path.extname(uri.fsPath);
    if (ext === EdgeTpuConfigSetting.ext) {
      this.configType = CfgType.edgeTpu;
    }
    switch (this.configType) {
      case CfgType.edgeTpu:
        configSetting = new EdgeTpuConfigSetting();
        break;
      case CfgType.one:
      default:
        configSetting = new OneConfigSetting();
    }
    this.uri = uri;
    this.rawObj = rawObj;
```
There seems to be an issue where you can set CfgType to edgetpu after creating OneConfigSetting. What do you think about this approach?

dayo09, line 154:
```suggestion
    const ext = path.extname(uri.fsPath);
    switch(ext) {
      case EdgeTpuConfigSetting.ext:
        this.configType = CfgType.edgeTpu;
        break;
      case OneConfigSetting.ext:
      default:
        this.configType = CfgType.one;
    }
```

I meant, like this..?

imsw0529: Thank you. I understand it. I will apply this :-D

dayo09 (approved):
LGTM, Thanks!! :-D

Issue #1678, opened 2023-09-25
[AHUB] MOCHA.NO_ASSERTION checker wrongly accuses teardown hook


dayo09:
## What?

MOCHA.NO_ASSERTION checker wrongly accuses teardown hook.

![image](https://github.com/Samsung/ONE-vscode/assets/17171963/a3ac3699-d90c-43d4-b4e2-285112e7d994)

https://github.com/Samsung/ONE-vscode/blob/main/src/Tests/OneExplorer/OneExplorer.test.ts#L38-L40

I reported this case to https://github.sec.samsung.net/RS-SA/analysishub-service-webapi/issues/4102

dayo09:
### Workaround

https://analysishub.sec.samsung.net/service/analyses/2400612#defects

By relocating the teardown function all the way down to the bottom, we can instantly dodge the warning.

dayo09:
It's fixed using a workaround.

PR #1686, opened 2023-09-26
[EdgeTPU] Update Toolchain test code

profornnan:
This commit updates the Backend.test.ts, ToolchainProvider.test.ts, InstallQuickInput.test.ts files so that the testing codes work independent from addition of Backend and ToolchainEnv.

AS-IS:
Even though the purpose is not to test the registration of backend and toolchainEnv, the test codes should be updated whenever new backends and toolchainEnvs are registered.

TO-BE:
The test codes only use the Backend and toolchainEnv of ONE and mockup regardless of backend and toolchainEnv registered in the extension.ts file.

Co-authored-by: Bumsoo Ko <rhqjatn2398@naver.com>

Related: https://github.com/Samsung/ONE-vscode/pull/1628

### DIFF
--- a/src/Tests/Backend/Backend.test.ts
+++ b/src/Tests/Backend/Backend.test.ts
@@ -48,44 +48,51 @@ class BackendMockup implements Backend {
 }
 
 suite("Backend", function () {
+  setup(function () {
+    Object.keys(globalBackendMap).forEach(
+      (key) => delete globalBackendMap[key]
+    );
+    Object.keys(gToolchainEnvMap).forEach(
+      (key) => delete gToolchainEnvMap[key]
+    );
+  });
+
+  teardown(function () {
+    Object.keys(globalBackendMap).forEach(
+      (key) => delete globalBackendMap[key]
+    );
+    Object.keys(gToolchainEnvMap).forEach(
+      (key) => delete gToolchainEnvMap[key]
+    );
+  });
+
   suite("backendAPI", function () {
     test("registers a OneToolchain", function () {
       let oneBackend = new OneToolchain();
-      assert.strictEqual(Object.entries(globalBackendMap).length, 1);
+      API.registerBackend(oneBackend);
 
       const entries = Object.entries(globalBackendMap);
       assert.strictEqual(entries.length, 1);
+
       // this runs once
       for (const [key, value] of entries) {
         assert.strictEqual(key, oneBackendName);
         assert.deepStrictEqual(value, oneBackend);
       }
     });
-    test("registers a backend", function () {
-      assert.strictEqual(Object.entries(globalBackendMap).length, 1);
 
+    test("registers a backend", function () {
       let backend = new BackendMockup();
       API.registerBackend(backend);
 
       const entries = Object.entries(globalBackendMap);
-      assert.strictEqual(entries.length, 2);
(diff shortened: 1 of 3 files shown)


src/Tests/Backend/Backend.test.ts
```diff
@@ -48,44 +48,51 @@ class BackendMockup implements Backend {
...
+  setup(function () {
+    Object.keys(globalBackendMap).forEach(
+      (key) => delete globalBackendMap[key]
+    );
+    Object.keys(gToolchainEnvMap).forEach(
+      (key) => delete gToolchainEnvMap[key]
+    );
+  });
+
+  teardown(function () {
```
dayo09, line 60: Could you move this teardown function to the bottom of this "Backend" test suite? There is absolutely no problem in your code but our company static analysis tool will spot it as a defect. (related to https://github.com/Samsung/ONE-vscode/issues/1678)

src/Tests/View/InstallQuickInput.test.ts
```diff
@@ -45,21 +46,30 @@ suite("View", function () {
...
+
+      gToolchainEnvMap[oneBackendName] = oneToolchainEnv;
       gToolchainEnvMap[backendName] = toolchainEnv;
     });
 
     teardown(function () {
-      if (gToolchainEnvMap[backendName] !== undefined) {
-        delete gToolchainEnvMap[backendName];
-      }
+      Object.keys(gToolchainEnvMap).forEach(
```
dayo09, line 70: ditto

src/Tests/Backend/Backend.test.ts
```diff
@@ -48,44 +48,51 @@ class BackendMockup implements Backend {
...
+    Object.keys(globalBackendMap).forEach(
+      (key) => delete globalBackendMap[key]
```
dayo09, line 62: IMO It's better if we provide delete function for backend, which recursively deleting toolchain and executors. 
Maybe you can add TODO.
I see that you have added the test hook to clean global variables for each test, to prevent the tests from affecting each others.

jyoungyun: The backend is basically installed by sub extension, so if you want to delete the backend, just remove the sub extension. However, it seems that addtional discussion is needed about removing ONE and EdgeTPU backend, which are installed by ONE-vscode. We will consider providing the delete function later.

src/Tests/Backend/Backend.test.ts
profornnan: Thank you for noticing me!! I updated Toolchain test code to move teardown downwards in https://github.com/Samsung/ONE-vscode/pull/1686/commits/153a61833e95f1ee7e6ff2fb97400f6fdadfa5ca 😄

src/Tests/Backend/Backend.test.ts
jyoungyun: For now, ONE-vscode always has two toolchain installed, ONE and EdgeTPU. Therefore, it seems correct to implement the Backend test with the two installed.

profornnan: I understand :-D
I will close this PR and create a new PR that implements the Backend test code with the ONE and EdgeTPU installed.

PR #1690, opened 2023-09-26
[MPQEditor] Ignore from Coverage

dayo09:
This commit is to ignore provider and panel classes in TC coverage.


jyoungyun (approved):
LGTM

PR #1694, opened 2023-09-26
[Execute] Use DEVICE view name

jyoungyun:
This commit uses DEVICE view name instead of TARGET DEVICES.


dayo09 (approved):
👍

PR #1696, opened 2023-09-29
[EdgeTPU] Expand createCfg method for other cfg types

Hyeon-Uk:
This PR is about extending the createCfg method for other cfg types.

I wanted to make this UI like this.
image

But I couldn't find this API, and this program also gets the name for the output as a `quick pick`, so I made it a `quick pick` like this.
image

The option '.edgetpucfg' appears only in the tflite file.
image

# About Modularizing
I don't think this PR is enough to modularize. So please comment about best modularizing or refactoring

### DIFF
--- a/src/OneExplorer/OneExplorer.ts
+++ b/src/OneExplorer/OneExplorer.ts
@@ -19,7 +19,6 @@ import * as fs from "fs";
 import * as path from "path";
 import * as vscode from "vscode";
 
-import { CfgEditorPanel } from "../CfgEditor/CfgEditorPanel";
 import { obtainWorkspaceRoots } from "../Utils/Helpers";
 import { Logger } from "../Utils/Logger";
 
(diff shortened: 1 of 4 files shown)


src/OneExplorer/OneExplorer.ts
```diff
@@ -1026,8 +1025,65 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+
+    if (extName === "tflite") {
+      options.push({
+        label: ".edgetpucfg",
+        description: "configuration file of edgetpu compiler",
+      });
+    }
+
+    const selectedOption = await vscode.window.showQuickPick(options);
+    const selectedLabel = selectedOption ? selectedOption.label : ".cfg";
```
dayo09, line 1052: When there is no selectedOption, it should just return and exit.

Because of this line, it behaves unexpectedly: When the quickpick asks me whether to create ".cfg" or "edgetpucfg" and I hit `esc` to exit, but it still creates .cfg.

src/OneExplorer/OneExplorer.ts
```diff
@@ -1026,8 +1025,65 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+    extName: string
+  ): Promise<CfgInfo> {
+    const options = [
+      { label: ".cfg", description: "configuration file of onecc compiler" },
+    ];
+
+    if (extName === "tflite") {
+      options.push({
+        label: ".edgetpucfg",
+        description: "configuration file of edgetpu compiler",
```
dayo09, line 1047:
```suggestion
        description: "configuration file of edge tpu compiler",
```

Very trivial but, :-D

src/OneExplorer/OneExplorer.ts
```diff
@@ -1026,8 +1025,65 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+   *
+   * @param modelName A base model's name
+   * @param extName A base model's extension name
+   *
+   */
+  private async generateCfgInfo(
+    modelName: string,
+    extName: string
+  ): Promise<CfgInfo> {
+    const options = [
```
dayo09, line 1040: How about adding title / placeHolder to the options? 😄 I recommend:
- title: Pick a configuration to create
- placeHolder: cfg / edgetpucfg

src/OneExplorer/OneExplorer.ts
```diff
@@ -1188,3 +1243,10 @@ input_path=${modelName}.${extName}
     return this._tree;
   }
 }
+
+type CfgInfo = {
```
dayo09, line 1247: How about moving this type definition nearby the related function or to the top?

Hyeon-Uk: I move this type to the top of this file :-D

src/OneExplorer/OneExplorer.ts
Hyeon-Uk: I've done this job!

src/OneExplorer/OneExplorer.ts
```diff
@@ -1026,8 +1032,69 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+      options.push({
+        label: ".edgetpucfg",
+        description: "configuration file of edge tpu compiler",
+      });
+    }
+
+    const selectedOption = await vscode.window.showQuickPick(options, {
+      title: "Pick a configuration to create",
+      placeHolder: "cfg / edgetpucfg",
+    });
```
dayo09, line 1061: Actually, this lines should regard the extName (.tflite / .onnx), so we need some conditional approches.

Hyeon-Uk: I've completed this request and modularization additionally! :-)

hohee-hee (approved):
LGTM :)

src/OneExplorer/OneExplorer.ts
```diff
@@ -1026,8 +1028,56 @@ export class OneTreeDataProvider implements vscode.TreeDataProvider<Node> {
...
+   */
+  private async generateCfgInfo(
+    modelName: string,
+    extName: string
+  ): Promise<CfgInfo | undefined> {
+    const options: vscode.QuickPickItem[] = [
+      { label: ".cfg", description: "configuration file of onecc compiler" },
+    ];
+
+    if (extName === "tflite") {
```
dayo09, line 1046: Thanks that you change this to show two options only if its' tflite.
However, the other cases (*.onnx, *.pb) still shows the quickpick with "ONLY ONE CHOICE". It's meaningless. could you make it skip the quickpick selection if not tflite?

Hyeon-Uk: Because I think cfg files are likely to be created with other extensions in the future, instead of skipping `.tflite` files, I refactor them so that they can be skipped if there is only one option to choose from.

dayo09 (approved):
LGTM, Thanks :-D

PR #1697, opened 2023-09-30
[EdgeTPU] Add EdgeTPUCompiler

profornnan:
Implemented the EdgeTPUCompiler for EdgeTPUToolchain
Implemented code to test the EdgeTPUCompiler
- constructor
- getToolchainTypes
- parseVersion
- getToolchains
- getInstalledToolchains
- prerequisitesForGetToolchains

### DIFF
--- a/src/Backend/EdgeTPU/EdgeTPUToolchain.ts
+++ b/src/Backend/EdgeTPU/EdgeTPUToolchain.ts
@@ -14,13 +14,19 @@
  * limitations under the License.
  */
 
-import { Command } from "../Command";
-import { DebianToolchain } from "../ToolchainImpl/DebianToolchain";
-
+import * as cp from "child_process";
+import * as vscode from "vscode";
 import * as ini from "ini";
 import * as fs from "fs";
 import * as path from "path";
+
 import { Logger } from "../../Utils/Logger";
+import { pipedSpawnSync } from "../../Utils/PipedSpawnSync";
+import { Command } from "../Command";
+import { Compiler } from "../Compiler";
+import { PackageInfo, ToolchainInfo, Toolchains } from "../Toolchain";
+import { DebianToolchain } from "../ToolchainImpl/DebianToolchain";
+import { Version } from "../Version";
 
 class EdgeTPUDebianToolchain extends DebianToolchain {
   run(cfg: string): Command {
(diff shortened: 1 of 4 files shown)


rhqjatn2398 (approved):
LGTM 👍

dayo09:
@jyoungyun PTAL :-D

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
```diff
@@ -77,4 +83,191 @@ class EdgeTPUDebianToolchain extends DebianToolchain {
...
+    minor = Number(minor);
+    patch = Number(patch);
+
+    if (isNaN(major) || isNaN(minor) || isNaN(patch)) {
+      throw Error("Invalid version format.");
+    }
+
+    if (splitedVersion.length === 2 && !option) {
+      return new Version(major, minor, undefined);
+    }
```
dayo09, line 140: Q. What are these lines for? It seems that patch will be '0' by default when the split length is 2.

dayo09:
There are too many duplicated lines in OneCompiler and EdgeTPUCompiler class. How about extracting some sharing part? Likewise CfgData did.

src/Backend/EdgeTPU/EdgeTPUToolchain.ts
profornnan: The latest version of EdgeTPU Compiler is 16.0.

Setting the patch version as the default(16.0.0) causes the installation of edgetpu-compiler using aptitude to fail.

Therefore, I have implemented it to be set as undefined if there is no patch.

However, it seems that if we use undefined in cases where there is no patch version, there may be no need for default value settings.

I will try to modifying it while extracting the common parts of the OneCompiler and EdgeTPUCompiler classes.

profornnan:
OneCompiler and EdgeTPUCompiler are already using the Compiler interface.
Therefore, similar to DebianToolchain, I have extracted the common parts into DebianCompiler.
However, it might be difficult to know what parameters are passed when extending.

So, I would like to confirm whether the current implementation direction is suitable, or if there are any other advisable approaches, such as creating an abstract class for common parts like CfgData and passing settings and classes through functions.

dayo09 (approved):
LGTM for me :-D

dayo09:
@jyoungyun PTAL :-D

jyoungyun (approved):
LGTM Thank you ! :)

PR #1698, opened 2023-10-02
[EdgeTPU] Update Toolchain run notification and log

profornnan:
Modify notiTitle in JobConfig.ts from 'Running onecc...' to 'Running tools...'
Use active toolchain name in notification and log of ToolchainProvider _run method
Add notiTitle in ToolchainEnv run method to provide notification differently for each toolchain

### DIFF
--- a/src/Toolchain/JobConfig.ts
+++ b/src/Toolchain/JobConfig.ts
@@ -27,7 +27,7 @@ class JobConfig extends JobCommand {
     super(cmd);
     this.jobType = JobType.tConfig;
     this.name = "config";
-    this.notiTitle = "Running onecc...";
+    this.notiTitle = "Running tools...";
     this.valid = true;
     this.isCancelable = true;
   }
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -318,7 +318,7 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
   public _run(cfg: string): boolean {
     const [activeToolchainEnv, activeToolchain] =
       this.checkAvailableToolchain();
-    if (activeToolchainEnv === undefined || activeToolchain === undefined) {
+    if (!activeToolchainEnv || !activeToolchain) {
       return false;
     }
 
@@ -338,7 +338,7 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
 
     Logger.info(
       this.tag,
-      `Run tools with ${cfg} cfg and ${activeToolchainName}-${activeToolchain.info.version?.str()} toolchain.`
+      `Run ${activeToolchainName}-${activeToolchain.info.version?.str()} toolchain with ${cfg} cfg.`
     );
     activeToolchainEnv.run(cfg, activeToolchain).then(
       () => notifySuccess(),
--- a/src/Toolchain/ToolchainProvider.ts
+++ b/src/Toolchain/ToolchainProvider.ts
@@ -316,27 +316,29 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
   }
 
   public _run(cfg: string): boolean {
-    /* istanbul ignore next */
-    const notifySuccess = () => {
-      vscode.window.showInformationMessage("Onecc has run successfully.");
-    };
-
-    /* istanbul ignore next */
-    const notifyError = () => {
-      this.error("Running onecc has failed.");
-    };
-
     const [activeToolchainEnv, activeToolchain] =
       this.checkAvailableToolchain();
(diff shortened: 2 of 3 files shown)


src/Toolchain/ToolchainProvider.ts
```diff
@@ -316,27 +316,29 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
...
-    if (activeToolchainEnv === undefined || activeToolchain === undefined) {
-      return false;
-    }
-
     Logger.info(
       this.tag,
-      `Run onecc with ${cfg} cfg and ${
-        activeToolchain.info.name
-      }-${activeToolchain.info.version?.str()} toolchain.`
+      `Run tools with ${cfg} cfg and ${activeToolchainName}-${activeToolchain.info.version?.str()} toolchain.`
```
dayo09, line 341:
```suggestion
      `Running ${cfg} with ${activeToolchainName}-${activeToolchain.info.version?.str()}`
```

How about this? It's too long and getting shortened by '...'

Check out this:
image

src/Toolchain/ToolchainProvider.ts
```diff
@@ -316,27 +316,29 @@ export class ToolchainProvider implements vscode.TreeDataProvider<BaseNode> {
   }
 
   public _run(cfg: string): boolean {
+    const [activeToolchainEnv, activeToolchain] =
+      this.checkAvailableToolchain();
+    if (activeToolchainEnv === undefined || activeToolchain === undefined) {
```
dayo09, line 321:
```suggestion
    if (!activeToolchainEnv || !activeToolchain) {
```

FYI, this is easy and shorter synonym.

src/Toolchain/JobConfig.ts
```diff
@@ -27,7 +27,7 @@ class JobConfig extends JobCommand {
     super(cmd);
     this.jobType = JobType.tConfig;
     this.name = "config";
-    this.notiTitle = "Running onecc...";
+    this.notiTitle = "Running tools...";
```
dayo09, line 30: How about this:

You can edit `run` in `ToolchainEnv.ts` to apply notiTitle according to the toolchain name.

image

dayo09: I know this is an awkward comment after answering you that 'Runnign tools' would do 😅

src/Toolchain/ToolchainProvider.ts
profornnan: Thank you for your advice~!! It's better then before 😄 I'll apply it

src/Toolchain/JobConfig.ts
profornnan: It seems great to know which toolchain we are using. I will accept your advice.

However, I have another question about this PR, and I would like to hear your opinion.

1. Is it okay to keep notiTitle in JobConfig.ts as "Running tools..."? or revert it back to 'Running onecc...?
2. Should I change line 341 as shown below because we use the toolchain name in notifications of ToolchainEnv and ToolchainProvider?

src/Toolchain/ToolchainProvider.ts
profornnan, line 341:
```suggestion
      `Run ${activeToolchainName}-${activeToolchain.info.version?.str()} toolchain with ${cfg} cfg.`
```

src/Toolchain/JobConfig.ts
dayo09: 1. Sure. I think it's anyways okay to keep `notiTitle`. 
2. Yes, I think it's better to read. https://github.com/Samsung/ONE-vscode/pull/1698#discussion_r1342807624


Maybe @jyoungyun have another opinions, let's wait for that.

dayo09:
Let me know when it's ready :-D

profornnan:
I have applied all the feedback. : )

rhqjatn2398 (approved):
LGTM 👍

src/Toolchain/ToolchainProvider.ts
glistening:
```js
    if (activeToolchainEnv === undefined || activeToolchain === undefined) {
```
and
```js
    if (!activeToolchainEnv || !activeToolchain) {
```
are different as I understand. The former returns `true` only `undefined` while the latter returns `true`, not only for `undefined`, but also for 0, empty string("", ''), NaN, null.

dayo09: @glistening That's right. I meant that here it would mean the same. Usually returning `0, empty string("", ''), NaN, null` from functions mean the similar result, so I take this abbreviation. :-D

dayo09 (approved):
LGTM :-D

dayo09:
@glistening I think you left the review to add the comment. I will merge PR right away :-D

PR #1703, opened 2023-10-05
DRAFT: [OneExplorer] Show under multiple parents

dayo09:
Currently, config nodes are shown under the first parent written in the file.
To resolve the issue, multimap for node map is required.


dayo09:
All merged :-D

Issue #1706, opened 2023-10-11
[Test] Fix async tests


dayo09:
### What?

Below kinds of bug arbitrarily occurs during PR test.
Let's fix those async tests to finish properly.
 
```
  1 pending
  2 failing
  1) MPQEditor
       MPQEditorProvider
         #updateDocumentBy
           update document by:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/ONE-vscode/ONE-vscode/out/Tests/MPQEditor/MPQEditor.test.js)
  	at listOnTimeout (node:internal/timers:569:17)
  	at process.processTimers (node:internal/timers:512:7)

  2) Utils
       Hash
         #generateHash()
           NEG: generate non-existing file hash:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/ONE-vscode/ONE-vscode/out/Tests/Utils/Hash.test.js)
  	at listOnTimeout (node:internal/timers:569:17)
  	at process.processTimers (node:internal/timers:512:7)

Error: 2 tests failed
	at /home/runner/work/ONE-vscode/ONE-vscode/out/Tests/index.js:149:26
	at done (/home/runner/work/ONE-vscode/ONE-vscode/node_modules/mocha/lib/mocha.js:1058:7)
Extension host test runner error Error: 2 tests failed
	at /home/runner/work/ONE-vscode/ONE-vscode/out/Tests/index.js:149:26
	at done (/home/runner/work/ONE-vscode/ONE-vscode/node_modules/mocha/lib/mocha.js:1058:7)
Asking native host service to exit with code 1.
[main 2023-10-11T05:32:36.366Z] Extension host with pid 2419 exited with code: 0, signal: unknown.
Exit code:   1
Failed to run tests: 
```

dayo09:
Done!

dayo09:
### It's not resolved

https://github.com/Samsung/ONE-vscode/actions/runs/6543377167/job/17768011868?pr=1710

```
  1) MPQEditor
       MPQEditorProvider
         #updateDocumentBy
           update document by:
[...]

PR #1714, opened 2023-10-17
[Test] Alter mocha timeout

dayo09:
This commit alter mocha timeout from 2000(default) to 10000 ms. It's to prevent timeout failure during test.

---

For #1706

I suspect it as a long latency problem. By rising timeout up to 10000 ms, (default: 2000 ms) the test passes at once. Maybe the CI machine sometime get slower and does not pass within the timeout. They always pass on local machine, so let's simply change the timeout.


mhs4670go (approved):
LGTM

PR #1724, opened 2023-11-01
[EdgeTPU] Bind EdgeTPUCfg Node to the backend's registration

dayo09:
This commit shows edgetpucfg node only if EdgeTPU Backend is registered.


jyoungyun (approved):
LGTM

PR #1672, opened 2023-09-22
[EdgeTpu] Separate ConfigObject's logic dependent on Backend by introducing ConfigSetting

imsw0529:
Related to issue #1621, draft #1639

This commit separate ConfigObject's logic dependent on Backend by introducing ConfigSetting

- Introduce abstract class ConfigSetting for separate logic
- Introduce class OneConfigSetting extends ConfigSetting for ONE Backend logic
- Apply OneConfigSetting to ConfigObject
- Add Test code for ConfigSetting associated code

Co-author: SeungHyunH <vkflek1232@naver.com>

ONE-vscode-DCO-1.0-Signed-off-by: Seongwon Im <imsw0529@gmail.com>

PR #1732, opened 2023-11-02
[EdgeTpu] Separate ConfigObject logic into ConfigSetting 

dayo09:
This commit separate ConfigObject's logic dependent on Backend by introducing ConfigSetting

- Introduce abstract class ConfigSetting for separate logic
- Introduce class OneConfigSetting extends ConfigSetting for ONE Backend logic
- Apply OneConfigSetting to ConfigObject
- Add Test code for ConfigSetting associated code

ONE-vscode-DCO-1.0-Signed-off-by: Seongwon Im <imsw0529@gmail.com>
Co-authored-by: SeungHyunH <vkflek1232@naver.com>

---

Original PR #1672


src/OneExplorer/ConfigObject.ts
```diff
@@ -112,12 +121,31 @@ export class ConfigObj {
...
 
+  /**
+   * @brief getter for config setting
+   */
+  public get configSetting(): ConfigSetting {
+    let configSetting: ConfigSetting;
+    switch (this.configType) {
+      case CfgType.one:
+      default:
+        configSetting = new OneConfigSetting();
```
jyoungyun, line 132: (Q) Does it support multiple config setting? Is this `configSetting` shared on all backends?

jyoungyun (approved):
LGTM

src/OneExplorer/ConfigObject.ts
dayo09: > Does it support multiple config setting?

The answer first, is : Yes.

Here, it represents the configuration type of a file. So each file can have one config setting. Here, *.cfg corresponds to 
OneConfigSetting and *.edgetpucfg corresponds to EdgeTPUConfigSetting (in the following PR).

